PR #112, opened 2014-12-08
Improve autocompletion

Fatme:
* Remove hardcoded commands
* Remove getPropSchemaAction
* Add completionData to ICommand interface - This way each command can specify data that will be used for autocompletion. For example build command can specify platforms("android", "ios", "wp8"),
plugin command can specify available plugin names, prop command can specify available property names i etc.

This is helper PR for this story http://teampulse.telerik.com/view#item/280276


rosen-vladimirov:
:+1:

PR #128, opened 2014-12-18
Replace underscore with lodash (finally!)

teobugslayer:


helpers.ts
```diff
@@ -3,7 +3,6 @@
 import fs = require("fs");
 import path = require("path");
 import util = require("util");
-import _ = require("underscore");
```
rosen-vladimirov: We need this import (of course change underscore with lodash). You have added a global _ to point to lodash in lib/common/bootstrap, which is loaded after appbuilder's options.ts, so at this point there's no _ and trying to execute npm install fails. options.ts has reference to common/helpers.ts and loads it before common/bootstrap.ts, so _ is not defined when the file is loaded and the installation fails.

rosen-vladimirov:
:+1:

PR #130, opened 2014-12-18
Support for WP8 Ion

teobugslayer:
See http://teampulse.telerik.com/view#item/262237

### DIFF
--- a/mobile/app-identifier.ts
+++ b/mobile/app-identifier.ts
@@ -193,6 +193,32 @@ export class IOSNativeScriptCompanionAppIdentifier implements Mobile.IAppIdentif
 
 }
 
+export class WP8CompanionAppIdentifier implements Mobile.IAppIdentifier {
+	get appIdentifier(): string {
+		return "{9155af5b-e7ed-486d-bc6b-35087fb59ecc}";
+	}
+
+	get deviceProjectPath(): string {
+		return ""; // this is used only on Android for Lollipop
+	}
+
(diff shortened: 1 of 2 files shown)


mobile/app-identifier.ts
```diff
@@ -193,6 +193,32 @@ export class IOSNativeScriptCompanionAppIdentifier implements Mobile.IAppIdentif
 
 }
 
+export class WP8CompanionAppIdentifier implements Mobile.IAppIdentifier {
+	get appIdentifier(): string {
+		return "";
+	}
+
+	get deviceProjectPath(): string {
+		return "";
```
rosen-vladimirov: Is this really empty string? Or maybe it is not used at all when it's inside the companion app?

rosen-vladimirov:
:+1:  after checking my comment (not sure if it's relevant at all)

mobile/app-identifier.ts
```diff
@@ -193,6 +193,32 @@ export class IOSNativeScriptCompanionAppIdentifier implements Mobile.IAppIdentif
 
 }
 
+export class WP8CompanionAppIdentifier implements Mobile.IAppIdentifier {
+	get appIdentifier(): string {
+		return "";
```
Fatme: Why this is empty string? It is strange to me wp8 companion app identifier to be empty string...

rosen-vladimirov:
:+1:

PR #134, opened 2015-01-09
Android Emulator now checks whether the SDK is correctly configured

teobugslayer:
Also, gives a proper message to the user

Implements http://teampulse.telerik.com/view#item/279833

### DIFF
--- a/mobile/android/android-emulator-services.ts
+++ b/mobile/android/android-emulator-services.ts
@@ -9,7 +9,7 @@ import osenv = require("osenv");
 import path = require("path");
 import util = require("util");
 import hostInfo = require("../../../common/host-info");
-import MobileHelper = require("./../mobile-helper");
+import MobileHelper = require("../mobile-helper");
 import options = require("../../options");
 import helpers = require("../../helpers");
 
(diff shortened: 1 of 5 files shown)


mobile/android/android-emulator-services.ts
```diff
@@ -32,6 +32,20 @@ class AndroidEmulatorServices implements Mobile.IEmulatorPlatformServices {
 		iconv.extendNodeEncodings();
 	}
 
+	public checkDependencies(): IFuture<void> {
+		var future = new Future<void>();
+		var proc = this.$childProcess.spawn('emulator', ['-help']);
+		proc.on("error", (args:any) => {
+			this.$errors.fail("The Android SDK is not configured properly. " +
```
rosen-vladimirov: or maybe future.throw(...) not sure which one's better

teobugslayer: They have the same outcome.

mobile/android/android-emulator-services.ts
```diff
@@ -32,6 +32,20 @@ class AndroidEmulatorServices implements Mobile.IEmulatorPlatformServices {
...
+				"Verify that you have installed the Android SDK and that you have added its `platform-tools` and `tools` directories to your PATH environment variable.");
+		});
+		proc.on("exit", (code:number) => {
+			future.return();
```
rosen-vladimirov: is it possible to have both error and exit events, we stick to the following code:
```Typescript
if (!future.isResolved()) { 
    future.return();
}
```
Same is valid for on error callback

rosen-vladimirov: From nodejs documentation:
```
Note that the exit-event may or may not fire after an error has occured. If you are listening on both events to fire a function, remember to guard against calling your function twice.
```
http://nodejs.org/api/child_process.html#child_process_event_error

rosen-vladimirov:
:+1:

PR #136, opened 2015-01-12
Use genymotion emulator

rosen-vladimirov:
Add geny option. Add check if it is specified to start the player with geny option and use player from Genymotion folder. Add check dependency for Genymotion if options.geny is specified.
Add check if --avd and --geny options are both specified and throws error in such case. Get path to VBox on Windows through registry by checking 64-bits registry and 32 bits registries. Do not start genymotion emulator in case it is already started - use VBoxManage list runningvms in order to check if it is running. Use 'VBoxManage list vms' to check if argument passed to --geny option is correct.

http://teampulse.telerik.com/view#item/274611


mobile/android/android-emulator-services.ts
```diff
@@ -120,15 +152,124 @@ class AndroidEmulatorServices implements Mobile.IEmulatorPlatformServices {
...
+		var vboxManage = this.$childProcess.spawn(vboxPath, ["list", parameter]);
+
+		vboxManage.stdout.on("data", (chunk: string) => {
+			if(chunk) {
+				allData += chunk;
+			}
+		});
+		vboxManage.stderr.on("data", (error: Error) => {
+			// maybe use log trace level to print the error
+			this.$errors.fail("Unable to get currently running virtual machines. Check if VirtualBox is installed correctly.");
```
rosen-vladimirov: Remove word running. In some cases we list all vms, not only running ones.

mobile/android/android-emulator-services.ts
```diff
@@ -120,15 +152,124 @@ class AndroidEmulatorServices implements Mobile.IEmulatorPlatformServices {
...
+		});
+
+		vboxManage.on("exit", () => {
```
Fatme: I suggest you to use`var childProcess = this.$childProcess.spawnFromEvent(vboxPath, ["list", parameter], "exit", {throwError: false}).wait(); 
if(childProcess. stderr) {
  // maybe use log trace level to print the error
  this.$errors.fail("Unable to get currently running virtual machines. Check if VirtualBox is installed correctly.");
}`

mobile/android/android-emulator-services.ts
```diff
@@ -33,13 +38,36 @@ class AndroidEmulatorServices implements Mobile.IEmulatorPlatformServices {
...
+	}
+
+	private checkGenymotionConfiguration(): IFuture<void> {
+		var future = new Future<void>();
+		var proc = this.$childProcess.spawn('player', ['-help']);
+		proc.on("error", (args: any) => {
+			this.$errors.fail("Genymotion is not configured properly. " +
+				"Verify that you have installed Genymotion and that you have added its installation directory to your PATH environment variable.");
+		});
+		proc.on("exit", (code: number) => {
```
Fatme: Use `this.$childProcess.spawnFromEvent("player", ["-help"], "exit", {throwError: false})`

mobile/android/android-emulator-services.ts
```diff
@@ -33,13 +38,36 @@ class AndroidEmulatorServices implements Mobile.IEmulatorPlatformServices {
...
+	private checkAndroidSDKConfiguration(): IFuture<void> {
 		var future = new Future<void>();
 		var proc = this.$childProcess.spawn('emulator', ['-help']);
-		proc.on("error", (args:any) => {
+		proc.on("error", (args: any) => {
 			this.$errors.fail("The Android SDK is not configured properly. " +
 				"Verify that you have installed the Android SDK and that you have added its `platform-tools` and `tools` directories to your PATH environment variable.");
 		});
-		proc.on("exit", (code:number) => {
+		proc.on("exit", (code: number) => {
```
Fatme: Again use `this.$childProcess.spawnFromEvent('emulator', ['-help'], "exit", {throwError: false})`

mobile/android/android-emulator-services.ts
```diff
@@ -169,7 +313,7 @@ class AndroidEmulatorServices implements Mobile.IEmulatorPlatformServices {
 			var emulatorDevices: string[] = [];
 			var outputRaw = this.$childProcess.execFile(this.$staticConfig.adbFilePath, ['devices']).wait().split(os.EOL);
 			_.each(outputRaw, (device: string) => {
-				var rx = device.match(/^(emulator-\d+)\s+device$/);
+				var rx = options.geny ? device.match(/^(\d+\.\d+\.\d+\.\d+:\d+)\s+device$/) : device.match(/^(emulator-\d+)\s+device$/);
```
rosen-vladimirov: if geny option is specified maybe we can shorten the regex: ^((\d+\.){3}\d+:\d+)\s+device$

mobile/android/android-emulator-services.ts
```diff
@@ -120,15 +152,124 @@ class AndroidEmulatorServices implements Mobile.IEmulatorPlatformServices {
...
+			return shouldStartEmulator;
+		}).future<boolean>()();
+	}
+
+	private getPathToVBox(): IFuture<string> {
+		if(hostInfo.isWindows()) {
+			var result = new Future<string>();
+			var regKey64 = new Winreg({
+				hive: Winreg.HKLM,
+				key: '\\SOFTWARE\\Oracle\\VirtualBox',
```
Fatme: Consider extracting this as constant

mobile/android/android-emulator-services.ts
```diff
@@ -120,15 +152,124 @@ class AndroidEmulatorServices implements Mobile.IEmulatorPlatformServices {
...
+				regVersion: '64'
+			});
+
+			var regKey32 = new Winreg({
+				hive: Winreg.HKLM,
+				key: '\\SOFTWARE\\Oracle\\VirtualBox',
+				regVersion: '32'
+			});
+
+			regKey64.get("InstallDir", (err: Error, item: any) => {
```
Fatme: InstallDir maybe should be constant

mobile/android/android-emulator-services.ts
Fatme: Goot catch! :smile_cat:

mobile/android/android-emulator-services.ts
```diff
@@ -120,15 +152,124 @@ class AndroidEmulatorServices implements Mobile.IEmulatorPlatformServices {
...
+		vboxManage.stderr.on("data", (error: Error) => {
+			// maybe use log trace level to print the error
+			this.$errors.fail("Unable to get currently running virtual machines. Check if VirtualBox is installed correctly.");
+		});
+
+		vboxManage.on("exit", () => {
+			_(allData.split(os.EOL)).each(row => {
+				// sample output row: "GN5" {1c268a7b-aa04-49c9-aabd-90595a747dec}
+				var rx = row.match(/^"(.+)"\s+\{(.+)\}$/);
+				if(rx && rx[1]) {
```
teobugslayer: Should you also check for rx[2] existance?

mobile/android/android-emulator-services.ts
```diff
@@ -120,15 +152,124 @@ class AndroidEmulatorServices implements Mobile.IEmulatorPlatformServices {
...
+
+	private shouldStartGenymotionEmulator(image: string): IFuture<boolean> {
+		return (() => {
+			if(!options.geny) {
+				return false;
+			}
+
+			var hasVmWithSpeciedImage = false;
+			var allVMs: VirtualMachine[] = this.listVirtualMachines("vms").wait();
+			_(allVMs).each((vm: VirtualMachine) => {
```
teobugslayer: Do we need the types?

mobile/android/android-emulator-services.ts
```diff
@@ -120,15 +152,124 @@ class AndroidEmulatorServices implements Mobile.IEmulatorPlatformServices {
...
+				if(vm.name === image || vm.identifier === image) {
+					hasVmWithSpeciedImage = true;
+					// break each
+					return false;
+				}
+			});
+
+			if(!hasVmWithSpeciedImage) {
+				this.$errors.fail("There is no VirtualBox machine with name %s. Please verify specified name is correct.", image);
```
teobugslayer: maybe drop "is correct"? It is implied by "verify".

mobile/android/android-emulator-services.ts
```diff
@@ -57,7 +85,11 @@ class AndroidEmulatorServices implements Mobile.IEmulatorPlatformServices {
 
 	public startEmulator(app: string, emulatorOptions?: Mobile.IEmulatorOptions): IFuture<void> {
 		return (() => {
-			var image = options.avd || this.getBestFit().wait();
+			if(options.avd && options.geny) {
+				this.$errors.fail("You cannot specify both --avd and --geny options. Please use only one of them.");
```
rosen-vladimirov: Should we print command's help here?

mobile/android/android-emulator-services.ts
rosen-vladimirov: And print the error with log trace as the comment says

mobile/android/android-emulator-services.ts
rosen-vladimirov: Also I think we shouldn't print command's help here. It looks like the user knows how to use it, just Genymotion's image name is incorrect.

mobile/android/android-emulator-services.ts
```diff
@@ -120,15 +152,124 @@ class AndroidEmulatorServices implements Mobile.IEmulatorPlatformServices {
...
+				hive: Winreg.HKLM,
+				key: '\\SOFTWARE\\Oracle\\VirtualBox',
+				regVersion: '32'
+			});
+
+			regKey64.get("InstallDir", (err: Error, item: any) => {
+				if(err) {
+					regKey32.get("InstallDir", (err: Error, item: any) => {
+						if(err) {
+							this.$errors.fail({ formatStr: "Make sure VirtualBox is installed correctly.", suppressCommandHelp: true });
```
rosen-vladimirov: Maybe if we fail with regKey64 and regKey32, we should try getting the value without passing regVersion. I'm not sure for which versions of REG tool /reg:32 (/reg:64) option are supported. If there's a system where we fail when passing them, maybe the fail is because of the usage of this parameters. So we can try executing the command without this parameter and see if there's a value in the registry.

mobile/android/android-emulator-services.ts
```diff
@@ -120,15 +152,124 @@ class AndroidEmulatorServices implements Mobile.IEmulatorPlatformServices {
...
+		return future;
+	}
+
+	private shouldStartGenymotionEmulator(image: string): IFuture<boolean> {
+		return (() => {
+			if(!options.geny) {
+				return false;
+			}
+
+			var hasVmWithSpeciedImage = false;
```
rosen-vladimirov: Typo -> Should be ```hasVmWithSpecifiedImage```

mobile/android/android-emulator-services.ts
rosen-vladimirov: Yep, also which registry should be checked first - 32 bits or 64bits? Maybe the order should depend on ```process.arch ```  (its values are x64 or ia32).

mobile/android/android-emulator-services.ts
rosen-vladimirov: Ignore this comment, no need to do this.

mobile/android/android-emulator-services.ts
teobugslayer: I've seen IPv6 based addresses and this regexp will fail with these.

mobile/android/android-emulator-services.ts
teobugslayer: IMHO, it does not matter. On 64 bit OS you'll have 64-bit registration and on 32-bit OS - 32 bit registration.

rosen-vladimirov:
Code is not working correctly when there's more than one running Genymotion emulator. In this case we always deploy to the first image, which is incorrect.

mobile/android/android-emulator-services.ts
teobugslayer: I think we should.

mobile/android/android-emulator-services.ts
teobugslayer: Actually, under Windows 7, the /reg:xx parameters are NOT supported.

mobile/android/android-emulator-services.ts
```diff
@@ -120,29 +147,113 @@ class AndroidEmulatorServices implements Mobile.IEmulatorPlatformServices {
...
+		}).future<string>()();
+	}
+
+	private getNameFromNativeEmulatorId(emulatorId: string): IFuture<string> {
+		var match = emulatorId.match(AndroidEmulatorServices.RUNNING_ANDROID_EMULATOR_REGEX);
+		var portNumber: string;
+		if(match && match[1]) {
+			portNumber = match[1];
+		} else {
+			Future.fromResult("");
```
tailsu: missing `return`

mobile/android/android-emulator-services.ts
```diff
@@ -150,30 +261,47 @@ class AndroidEmulatorServices implements Mobile.IEmulatorPlatformServices {
...
+			var genymotionEmulators: string[] = [];
+			_(adbDevicesOutput).each(row => {
+				var isNativeEmulator = row.match(AndroidEmulatorServices.RUNNING_ANDROID_EMULATOR_REGEX);
+				if(!isNativeEmulator) {
+					var match = row.match(/^(.+?)\s+device$/);
+					if(match && match[1]) {
+						// possible genymotion emulator
+						var emulatorId = match[1];
+						var manufacturer = this.$childProcess.execFile(this.$staticConfig.adbFilePath, ["-s", emulatorId, "shell", "getprop", "ro.product.manufacturer"]).wait();
+						if(manufacturer.match(/^Genymotion/)) {
```
tailsu: case-insensitive

mobile/android/android-emulator-services.ts
```diff
@@ -120,29 +147,113 @@ class AndroidEmulatorServices implements Mobile.IEmulatorPlatformServices {
...
 
 			while(hasTimeLeft || isInfiniteWait) {
-				if(runningEmulators.length > initiallyRunningEmulators.length || (runningEmulators.length > 0 && !options.avd)) {
-					break;
+				emulatorId = this.getRunningEmulatorId(image).wait();
+				if(emulatorId) {
+					return emulatorId;
 				}
 
 				this.sleep(10000); // the emulator definitely takes its time to wake up
```
tailsu: Why not move `sleep()` to the utilities lib?

mobile/android/android-emulator-services.ts
```diff
@@ -120,29 +147,113 @@ class AndroidEmulatorServices implements Mobile.IEmulatorPlatformServices {
...
+		return ((): string => {
+			var runningEmulators = this.getRunningEmulators().wait();
+			if(runningEmulators.length === 0) {
+				return "";
+			}
+
+			// if we get here, there's at least one running emulator
+			var emulatorId: string;
+
+			_(runningEmulators).each(emulator => {
```
ligaz: You can use [`_.find`](https://lodash.com/docs#find) here to simplify this.

mobile/android/android-emulator-services.ts
```diff
@@ -150,30 +261,47 @@ class AndroidEmulatorServices implements Mobile.IEmulatorPlatformServices {
...
-			}
-
 			return emulatorId;
 		}).future<string>()();
 	}
 
+	private getRunningGenymotionEmulators(adbDevicesOutput: string[]): IFuture<string[]> {
+		return ((): string[]=> {
+			var genymotionEmulators: string[] = [];
+			_(adbDevicesOutput).each(row => {
```
tailsu: This is a great place to try the fiber fixes by running this in parallel!

rosen-vladimirov: Yep, I was thinking the same :) Hope it will work fine.

mobile/android/android-emulator-services.ts
teobugslayer: or maybe (in pseudo-code):
```
var res = future.fromResult("");
if (!(match && match[1])) { return res; }
......
return res;
```
and delete the future var;

mobile/android/android-emulator-services.ts
teobugslayer: I would postpone until it is needed anywhere else - the utilities lib is already a bag of all tricks

mobile/android/android-emulator-services.ts
```diff
@@ -120,29 +147,113 @@ class AndroidEmulatorServices implements Mobile.IEmulatorPlatformServices {
...
+			if(runningEmulators.length === 0) {
+				return "";
+			}
+
+			// if we get here, there's at least one running emulator
+			var emulatorId: string;
+
+			_(runningEmulators).each(emulator => {
+				var emulatorName = options.geny ? this.getNameFromGenymotionEmulatorId(emulator).wait() : this.getNameFromNativeEmulatorId(emulator).wait();
+				if(emulatorName === image) {
```
teobugslayer: I would move the function choice outside of the for each cycle to shorten this line a bit.

mobile/android/android-emulator-services.ts
rosen-vladimirov: Maybe I do not understand your idea, but I need the future in order to resolve it when client.end is called. Using Future.fromResult is resolving the future, so I cannot resolve it in end event of the client (leads to Future resolved more than once error).

mobile/android/android-emulator-services.ts
```diff
@@ -150,17 +253,45 @@ class AndroidEmulatorServices implements Mobile.IEmulatorPlatformServices {
...
+			var futures: IFuture<string>[] = [];
+			_(adbDevicesOutput).filter(r => !r.match(AndroidEmulatorServices.RUNNING_ANDROID_EMULATOR_REGEX)).each(row => {
+				var match = row.match(/^(.+?)\s+device$/);
+				if(match && match[1]) {
+					// possible genymotion emulator
+					var emulatorId = match[1];
+					var future = this.checkForGenymotionProductManufacturer(emulatorId);
+					futures.push(future);
+				}
+			});
```
Fatme: You can do improve this: `var futures = _(adbDevicesOutput).filter(r => !r.match(AndroidEmulatorServices.RUNNING_ANDROID_EMULATOR_REGEX)).map(row => {
				var match = row.match(/^(.+?)\s+device$/);
				if(match && match[1]) {
					// possible genymotion emulator
					var emulatorId = match[1];
					return this.checkForGenymotionProductManufacturer(emulatorId);
				}`

mobile/android/android-emulator-services.ts
```diff
@@ -150,17 +253,45 @@ class AndroidEmulatorServices implements Mobile.IEmulatorPlatformServices {
...
+
+			Future.wait(futures);
+
+			_(futures).each(future => {
+				var manufacturer = future.get();
+				if(manufacturer) {
+					genymotionEmulators.push(manufacturer.toString());
+				}
+			});
+
```
Fatme: `var genymotionEmulators = _(futures).filter(future => { var manufacturer future.get(); return manufacturer !== undefined; });`

mobile/android/android-emulator-services.ts
```diff
@@ -168,12 +299,17 @@ class AndroidEmulatorServices implements Mobile.IEmulatorPlatformServices {
...
-			});
+			if(options.geny) {
+				emulatorDevices = this.getRunningGenymotionEmulators(outputRaw).wait();
+			} else {
+				_.each(outputRaw, (device: string) => {
+					var rx = device.match(AndroidEmulatorServices.RUNNING_ANDROID_EMULATOR_REGEX);
+
+					if(rx && rx[1]) {
+						emulatorDevices.push(rx[1]);
+					}
```
Fatme: `emulatorDevices = _(outputRaw).filter(device => var rx = device.match(AndroidEmulatorServices.RUNNING_ANDROID_EMULATOR_REGEX);
return rx && rx[1])`

mobile/android/android-emulator-services.ts
```diff
@@ -33,17 +39,34 @@ class AndroidEmulatorServices implements Mobile.IEmulatorPlatformServices {
...
+	}
 
-		return future;
+	private checkAndroidSDKConfiguration(): IFuture<void> {
+		return (() => {
+			var proc = this.$childProcess.spawnFromEvent('emulator', ['-help'], "exit", undefined, { throwError: false }).wait();
+
+			if(proc.stderr) {
+				this.$errors.fail("The Android SDK is not configured properly. " +
+					"Verify that you have installed the Android SDK and that you have added its `platform-tools` and `tools` directories to your PATH environment variable.");
```
teobugslayer: Do we need the help of the command printed here? It looks like the command line is OK but the dependency is broken.

mobile/android/android-emulator-services.ts
```diff
@@ -33,17 +39,34 @@ class AndroidEmulatorServices implements Mobile.IEmulatorPlatformServices {
...
+		}).future<void>()();
+	}
+
+	private checkGenymotionConfiguration(): IFuture<void> {
+		return (() => {
+			var proc = this.$childProcess.spawnFromEvent("genyshell", ["-h"], "exit", undefined, { throwError: false }).wait();
+
+			if(proc.stderr) {
+				this.$errors.fail("Genymotion is not configured properly. " +
+					"Verify that you have installed Genymotion and that you have added its installation directory to your PATH environment variable.");
```
teobugslayer: Do we need the help of the command printed here? It looks like the command line is OK but the dependency is broken.

mobile/android/android-emulator-services.ts
```diff
@@ -120,29 +147,105 @@ class AndroidEmulatorServices implements Mobile.IEmulatorPlatformServices {
...
+	private getNameFromGenymotionEmulatorId(emulatorId: string): IFuture<string> {
+		return (() => {
+			var modelOutputLines: string = this.$childProcess.execFile(this.$staticConfig.adbFilePath, ["-s", emulatorId, "shell", "getprop", "ro.product.model"]).wait();
+			this.$logger.trace(modelOutputLines);
+			var model = (<string>_.first(modelOutputLines.split(os.EOL))).trim();
+			return model;
+		}).future<string>()();
+	}
+
+	private getNameFromNativeEmulatorId(emulatorId: string): IFuture<string> {
```
teobugslayer: What about naming this function similar to "getNameFromSdkEmulatorId" -> both are native emulators.

Fatme:
:+1:

teobugslayer:
:+1:

PR #137, opened 2015-01-15
Add function to print message after timeout

rosen-vladimirov:
Add new function to helpers.ts which prints a message on the same line (using printInfoMessageOnSameLine method from the same helpers.ts) after a specific timeout had passed. Using future to guarantee that the function will not finish until timeout is passed and message is printed.

This function is needed when building project in order to show some kind of a progress (in this case Building........) when executing long operation. This way the users will know something is still working there.

http://teampulse.telerik.com/view#item/278000


teobugslayer:
:+1:

PR #139, opened 2015-01-19
FileSystem.writeJson defaults to a TAB character for pretty-printing

teobugslayer:
See http://teampulse.telerik.com/view#item/282660


rosen-vladimirov:
:+1:

PR #140, opened 2015-01-19
Add shorthands for help command

rosen-vladimirov:
Add some shorthands for help command: -h, ?, /?. Fix usage of shorthands, so when you specify -v for example to really use verbose mode.

http://teampulse.telerik.com/view#item/284030


ErjanGavalji:
:+1:

teobugslayer:
:-1:

PR #142, opened 2015-01-19
Run application on iOS device on OSX

Fatme:
http://teampulse.telerik.com/view#item/283764

### DIFF
--- a/commands/device/list-applications.ts
+++ b/commands/device/list-applications.ts
@@ -10,7 +10,7 @@ export class ListApplicationsCommand implements ICommand {
 
 	public execute(args: string[]): IFuture<void> {
 		return (() => {
-			this.$devicesServices.initialize(undefined, options.device, {skipInferPlatform: true}).wait();
+			this.$devicesServices.initialize({ deviceId: options.device, skipInferPlatform: true }).wait();
 
 			var action = (device: Mobile.IDevice) =>  { return (() => device.listApplications()).future<void>()(); };
 			this.$devicesServices.execute(action).wait();
--- a/commands/device/list-applications.ts
+++ b/commands/device/list-applications.ts
@@ -1,0 +1,20 @@
+///<reference path="../../../.d.ts"/>
+"use strict";
+
+import options = require("./../../options");
+
+export class ListApplicationsCommand implements ICommand {
+	constructor(private $devicesServices: Mobile.IDevicesServices) { }
+
+	allowedParameters: ICommandParameter[] = [];
+
+	public execute(args: string[]): IFuture<void> {
+		return (() => {
+			this.$devicesServices.initialize(undefined, options.device, {skipInferPlatform: true}).wait();
+
+			var action = (device: Mobile.IDevice) =>  { return (() => device.listApplications()).future<void>()(); };
+			this.$devicesServices.execute(action).wait();
+		}).future<void>()();
+	}
+}
+$injector.registerCommand("device|list-applications", ListApplicationsCommand);
--- a/mobile/ios/ios-core.ts
+++ b/mobile/ios/ios-core.ts
@@ -393,7 +393,7 @@ export class CoreFoundation implements  Mobile.ICoreFoundation {
 		return result;
 	}
 
-	public cfTypeFrom(value: {[key: string]: any}): NodeBuffer {
+	public cfTypeFrom(value: IDictionary<any>): NodeBuffer {
 		var keys = _.keys(value);
 		var values = _.values(value);
 
--- a/mobile/ios/ios-core.ts
+++ b/mobile/ios/ios-core.ts
@@ -57,6 +57,7 @@ export class CoreTypes {
 
 	public static am_device_notification_callback = ffi.Function("void", [ref.refType(CoreTypes.am_device_notification_callback_info), CoreTypes.voidPtr]);
 	public static am_device_install_application_callback = ffi.Function("void", [CoreTypes.cfDictionaryRef, CoreTypes.voidPtr]);
+	public static am_device_mount_image_callback = ffi.Function("void", [CoreTypes.voidPtr, CoreTypes.intType]);
 	public static cf_run_loop_timer_callback = ffi.Function("void", [CoreTypes.voidPtr, CoreTypes.voidPtr]);
 }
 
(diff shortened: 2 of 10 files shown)


commands/device/list-applications.ts
```diff
@@ -0,0 +1,20 @@
...
+import options = require("./../../options");
+
+export class ListApplicationsCommand implements ICommand {
+	constructor(private $devicesServices: Mobile.IDevicesServices) { }
+
+	allowedParameters: ICommandParameter[] = [];
+
+	public execute(args: string[]): IFuture<void> {
+		return (() => {
+			this.$devicesServices.initialize(undefined, options.device, {skipInferPlatform: true}).wait();
```
teobugslayer: It seems that the first parameter is not always required, and the second one is declared as optional (```initialize(platform: string, deviceOption?: string, options?: IDevicesServicesInitializationOptions): IFuture<void>;```)

I think we should move **all** parameters into the IDevicesServicesInitializationOptions struct. Opinions?

mobile/ios/ios-core.ts
```diff
@@ -385,7 +394,7 @@ export class CoreFoundation implements  Mobile.ICoreFoundation {
 		return result;
 	}
 
-	private cfTypeFrom(value: {[key: string]: any}): NodeBuffer {
+	public cfTypeFrom(value: {[key: string]: any}): NodeBuffer {
```
teobugslayer: Is this an IStringDictionary?

mobile/ios/ios-core.ts
```diff
@@ -407,16 +424,39 @@ export class CoreFoundation implements  Mobile.ICoreFoundation {
...
+				offset += CoreTypes.pointerSize;
+
+				var jsKey = this.cfTypeTo(keyPointer);
+				var jsValue = this.cfTypeTo(valuePointer);
+				jsDictionary[jsKey] = jsValue;
+			}
+
+			return jsDictionary;
+		} else { // We don't need it for now
+			return "";
```
teobugslayer: Isn't it better to throw an Error here? If we do not handle this case, it is better to catch accidental wrong calls to this function.

mobile/ios/ios-device.ts
```diff
@@ -1,23 +1,42 @@
 ///<reference path="../../../.d.ts"/>
 
+import Future = require("fibers/future");
```
teobugslayer: missing ```use strict```

mobile/ios/ios-device.ts
```diff
@@ -107,8 +126,126 @@ export class IOSDevice implements Mobile.IIOSDevice {
...
+			_.each(supportPaths, (sp: string) => {
+				var parts = sp.split(' ');
+				var version = parts[0];
+				var versionParts = version.split(".");
+
+				var supportPathData = {
+					version: version,
+					majorVersion: versionParts[0],
+					minorVersion: versionParts[1],
+					build: parts.length > 1 ? parts[1].replace("(", "").replace(")", "") : null,
```
teobugslayer: An regexp, perhaps? ```parts[1].replace(/[()]/, () => "")```

mobile/ios/ios-device.ts
```diff
@@ -107,8 +126,126 @@ export class IOSDevice implements Mobile.IIOSDevice {
...
+					path: path.join(developerDiskImagePath, sp)
+				}
+
+				if(supportPathData.majorVersion === productMajorVersion) {
+					if(!supportPath) {
+						supportPath = supportPathData;
+					} else {
+						// is this better than the last match?
+						if(supportPathData.minorVersion === productMinorVersion) {
```
teobugslayer: Can we flatten the if chain a bit? it is really hard to read.

mobile/ios/ios-device.ts
```diff
@@ -107,8 +126,126 @@ export class IOSDevice implements Mobile.IIOSDevice {
...
+	}
+
+	private getDeviceValue(value: string): string {
+		var deviceCopyValue = this.$mobileDevice.deviceCopyValue(this.devicePointer, null, this.$coreFoundation.createCFString(value));
+		return this.$coreFoundation.convertCFStringToCString(deviceCopyValue);
+	}
+
+	private lookupApplications(): IDictionary<any> {
+		this.connect();
+		try {
```
teobugslayer: this pattern -> try / startSession / try / finally / stopSession / finally  / disconnect is used quite often (and even mis-used in the getValue function). Should we extract it in a helper method?

mobile/ios/ios-device.ts
```diff
@@ -107,8 +126,126 @@ export class IOSDevice implements Mobile.IIOSDevice {
...
+				var cfImagePath = this.$coreFoundation.createCFString(imagePath);
+
+				var cfOptions = this.$coreFoundation.cfTypeFrom({
+					ImageType: "Developer",
+					ImageSignature: signature
+				});
+
+				var result = this.$mobileDevice.deviceMountImage(this.devicePointer, cfImagePath, cfOptions, this.mountImageCallbackPtr);
+
+				if(result !== 0 && result !== 3892314230) { // 3892314230 - already mounted
```
teobugslayer: Considering that you already named this constant with a comment, perhaps we can make it official by introducing a constant? After the TypeScript 1.4 PR you can even literally use ```const```.

mobile/ios/ios-proxy-services.ts
```diff
@@ -294,4 +296,47 @@ export class IOSSyslog {
...
+		this.send("vCont;c");
+	}
+
+	private send(packet: string): void {
+		var sum = 0;
+		for(var i=0; i< packet.length; i++) {
+			sum += packet.charCodeAt(i);
+		}
+		sum = sum & 255;
+		var data = util.format("$%s#%s", packet, sum.toString(16));
```
teobugslayer: i think we can extract calculation of the sum into a helper method.

mobile/ios/ios-proxy-services.ts
```diff
@@ -294,4 +296,47 @@ export class IOSSyslog {
...
+		var commands = ['C', 'c', 'S', 's', 'vCont', 'vAttach', 'vRun', 'vStopped', '?'];
+		var stopReply = _.any(commands, command => packet.startsWith(command));
+
+		if(stopReply) {
+			var resume = true;
+			while(resume) {
+				resume = false;
+				// TODO: extend the protocol communication
+			}
+		}
```
teobugslayer: Is the whole **if** statement needed?

mobile/ios/ios-device.ts
Fatme: Yes, we can extract it. But we should pass callback to the new function.

mobile/ios/ios-proxy-services.ts
Fatme: Yes, we  can but I use this logic only here so I left it inside the function

commands/device/list-applications.ts
rosen-vladimirov: I totally agree with this idea (mostly because this way we'll replace the 
```Typescript
{skipInferPlatform: true}
```
code with just a simple property of the object.

mobile/ios/ios-device.ts
teobugslayer: According to Mist guys, this is a well-known and recommended pattern.

mobile/ios/ios-core.ts
Fatme: Currently we are reading all properties for the application, but actually we use only two of them - `Path` and `CFBundleExecutable` (both of them are strings).  We don't need the another properties so we don't need to convert them. If we throw error here we won't be able to execute `device list-applications` command  because it will throw error every time. So I think we shouldn't throw error here.

teobugslayer: Thanks for the explanation!

teobugslayer:
:+1:

mobile/ios/ios-device.ts
Fatme: Unfortunately I could'n simplify this code. If you have any concrete idea I am open to any suggestion.

PR #144, opened 2015-01-20
Share 7zip between CLIs

teobugslayer:
Switch to external 7zip binary to uncompress files. This supports both zip and lzma compression methods.

See http://teampulse.telerik.com/view#item/284635


teobugslayer:
Please do not review this PR, it's work in progress

rosen-vladimirov:
:+1:

PR #147, opened 2015-01-21
Unzip specific files

rosen-vladimirov:
Add method to unzip specific files only. This way you can extract only some files (based on the filter) from any zip archive.

http://teampulse.telerik.com/view#item/284263


teobugslayer:
:+1:

PR #149, opened 2015-01-21
Implement debug command

slavchev:
see https://github.com/NativeScript/nativescript-cli/pull/193
ping @teobugslayer @Fatme

### DIFF
--- a/definitions/mobile.d.ts
+++ b/definitions/mobile.d.ts
@@ -17,6 +17,7 @@ declare module Mobile {
 		deploy(packageFile: string, packageName: string): IFuture<void>;
 		sync(localToDevicePaths: ILocalToDevicePathData[], appIdentifier: IAppIdentifier, liveSyncUrl: string): IFuture<void>;
 		sync(localToDevicePaths: ILocalToDevicePathData[], appIdentifier: IAppIdentifier, liveSyncUrl: string, options: ISyncOptions): IFuture<void>;
+        debug(packageFile: string, packageName: string): IFuture<void>;
 		openDeviceLogStream(): void;
 	}
 
--- a/definitions/mobile.d.ts
+++ b/definitions/mobile.d.ts
@@ -17,7 +17,7 @@ declare module Mobile {
 		deploy(packageFile: string, packageName: string): IFuture<void>;
 		sync(localToDevicePaths: ILocalToDevicePathData[], appIdentifier: IAppIdentifier, liveSyncUrl: string): IFuture<void>;
 		sync(localToDevicePaths: ILocalToDevicePathData[], appIdentifier: IAppIdentifier, liveSyncUrl: string, options: ISyncOptions): IFuture<void>;
-        debug(packageFile: string, packageName: string): IFuture<void>;
+		debug(packageFile: string, packageName: string): IFuture<void>;
 		openDeviceLogStream(): void;
 	}
 
--- a/mobile/android/android-device.ts
+++ b/mobile/android/android-device.ts
@@ -10,6 +10,7 @@ import helpers = require("../../helpers");
 import os = require("os");
 import hostInfo = require("../../host-info");
 import options = require("./../../options");
+import Fiber = require("fibers");
 
 interface IAndroidDeviceDetails {
 	model: string;
@@ -52,14 +53,16 @@ export class AndroidDevice implements Mobile.IDevice {
 	private version: string;
 	private vendor: string;
 	private _tmpRoots: IStringDictionary = {};
-	private _installedApplications: string[];
+    private _installedApplications: string[];
+    private defaultNodeInspectorUrl: string = "http://127.0.0.1:8080/debug";
 
 	constructor(private identifier: string, private adb: string,
 		private $logger: ILogger,
 		private $fs: IFileSystem,
 		private $childProcess: IChildProcess,
 		private $errors: IErrors,
-		private $staticConfig: IStaticConfig) {
+        private $staticConfig: IStaticConfig,
+        private $opener: IOpener) {
 		var details: IAndroidDeviceDetails = this.getDeviceDetails().wait();
 		this.model = details.model;
 		this.name = details.name;
@@ -168,7 +171,7 @@ export class AndroidDevice implements Mobile.IDevice {
 
 	private startPackageOnDevice(packageName: string): IFuture<void> {
 		return (() => {
-			var startPackageCommand = this.composeCommand("shell am start -a android.intent.action.MAIN -n %s/%s", packageName, this.$staticConfig.START_PACKAGE_ACTIVITY_NAME);
+            var startPackageCommand = this.composeCommand("shell am start -a android.intent.action.MAIN -n %s/%s", packageName, this.$staticConfig.START_PACKAGE_ACTIVITY_NAME);
 			var result = this.$childProcess.exec(startPackageCommand).wait();
 			return result[0];
 		}).future<void>()();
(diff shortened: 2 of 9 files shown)


mobile/ios/ios-device.ts
```diff
@@ -136,6 +136,12 @@ export class IOSDevice implements Mobile.IIOSDevice {
 		}).future<void>()();
 	}
 
+    public debug(packageFile: string, packageName: string): IFuture<void> {
+        return (() => {
+            // TODO: add implementation
```
teobugslayer: Consider giving a helpful error message in here: 
```this.$errors.fail({formatStr:"this will come in a future version", suppressCommandHelp: true})``` or similar.

File: mobile/android/android-device.ts [hunk 2]
```diff
@@ -168,7 +171,7 @@ export class AndroidDevice implements Mobile.IDevice {
 
 	private startPackageOnDevice(packageName: string): IFuture<void> {
 		return (() => {
-			var startPackageCommand = this.composeCommand("shell am start -a android.intent.action.MAIN -n %s/%s", packageName, this.$staticConfig.START_PACKAGE_ACTIVITY_NAME);
+            var startPackageCommand = this.composeCommand("shell am start -a android.intent.action.MAIN -n %s/%s", packageName, this.$staticConfig.START_PACKAGE_ACTIVITY_NAME);
```
teobugslayer: Unneeded empty space change

mobile/android/android-device.ts
```diff
@@ -185,9 +188,166 @@ export class AndroidDevice implements Mobile.IDevice {
...
-	}
-
-	private ensureFullAccessPermissions(devicePath: string): IFuture<void> {
+    }
+
+    private tcpForward(src: Number, dest: Number): IFuture<void> {
+        return (() => {
+            var tcpForwardCommand = this.composeCommand("forward tcp:%d tcp:%d", src.toString(), dest.toString());
+            this.$childProcess.exec(tcpForwardCommand).wait();
+        }).future<void>()();
```
teobugslayer: You can simply return the exec() return value and remove the wrappers.

mobile/android/android-device.ts
```diff
@@ -185,9 +188,166 @@ export class AndroidDevice implements Mobile.IDevice {
...
+    }
+
+    private startDebuggerClient(port: Number): IFuture<void> {
+        return (() => {
+            var nodeInspectorModuleFilePath = require.resolve("node-inspector");
+            var nodeInspectorModuleDir = path.dirname(nodeInspectorModuleFilePath);
+            var nodeInspectorFullPath = path.join(path.join(nodeInspectorModuleDir, "bin"), "inspector");
```
teobugslayer: path.join accepts many strings to combine. You can remove one call.

mobile/android/android-device.ts
```diff
@@ -185,9 +188,166 @@ export class AndroidDevice implements Mobile.IDevice {
...
+            this.$childProcess.spawn("node", [nodeInspectorFullPath, "--debug-port", port.toString()], { stdio: ["ignore", "ignore", "ignore"], detached: true });
```
teobugslayer: on ubuntu, node is a time server daemon :) consider using ```process.argv[0]``` instead of hard-coding it.

mobile/android/android-device.ts
```diff
@@ -185,9 +188,166 @@ export class AndroidDevice implements Mobile.IDevice {
...
+        }).future<void>()();
+    }
+
+    private openDebuggerClient(url: string): IFuture<void> {
+        return (() => {
+            this.$opener.open(url, "chrome");
```
teobugslayer: Why "chrome" is hard-coded? What's wrong with using the default browser?

mobile/android/android-device.ts
```diff
@@ -185,9 +188,166 @@ export class AndroidDevice implements Mobile.IDevice {
...
+    }
+
+    private printDebugPort(packageName: string): void {
+        var res = this.$childProcess.spawnFromEvent(this.adb, ["shell", "am", "broadcast", "-a", packageName + "-GetDgbPort"], "exit").wait();
+        this.$logger.info(res.stdout);
+    }
+
+    private attachDebugger(packageName: string): void {
+        var startDebuggerCommand = this.composeCommand("shell am broadcast -a \"%s-Debug\" --ez enable true", packageName);
+        var port = options["debug-port"];
```
teobugslayer: I believe that options.debugPort is easier to the eyes, but the code is correct.

mobile/android/android-device.ts
```diff
@@ -185,9 +188,166 @@ export class AndroidDevice implements Mobile.IDevice {
...
+        if (port > 0) {
```
teobugslayer: what if port is "zzzz"? You should check this in the command validation callback and here just assume that the value is correct. Also, what about if port is 128000?

mobile/android/android-device.ts
```diff
@@ -185,9 +188,166 @@ export class AndroidDevice implements Mobile.IDevice {
...
+            if (match) {
+                port = match[0].substring(7);
+            } else {
+                port = 0;
+            }
+        }
+        if (port > 0) {
+            this.tcpForward(port, port).wait();
+        }
+        this.startDebuggerClient(port).wait();
```
teobugslayer: At this moment, it is possible for port to be 0. Should we call startDebugger()?

mobile/android/android-device.ts
```diff
@@ -185,9 +188,166 @@ export class AndroidDevice implements Mobile.IDevice {
...
+    }
+
+    private detachDebugger(packageName: string): void {
+        var stopDebuggerCommand = this.composeCommand("shell am broadcast -a \"%s-Debug\" --ez enable false", packageName);
+        this.$childProcess.exec(stopDebuggerCommand).wait();
+    }
+
+    private startAppWithDebugger(packageFile: string, packageName: string): void {
+        var uninstallCommand = this.composeCommand("shell pm uninstall \"%s\"", packageName)
+			    this.$childProcess.exec(uninstallCommand).wait();
```
teobugslayer: This code is hard to read. i guess the intent was to create a var and execute a function?

mobile/android/android-device.ts
```diff
@@ -185,9 +188,166 @@ export class AndroidDevice implements Mobile.IDevice {
...
+        if (dbgPort > 0) {
+            this.tcpForward(dbgPort, dbgPort).wait();
+            this.startDebuggerClient(dbgPort).wait();
+            this.openDebuggerClient(this.defaultNodeInspectorUrl + "?port=" + dbgPort).wait();
+        }
+    }
+
+    public debug(packageFile: string, packageName: string): IFuture<void> {
+        return (() => {
+            if (options["get-port"]) {
```
teobugslayer: I believe that it is much better to nuke this function and instead implement compound commands.

mobile/android/android-device.ts
```diff
@@ -185,9 +188,166 @@ export class AndroidDevice implements Mobile.IDevice {
...
+    private checkIfRunning(packageName: string): boolean {
+        var isRunning = this.checkIfFileExists(packageName, "envDebug.out").wait();
+        return isRunning;
+    }
+
+    private readDebugPort(packageName: string): IFuture<number> {
+        return (() => {
+            var res = this.$childProcess.spawnFromEvent(this.adb, ["shell", "cat", "/sdcard/Android/data/" + packageName + "/files/envDebug.out"], "exit").wait();
+            var isRunning = res.stdout.indexOf('yes') > -1;
+            return 1;
```
teobugslayer: Are you sure that the return value is 1 ;) looks like a left-over debugging statement to me.

mobile/android/android-device.ts
```diff
@@ -185,9 +188,166 @@ export class AndroidDevice implements Mobile.IDevice {
...
+        }).future<number>()();
+    }
+
+    private sleep(ms: number): void {
+        var fiber = Fiber.current;
```
teobugslayer: This is the 3rd time we need this function :) You can move it out in common/helpers.ts and use it on all places from there,

mobile/android/android-device.ts
```diff
@@ -185,9 +188,166 @@ export class AndroidDevice implements Mobile.IDevice {
...
+
+    private startAndGetPort(packageName: string): IFuture<number> {
+        return (() => {
+            var port = -1;
+
+            var clearDebugEnvironmentCommand = this.composeCommand('shell rm "/sdcard/Android/data/%s/files/envDebug.in"', packageName);
+            this.$childProcess.exec(clearDebugEnvironmentCommand).wait();
+
+            var isRunning = false;
+            for (var i = 0; i < 10; i++) {
```
teobugslayer: Why 10 and not say 100?

mobile/android/android-device.ts
```diff
@@ -185,9 +188,166 @@ export class AndroidDevice implements Mobile.IDevice {
...
+                this.sleep(1000 /* ms */);
```
teobugslayer: Why 1000 and not say 10000?

definitions/mobile.d.ts
```diff
@@ -17,6 +17,7 @@ declare module Mobile {
 		deploy(packageFile: string, packageName: string): IFuture<void>;
 		sync(localToDevicePaths: ILocalToDevicePathData[], appIdentifier: IAppIdentifier, liveSyncUrl: string): IFuture<void>;
 		sync(localToDevicePaths: ILocalToDevicePathData[], appIdentifier: IAppIdentifier, liveSyncUrl: string, options: ISyncOptions): IFuture<void>;
+        debug(packageFile: string, packageName: string): IFuture<void>;
```
teobugslayer: Consider keeping the source code the same formatting throughout the whole file.

mobile/android/android-device.ts
```diff
@@ -52,14 +53,16 @@ export class AndroidDevice implements Mobile.IDevice {
 	private version: string;
 	private vendor: string;
 	private _tmpRoots: IStringDictionary = {};
-	private _installedApplications: string[];
+    private _installedApplications: string[];
+    private defaultNodeInspectorUrl: string = "http://127.0.0.1:8080/debug";
```
teobugslayer: What happens when port 8080 is already taken?

Fatme: `:string` is not needed. It is inferned from the value

mobile/android/android-device.ts
```diff
@@ -185,9 +188,166 @@ export class AndroidDevice implements Mobile.IDevice {
...
+    private openDebuggerClient(url: string): IFuture<void> {
+        return (() => {
+            this.$opener.open(url, "chrome");
+        }).future<void>()();
+    }
+
+    private printDebugPort(packageName: string): void {
+        var res = this.$childProcess.spawnFromEvent(this.adb, ["shell", "am", "broadcast", "-a", packageName + "-GetDgbPort"], "exit").wait();
+        this.$logger.info(res.stdout);
+    }
```
Fatme: This function should be wrapped with`(() => { }).future<void>()();`

mobile/android/android-device.ts
slavchev: So far only webkit enabled browsers support the debug protocol of node-inspector. On the last meeting we agreed to use Chrome as the most common choice.

mobile/android/android-device.ts
slavchev: I wanted to make the caller of these functions more clear. I find the string manipulation quite distracted when used often. Any suggestions will be appreciated.

mobile/android/android-device.ts
slavchev: The main intent was to emulate polling. I set the timeout for 10 second. These 10 seconds are distributed in a a loop with 1 second wait. Hence, 10x1000 = 10,000 ms = 10 seconds provide some responsiveness (it is not needed to wait all the 10 seconds if there is a response). Observable pattern is the right choice but adb tool does not provide it.

mobile/android/android-device.ts
slavchev: This was the initial implementation. At the end I opted for simple synchronous code which in turn is called from-within Future's.

mobile/ios/ios-device.ts
slavchev: Done with https://github.com/telerik/mobile-cli-lib/commit/1a8d23f43fcd26503d703143ff750445c94f88a8#diff-3efdd98dbb45e19bb229436517332d9dR141

mobile/android/android-device.ts
slavchev: Done with https://github.com/telerik/mobile-cli-lib/commit/1a8d23f43fcd26503d703143ff750445c94f88a8#diff-63f4603786adf475e860b557e6e5ea98R174

mobile/android/android-device.ts
slavchev: Done with https://github.com/telerik/mobile-cli-lib/commit/1a8d23f43fcd26503d703143ff750445c94f88a8#diff-63f4603786adf475e860b557e6e5ea98R193

mobile/android/android-device.ts
slavchev: Done with https://github.com/telerik/mobile-cli-lib/commit/1a8d23f43fcd26503d703143ff750445c94f88a8#diff-63f4603786adf475e860b557e6e5ea98R202

mobile/android/android-device.ts
slavchev: Done with https://github.com/telerik/mobile-cli-lib/commit/1a8d23f43fcd26503d703143ff750445c94f88a8#diff-63f4603786adf475e860b557e6e5ea98R203

mobile/android/android-device.ts
slavchev: Done with https://github.com/telerik/mobile-cli-lib/commit/1a8d23f43fcd26503d703143ff750445c94f88a8#diff-63f4603786adf475e860b557e6e5ea98R220

mobile/android/android-device.ts
slavchev: Done with https://github.com/telerik/mobile-cli-lib/commit/1a8d23f43fcd26503d703143ff750445c94f88a8#diff-63f4603786adf475e860b557e6e5ea98R233

mobile/android/android-device.ts
slavchev: Done with https://github.com/telerik/mobile-cli-lib/commit/1a8d23f43fcd26503d703143ff750445c94f88a8#diff-63f4603786adf475e860b557e6e5ea98R233

mobile/android/android-device.ts
slavchev: Done with https://github.com/telerik/mobile-cli-lib/commit/1a8d23f43fcd26503d703143ff750445c94f88a8#diff-63f4603786adf475e860b557e6e5ea98L295

mobile/android/android-device.ts
slavchev: Done with https://github.com/telerik/mobile-cli-lib/commit/1a8d23f43fcd26503d703143ff750445c94f88a8#diff-b823ab15d1844ca2c6b85896ef654162R201 and https://github.com/telerik/mobile-cli-lib/commit/1a8d23f43fcd26503d703143ff750445c94f88a8#diff-63f4603786adf475e860b557e6e5ea98R316

mobile/android/android-device.ts
slavchev: See https://github.com/telerik/mobile-cli-lib/pull/149#discussion_r23315534

definitions/mobile.d.ts
slavchev: Done with https://github.com/telerik/mobile-cli-lib/commit/1a8d23f43fcd26503d703143ff750445c94f88a8#diff-c8d8b75ca33f89ad4020410820a79a5cR20

mobile/android/android-device.ts
slavchev: @teobugslayer As discussed earlier the port is hard coded for this version
@Fatme done with https://github.com/telerik/mobile-cli-lib/commit/1a8d23f43fcd26503d703143ff750445c94f88a8#diff-63f4603786adf475e860b557e6e5ea98R57

mobile/android/android-device.ts
teobugslayer: ping @ikoevska - probably we should add this requirement in the online help as well?

mobile/android/android-device.ts
```diff
@@ -185,9 +188,152 @@ export class AndroidDevice implements Mobile.IDevice {
...
+            var isRunning = false;
+            for (var i = 0; i < 10; i++) {
+                helpers.sleep(1000 /* ms */);
+                isRunning = this.checkIfRunning(packageName);
+                if (isRunning)
+                    break;
+            }
+
+            if (isRunning) {
+                var setEnvironmentCommand = this.composeCommand("shell \"cat /dev/null > \"/sdcard/Android/data/%s/files/envDebug.in\"\"", packageName);
```
lubogospod: You can extract path "/sdcard/Android/data/%s/files/" as constant/variable, and use it is all references of this path.

teobugslayer:
:+1:  after addressing @lubogospod's comment

mobile/android/android-device.ts
slavchev: Done with https://github.com/telerik/mobile-cli-lib/commit/ca5bd454109f9a89c7587b298005ff7046bdbfb7#diff-63f4603786adf475e860b557e6e5ea98R46

slavchev:
ping @ErjanGavalji

PR #153, opened 2015-01-26
Run ios apps on ios device on windows

Fatme:
http://teampulse.telerik.com/view#item/283764


mobile/ios/ios-core.ts
```diff
@@ -683,35 +708,40 @@ class WinSocket implements Mobile.IiOSDeviceSocket {
...
+				type = "string";
+				value = values[i];
+			}
 
-			plist.value[keys[i]] = {type: type, value: value};
+			plistData.value[keys[i]] = {type: type, value: value};
 		}
 
-		this.$logger.trace("created plist: '%s'" + plist);
+		this.$logger.trace("created plist: '%s'" + plistData);
```
rosen-vladimirov: is the + correct, shouldn't it be:
```Typescript
this.$logger.trace("created plist: '%s'", plistData);
```

options.ts
```diff
@@ -22,15 +22,15 @@ var knownOpts: any = {
...
-        //
-        "debug-brk": Boolean,
-        "debugBrk": Boolean,
-        "debug-port": Number,
-        "debugPort": Number,
-        "get-port": Boolean,
-        "getPort": Boolean,
-        "start": Boolean,
-        "stop": Boolean
+		"debug-brk": Boolean,
```
rosen-vladimirov: indentation issue

mobile/ios/ios-core.ts
```diff
@@ -543,7 +545,11 @@ export class MobileDevice implements Mobile.IMobileDevice {
 	}
 
 	public deviceMountImage(devicePointer: NodeBuffer, imagePath: NodeBuffer, options: NodeBuffer, mountCallBack: NodeBuffer): number {
-		return this.mobileDeviceLibrary.AMDeviceMountImage(devicePointer, imagePath, options, mountCallBack, null);
+		if(hostInfo.isDarwin()) {
+			return this.mobileDeviceLibrary.AMDeviceMountImage(devicePointer, imagePath, options, mountCallBack, null);
+		}
+
+		this.$errors.fail("AMDeviceMountImage is exported only on Darwin OS");
```
teobugslayer: I believe that this error message is baffling to the user. Consider giving a more helpful user message.

rosen-vladimirov:
:+1:  after fixing the minor issues

mobile/ios/ios-core.ts
```diff
@@ -652,16 +648,45 @@ class WinSocket implements Mobile.IiOSDeviceSocket {
 		this.close();
 	}
 
-	public receiveMessage(): IFuture<void> {
-		return(() => {
-			var reply = this.receiveMessageHelper();
-			if(!this.$mobileDevice.isDataReceivingCompleted(reply)) {
-				reply = this.receiveMessageHelper();
+	public receiveMessage(): IFuture<Mobile.IiOSSocketResponseData> {
```
teobugslayer: Why do we need to return a future?

mobile/ios/ios-core.ts
```diff
@@ -652,16 +648,45 @@ class WinSocket implements Mobile.IiOSDeviceSocket {
...
+		}).future<Mobile.IiOSSocketResponseData>()();
+	}
+
+	public sendMessage(data: any): void {
+		var payload: NodeBuffer = new Buffer(plistlib.toString(this.createPlist(data)));
+		var packed: any = bufferpack.pack(">i", [payload.length]);
+		var message = Buffer.concat([packed, payload]);
+
+		var writtenBytes = this.sendCore(message);
+		this.$logger.debug("WinSocket-> sending message: '%s', written bytes: '%s'", message.toString(), writtenBytes.toString());
```
teobugslayer: Is this function callable on Mac? If yes, does "WinSocket" apply?

mobile/ios/ios-core.ts
```diff
@@ -683,35 +708,40 @@ class WinSocket implements Mobile.IiOSDeviceSocket {
...
+		this.$logger.debug("WinSocket-> sendCore: writtenBytes '%s'", writtenBytes);
+		return writtenBytes;
 	}
 
-	private createPlist(data: {[key: string]: {}}) : {} {
+	private createPlist(data: IDictionary<any>) : {} {
 		var keys = _.keys(data);
 		var values = _.values(data);
-		var plist: {type:string; value:any} = {type: "dict", value: {}} ;
+		var plistData: {type:string; value:any} = {type: "dict", value: {}};
```
teobugslayer: Is the type IDictionary\<any\>?

mobile/ios/ios-core.ts
Fatme: No, it is not callable on Mac. We have PosixSocket implementation for OSX

mobile/ios/ios-core.ts
Fatme: WinSocket and PosixSocket classes implament the same interface. The `receiveMessage` functon in PosixSocket must return feature so we need to return feature in WinSocket too.

mobile/ios/ios-core.ts
Fatme: The message should never reach to the user. I added it only if someone call deviceMountImage function on Windows by mistake.

mobile/ios/ios-core.ts
Fatme: No, it is not `IDictionary<any>` because we have `value`

teobugslayer:
:+1:

PR #156, opened 2015-01-27
Fix checking for Android SDK

teobugslayer:
Fix error handling on process spawn

Fixes http://teampulse.telerik.com/view#item/279833

### DIFF
--- a/child-process.ts
+++ b/child-process.ts
@@ -62,7 +62,7 @@ export class ChildProcess implements IChildProcess {
 		}
 
 		childProcess.on(event, (arg: any) => {
-			var exitCode = typeof arg == 'number' ? arg : arg && arg.code;
+			var exitCode = typeof arg === 'number' ? arg : arg && arg.code;
 			var result = {
 				stdout: capturedOut,
 				stderr: capturedErr,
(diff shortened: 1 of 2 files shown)


child-process.ts
```diff
@@ -79,11 +79,17 @@ export class ChildProcess implements IChildProcess {
...
 						errorMessage += util.format(' Error output: \n %s', capturedErr);
 					}
-					future.throw(errorMessage);
+					future.throw(new Error(errorMessage));
 				}
 			}
 		});
 
+		if (event !== "error") {
+			childProcess.on("error", (err: Error) => {
```
Fatme: maybe `childProcess.once("error")`

child-process.ts
```diff
@@ -79,11 +79,17 @@ export class ChildProcess implements IChildProcess {
 					if (capturedErr) {
 						errorMessage += util.format(' Error output: \n %s', capturedErr);
 					}
-					future.throw(errorMessage);
+					future.throw(new Error(errorMessage));
```
Fatme: Why we need `new Error`?

child-process.ts
teobugslayer: Sure!

child-process.ts
teobugslayer: Because otherwise we throw string, which is allowed by the language but not a good practice.

Fatme:
:+1:

PR #157, opened 2015-01-27
Add fail without printing help

rosen-vladimirov:
Add failWithoutHelp method to Errors. It should be used when you want to break the execution without printing help.


teobugslayer:
:+1: Now we have to use it in our entire code base!

PR #159, opened 2015-01-29
Remove adbFilePath and sevenZipFilePath from staticConfig

Fatme:
We don't  need adbFilePath and sevenZipFilePath to be part of IStaticConfig interface anymore because they are already part of common lib.


teobugslayer:
:+1:

Fatme:
I rebased the PR but merge button is still disabled. I'll cherypick the commit in another PR.

PR #161, opened 2015-01-29
Don't try to convert the message if the format used for communication is xml

Fatme:
Fixing deploy on ios device on windows OS


teobugslayer:
:+1:

PR #162, opened 2015-01-29
Add check if 7za can be executed on Linux

rosen-vladimirov:
If user is on x64 Linux and 32 bit libraries, describe in our documentation, are not installed, user will not be able to use 7zip. This means - unable to create projects, unable to build etc. So on post-install we check if we are able to execute the 7za file and if ENOENT is return we show the user where he can find system requirements.

Fix host-info paths in some files and remove host-info reference from some other files.

http://teampulse.telerik.com/view#item/285027

### DIFF
--- a/commands/post-install.ts
+++ b/commands/post-install.ts
@@ -1,16 +1,37 @@
 ///<reference path="../../.d.ts"/>
 "use strict";
 
-import options = require("../../common/options");
+import options = require("../options");
+import hostInfo = require("../host-info");
 
 export class PostInstallCommand implements ICommand {
+	private static MISSING_LINUX_32BITS_LIBS_MESSAGE = "It looks like you are using x64 Linux OS, but you are missing some of the required 32bits libraries." +
+		"You can find all system requirements on http://docs.telerik.com/platform/appbuilder/running-appbuilder/running-the-cli/system-requirements-cli#linux-systems";
+
 	constructor(private $autoCompletionService: IAutoCompletionService,
 		private $fs: IFileSystem,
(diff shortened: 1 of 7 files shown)


host-info.ts
```diff
@@ -14,13 +14,17 @@ export function isWindows32() {
...
-	return process.platform.toUpperCase() === "DARWIN";
+	return process.platform === "darwin";
 }
 
 export function isLinux() {
 	return process.platform === "linux";
 }
 
+export function isLinux64(): boolean {
+	return isLinux() && process.arch === "x64";
```
teobugslayer: this function will not work with 32-bit node on 64-bit linux

commands/post-install.ts
```diff
@@ -1,17 +1,38 @@
...
+		return (() => {
+			try {
+				var proc = this.$childProcess.spawnFromEvent(this.$staticConfig.sevenZipFilePath, ["-h"], "exit", undefined, { throwError: false }).wait();
+
+				if(proc.stderr) {
+					this.$errors.fail({ formatStr: PostInstallCommand.MISSING_LINUX_32BITS_LIBS_MESSAGE, suppressCommandHelp: true });
+				}
+			} catch(e) {
+				var message: string = (e.code === "ENOENT") ? PostInstallCommand.MISSING_LINUX_32BITS_LIBS_MESSAGE : e.message;
+				this.$errors.fail({ formatStr: message, suppressCommandHelp: true });
```
teobugslayer: failWithoutHelp?

rosen-vladimirov:
closing and will open against release branch

PR #168, opened 2015-02-02
Fix run command

Fatme:
Fixing the issues from this story http://teampulse.telerik.com/view#item/283764

### DIFF
--- a/mobile/ios/ios-device.ts
+++ b/mobile/ios/ios-device.ts
@@ -70,10 +70,6 @@ export class IOSDevice implements Mobile.IIOSDevice {
 		return "Apple";
 	}
 
-	public getInstalledApplications(): IFuture<string[]> {
-		throw new Error("Not implemented");
-	}
-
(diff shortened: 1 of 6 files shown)


mobile/ios/ios-device.ts
```diff
@@ -357,9 +353,13 @@ export class IOSDevice implements Mobile.IIOSDevice {
...
 	}
 
-	public listApplications(): void {
-		var applications = this.lookupApplications();
-		_(_.sortBy(_.keys(applications))).each((bundleId: string) => this.$logger.info(bundleId));
+	public getInstalledApplications():  IFuture<string[]> {
+		return (() => {
+			var applications = this.lookupApplications();
+			var keys = _.keys(applications);
+			return _.sortBy(keys);
```
teobugslayer: Untested, but perhaps:

``` return _(this.lookupApplications()).keys().sortBy().value();```

teobugslayer:
Good job, :+1:

PR #169, opened 2015-02-02
Convert js callstack to ts

Fatme:
1. we should not fail in the catch clause, because that would stop further executing of this function, and the program would exit with the wrong exit code
2. We cannot use this.$xxx resolve, because this is a free-standing function. The old code would throw.


teobugslayer:
:+1:

PR #170, opened 2015-02-03
Fix genymotion check

rosen-vladimirov:
When you specify --geny option of $ appbuilder emulate android command, we check for genyshell in the PATH environment variable. As on MacOS genyshell and the real program that we need for genymotion (player) are in different directories, we fail with incorrect message.

Fixes http://teampulse.telerik.com/view#item/285302

### DIFF
--- a/mobile/android/android-emulator-services.ts
+++ b/mobile/android/android-emulator-services.ts
@@ -63,7 +63,7 @@ class AndroidEmulatorServices implements Mobile.IEmulatorPlatformServices {
 				}
 			} catch (e) {
 				var message: string = (e.code === "ENOENT") ? AndroidEmulatorServices.MISSING_SDK_MESSAGE : e.message;
-				this.$errors.fail({formatStr: message, suppressCommandHelp: true});
+				this.$errors.failWithoutHelp(message);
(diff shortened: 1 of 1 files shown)


nmladenov:
:+1:

mobile/android/android-emulator-services.ts
```diff
@@ -71,12 +71,13 @@ class AndroidEmulatorServices implements Mobile.IEmulatorPlatformServices {
 	private checkGenymotionConfiguration(): IFuture<void> {
 		return (() => {
 			try {
-				var proc = this.$childProcess.spawnFromEvent("genyshell", ["-h"], "exit", undefined, { throwError: false }).wait();
+				var proc = this.$childProcess.spawnFromEvent("player", ["-h"], "exit", undefined, { throwError: false }).wait();
```
teobugslayer: This is invalid command line - probably we should remove `-h` parameter

rosen-vladimirov: yep, I'll remove it :)

teobugslayer:
:+1: after addressing my minor comment.

PR #172, opened 2015-02-03
Fix incorrect message for genymotion

rosen-vladimirov:
If Genymotion is not configured properly, the error message is not correct. Using correct constant.


teobugslayer:
:+1:

PR #173, opened 2015-02-04
Throw error if device run command is executed on windows

Fatme:


mobile/ios/ios-device.ts
```diff
@@ -365,6 +365,10 @@ export class IOSDevice implements Mobile.IIOSDevice {
 
 	public runApplication(applicationId: string): IFuture<void> {
 		return (() => {
+			if(hostInfo.isWindows()) {
+				this.$errors.fail("$appbuilder device run command is not supported on Windows for iOS devices.");
```
teobugslayer: Should we print the command help here? If no,, consider using failWithoutHelp

Fatme: I think we should.

teobugslayer:
:+1:

PR #174, opened 2015-02-06
Throw error if run command is executed and more than one device found

Fatme:


commands/device/run-application.ts
```diff
@@ -15,6 +16,10 @@ export class RunApplicationOnDeviceCommand implements ICommand {
 		return (() => {
 			this.$devicesServices.initialize({ deviceId: options.device, skipInferPlatform: true }).wait();
 
+			if (this.$devicesServices.deviceCount > 1) {
+				this.$errors.fail("More than one device found. Specify device explicitly with --device option.");
```
teobugslayer: Maybe also add - "To discover device ID, use $appbuilder device"

teobugslayer:
:+1: for the code, but please consider my remark about the help message.

tytodorov:
:+1: With Tosho's remark

PR #176, opened 2015-02-06
Update Lodash definitions, createTable helper

rosen-vladimirov:
Update Lodash definitions to allow chaining of .values method. Add createTable helper method which creates cli-table from passed headers and data.

Fixes are required for this issue: http://teampulse.telerik.com/view#item/285533


teobugslayer:
:+1:

tytodorov:
:+1:

PR #181, opened 2015-02-10
Remove default value for START_PACKAGE_ACTIVITY

rosen-vladimirov:
The default value doesn't allow us to use constructions like public get START_PACKAGE_ACTIVITY_NAME(): string { ... } in each CLI's. We need such construction in appbuilder-cli to support different start package activity names, based on the project type.


Fatme:
:+1:

tytodorov:
:+1:

PR #183, opened 2015-02-10
Fix exception on double command arguments

teobugslayer:
Fix an error message to work in TNS

Fixes http://teampulse.telerik.com/view#item/280191


rosen-vladimirov:
:+1:

teobugslayer:
@fatme, I would suggest using "this is an array" and the command does split parameters on " ".

teobugslayer:
Closing this PR - I will re-implement it in another PR

PR #187, opened 2015-02-11
Add option for case-insensitive unzip for 7zip

Mitko-Kerezov:
Related to http://teampulse.telerik.com/view#item/285465
7zip is case-sensitive by default

### DIFF
--- a/file-system.ts
+++ b/file-system.ts
@@ -56,14 +56,19 @@ export class FileSystem implements IFileSystem {
 		return result;
 	}
 
-	public unzip(zipFile: string, destinationDir: string, options?: { overwriteExisitingFiles: boolean}, fileFilters?: string[]): IFuture<void> {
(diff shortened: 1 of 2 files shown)


teobugslayer:
:+1:

file-system.ts
```diff
@@ -56,14 +56,19 @@ export class FileSystem implements IFileSystem {
 		return result;
 	}
 
-	public unzip(zipFile: string, destinationDir: string, options?: { overwriteExisitingFiles: boolean}, fileFilters?: string[]): IFuture<void> {
+	public unzip(zipFile: string, destinationDir: string, options?: { overwriteExisitingFiles?: boolean; caseSensitive?: boolean}, fileFilters?: string[]): IFuture<void> {
 		return (() => {
 			var shouldOverwriteFiles = options ? options.overwriteExisitingFiles : true;
+			var isCaseSensitive = options ? options.caseSensitive : true;
```
Fatme: It should be `var isCaseSensitive = options && options.caseSensitive ? options.caseSensitive : true;`

PR #188, opened 2015-02-11
Bail out with error message when a parameter on the command line is give...

teobugslayer:
...n more than once.

Fixes http://teampulse.telerik.com/view#item/280191

### DIFF
--- a/declarations.d.ts
+++ b/declarations.d.ts
@@ -111,6 +111,7 @@ interface IErrors {
 	failWithoutHelp(message: string, ...args: any[]): void;
 	beginCommand(action: () => IFuture<boolean>, printCommandHelp: () => IFuture<boolean>): IFuture<boolean>;
 	verifyHeap(message: string): void;
+	executeAction(action: Function): any;
 }
 
 interface ICommandOptions {
--- a/errors.ts
+++ b/errors.ts
@@ -1,8 +1,10 @@
 ///<reference path="../.d.ts"/>
+"use strict";
 
 import util = require("util");
 import path = require("path");
 
+// we need this to overwrite .stack property (read-only in Error)
 function Exception() {}
 Exception.prototype = new Error();
 
@@ -63,10 +65,11 @@ export function installUncaughtExceptionListener(): void {
 		console.log(callstack || err.toString());
 
 		try {
-		 	var analyticsService = $injector.resolve("analyticsService");
+			var analyticsService = $injector.resolve("analyticsService");
 			analyticsService.trackException(err, callstack);
 		} catch (e) {
-			$injector.resolve("$logger").error("Error while reporting exception: " + e);
+			// Do not replace with logger due to cyclic dependency
+			console.log("Error while reporting exception: " + e);
 		}
 
 		process.exit(ErrorCodes.UNKNOWN);
@@ -76,7 +79,8 @@ export function installUncaughtExceptionListener(): void {
 export class Errors implements IErrors {
 	constructor(
 		private $logger: ILogger,
-		private $config: Config.IConfig) {}
+		private $config: Config.IConfig) {
+	}
 
 	fail(optsOrFormatStr: any, ...args: any[]): void {
 		var opts = optsOrFormatStr;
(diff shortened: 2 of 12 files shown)


rosen-vladimirov:
:+1:

helpers.ts
```diff
@@ -90,27 +80,25 @@ export function getParsedOptions(options: any, shorthands: any, clientName: stri
 
 export function validateYargsArguments(parsed: any, knownOpts: any, shorthands: any, clientName?: string): void {
 	if(path.basename(process.argv[1]).indexOf(clientName) !== -1) {
+		var errors = $injector.resolve("$errors");
```
rosen-vladimirov: It looks like errors cannot be resolved here. The problem is that bootstrap is resolved before starting the fiber and before the injector is created (from appbuilder-cli.ts: 
```Typescript
require("./bootstrap");
import fiberBootstrap = require("./fiber-bootstrap");

fiberBootstrap.run(() => {
```
So when options is required in bootstrap, it calls helpers.getParsedOptions, which calls this method. We try to resolve errors, which depends on logger and config. Configuration's constructor has wait, so we receive can't wait without a fiber as we are not in fiberBootstrap's run.

rosen-vladimirov: disregard this comment after latest fixes

errors.ts
```diff
@@ -66,7 +67,7 @@ export function installUncaughtExceptionListener(): void {
 		 	var analyticsService = $injector.resolve("analyticsService");
```
rosen-vladimirov: Ok, it's kind of fun here. In case you pass incorrect -- option (for example $ appbuilder --gosho) we fail when bootstrap is requiring options.ts from appbuilder-cli. As the "require" is not finished, when we resolve analyticsService here (which has "import options = require("../options");" on line 7, we try to parse options again and we throw error once again. So the result is that our error for incorrect option is catched, after that we try resolving analyticsService we throw it again and we catch it two lines below.
The output is:
Error: The option 'gosho' is not supported. To see command's options, use '$ appbuilder emulate --help'. To see all commands use '$ appbuilder help'.
    at Errors.fail (D:\Work\icenium-cli\lib\common\errors.ts:98:68)
    at Errors.failWithoutHelp (D:\Work\icenium-cli\lib\common\errors.ts:107:7)
    at <anonymous> (D:\Work\icenium-cli\lib\common\helpers.ts:88:11)
    at Function.forEach (D:\Work\icenium-cli\node_modules\lodash\dist\lodash.js:3297:15)
    at validateYargsArguments (D:\Work\icenium-cli\lib\common\helpers.ts:84:4)
    at Object.getParsedOptions (D:\Work\icenium-cli\lib\common\helpers.ts:77:1)
    at Object.<anonymous> (D:\Work\icenium-cli\lib\options.ts:48:21)
    at Module._compile (module.js:456:26)
    at Object.Module._extensions..js (module.js:474:10)
    at Module.load (module.js:356:32)
Error while reporting exception: Exception: The option 'gosho' is not supported. To see command's options, use '$ appbuilder emulate --help'. To see all commands use '$ appbuilder help'.

declarations.d.ts
```diff
@@ -109,8 +109,9 @@ interface IErrors {
 	fail(formatStr: string, ...args: any[]): void;
 	fail(opts: {formatStr?: string; errorCode?: number; suppressCommandHelp?: boolean}, ...args: any[]): void;
 	failWithoutHelp(message: string, ...args: any[]): void;
-	beginCommand(action: () => IFuture<boolean>, printCommandHelp: () => IFuture<boolean>): IFuture<boolean>;
+	beginCommand(action: () => IFuture<boolean>, printCommandHelp?: () => IFuture<boolean>): IFuture<boolean>;
```
rosen-vladimirov: I think the second parameter should remain mandatory

errors.ts
```diff
@@ -63,10 +65,10 @@ export function installUncaughtExceptionListener(): void {
 		console.log(callstack || err.toString());
 
 		try {
-		 	var analyticsService = $injector.resolve("analyticsService");
+			var analyticsService = $injector.resolve("analyticsService");
 			analyticsService.trackException(err, callstack);
 		} catch (e) {
-			$injector.resolve("$logger").error("Error while reporting exception: " + e);
+			console.log("Error while reporting exception: " + e);
```
rosen-vladimirov: maybe a comment - "Do not replace with logger due to cyclic dependency and errors"

errors.ts
```diff
@@ -112,7 +115,7 @@ export class Errors implements IErrors {
 					? resolveCallStack(ex.stack)
 					: "\x1B[31;1m" + ex.message + "\x1B[0m");
 
-				if (!ex.suppressCommandHelp) {
+				if (printCommandHelp && !ex.suppressCommandHelp) {
```
rosen-vladimirov: I think this change is not needed, we can use printCommandHelp as mandatory parameter.

rosen-vladimirov:
:+1:  but I think we should rollback change in beginCommand parameters, there's no need to change them.

PR #189, opened 2015-02-11
Fix command params validation

Fatme:


rosen-vladimirov:
:+1:  but why there are two commits :D

teobugslayer:
:+1:

PR #192, opened 2015-02-12
Fix usage on Linux

rosen-vladimirov:
Make sure iosDeviceDiscovery will not make commands fail on Linux. On Linux we MUST use IOSDeviceDiscoveryStub as we cannot execute iosDevice related commands.


teobugslayer:
:+1:

Fatme:
:+1:

PR #194, opened 2015-02-13
Fix cyclic dependency in errors

teobugslayer:


errors.ts
```diff
@@ -77,17 +77,15 @@ export function installUncaughtExceptionListener(): void {
...
 
 export class Errors implements IErrors {
-	constructor(
-		private $logger: ILogger,
-		private $config: Config.IConfig) {
-	}
+
+	_printCallStack: boolean = false;
+	public get printCallStack(): boolean { return this._printCallStack; }
+	public set printCallStack(newVal: boolean) { this._printCallStack = newVal; }
```
Fatme: Do we need `get` and `set`. Just `public printCallStack = false` should be enough

teobugslayer: Yeah, I guess, the C# developer has spoken in me :)

Fatme:
:+1:

PR #195, opened 2015-02-13
Remove future.wrap. Do not use it ever!

teobugslayer:


file-system.ts
```diff
@@ -10,11 +10,8 @@ import rimraf = require("rimraf");
 import hostInfo = require("./host-info");
 
 export class FileSystem implements IFileSystem {
-	private _stat = Future.wrap(fs.stat);
-	private _readFile = Future.wrap(fs.readFile);
-	private _writeFile = Future.wrap<void>(fs.writeFile);
-	private _readdir = Future.wrap(fs.readdir);
-	private _chmod = Future.wrap(fs.chmod);
+
```
rosen-vladimirov: blank lines

rosen-vladimirov:
awesome :+1:

file-system.ts
```diff
@@ -151,15 +148,47 @@ export class FileSystem implements IFileSystem {
...
+			}
+		});
+		return future;
 	}
 
-	public readText(filename: string, encoding?: string): IFuture<string> {
-		return <IFuture<string>> <any> this._readFile(filename, { encoding: encoding || "utf8" });
+	public readText(filename: string, options?: any): IFuture<string> {
+		options = options || { encoding: "utf8" };
+		if (_.isString(options)) {
```
Fatme: Can we simplify this? I suggest to introduce interface `IFsReadTextOptions: { encoding: string }`. This way this check will be removed.

Fatme: I also suggest to rename `options` to `opts` for example, because we may have `import options = require("options")`

file-system.ts
```diff
@@ -175,7 +204,15 @@ export class FileSystem implements IFileSystem {
...
-			this._writeFile(filename, data, { encoding: encoding }).wait();
+			var future = new Future<void>();
+			fs.writeFile(filename, data, { encoding: encoding }, (err: Error) => {
+				if(err) {
+					future.throw(err);
+				} else {
+					future.return();
+				}
+			});
+			future.wait();
```
Fatme: You can remove `(() => { }).future<void>()()` and just `return future`

Fatme: Ignore the comment! We can't do it!

rosen-vladimirov: we have wait for createDirectory. that's why wrapping future is used

PR #197, opened 2015-02-13
Remove watch and replace it with gaze module

Fatme:


teobugslayer:
:+1: and it is absolutely lovely change :)

PR #198, opened 2015-02-16
Fix code to work with lodash 3.2

Fatme:


mobile/android/android-emulator-services.ts
```diff
@@ -327,10 +327,9 @@ class AndroidEmulatorServices implements Mobile.IEmulatorPlatformServices {
 		return (() => {
 			var minVersion = this.$emulatorSettingsService.minVersion;
 
-			var best =_(this.getAvds().wait())
+			var best = _.chain(this.getAvds().wait())
```
teobugslayer: Is _.chain() needed?

teobugslayer:
:+1: after addressing my minor comment

mobile/android/android-emulator-services.ts
rosen-vladimirov: Yes, because there's problem with lodash definitions. After we fix them, _.chain will not be needed.

PR #200, opened 2015-02-17
Do not show "null" message on Linux

rosen-vladimirov:
When you do not have connected devices and you execute $appbuilder device on Linux, we print:
null
Cannot find connected ...

Remove printing of error in case there's no error.

### DIFF
--- a/mobile/mobile-core/device-discovery.ts
+++ b/mobile/mobile-core/device-discovery.ts
@@ -139,22 +139,28 @@ class IOSDeviceDiscovery extends DeviceDiscovery {
 
 class IOSDeviceDiscoveryStub extends DeviceDiscovery {
 	constructor(private $logger: ILogger,
+		private $staticConfig: IStaticConfig,
 		private error: string) {
 		super();
 	}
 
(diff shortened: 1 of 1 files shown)


mobile/mobile-core/device-discovery.ts
```diff
@@ -144,7 +144,10 @@ class IOSDeviceDiscoveryStub extends DeviceDiscovery {
 	}
 
 	public startLookingForDevices(): IFuture<void> {
-		this.$logger.warn(this.error);
+		if(this.error) {
```
Fatme: Maybe we need to print something like `ios releated commands are not supported on Linux` if the os isLinux

rosen-vladimirov: but if you execute  $ appbuilder device, you do not have any idea, that we'll execute ios related code and for me, such warning will seem strange

rosen-vladimirov: done :)

Fatme:
:+1:

PR #201, opened 2015-02-17
Remove bad try-catch

teobugslayer:
The code inside it calls this.$errors.fail and expects the program execution to stop.
Due to our implementation, the catch will prevent this from working and the program will continue executing in a bad state.


Fatme:
:+1:

PR #202, opened 2015-02-17
Fail the program even inside try / catch

teobugslayer:
with current implementation, one can write try { $errors.fail() } catch {} and the program will continie execution in bad state
this ensures that even when the program continues, at least on the next event loop cycle it will terminate
this does not resolve the issue that after the catch and before the next tick, there is plenty of code which can corrupt user data
but is better than nothing


teobugslayer:
ping @ligaz @tailsu

teobugslayer:
This PR is obsolete because of the future work integrating the CLI and Mist.

PR #203, opened 2015-02-17
Remove unneeded chmod on user's machine on package install

teobugslayer:
The package.tgz contains proper unix file attributes so we do not need to explicitly set them any more.


Fatme:
:+1:

PR #205, opened 2015-02-18
Infrastructure for the NativeScript library add command

fealebenpae:


child-process.ts
```diff
@@ -18,7 +18,13 @@ export class ChildProcess implements IChildProcess {
 			} else {
 				future.return(stdout);
 			}
-		});
+		};
+
+		if (typeof options === 'object') {
```
Fatme: `if(options) { .. } is shorter`

file-system.ts
```diff
@@ -296,6 +296,12 @@ export class FileSystem implements IFileSystem {
 		}).future<boolean>()();
 	}
 
+	public isRelativePath(p: string): boolean {
+		var normal = path.normalize(p);
+		var absolute = path.resolve(p);
+		return normal != absolute;
```
Fatme: `normal !== absolute`

Fatme:
:+1: after checking my minor suggestions

PR #206, opened 2015-02-19
Fix unzip

rosen-vladimirov:
We should NOT overwrite existing files only when overwriteExisitingFiles is set to false. In any other cases files should be replaced.
We should work with case insensitive files only when caseSensitive is false.


teobugslayer:
nice catch, :+1:

PR #209, opened 2015-02-23
Support multilevel hierarchical commands

rosen-vladimirov:
Add support for multilevel hierarchical commands. This way we can use commands like $ appbuilder appmanager upload android declared as "appmanager|upload|android".
Add support for passing multiple arguments to dynamic calls, by changing regular expressions.
Add DynamicHelpService which is used to check different conditions when printing our help.
Use _.isBoolean where possible. Add dynamicHelpProvider which should be implemented by each CLI.
Use lodash _.template as it uses microTemplate internally and pass the help through it. This way we can write javascript directly in our help.txt.
Replace of all entries in the help content (when executing), which match our dynamicCallRegex with this.$injector.dynamicCall($1).wait()
This way you can call dynamic methods, by just using:
\<%= #{cordovaProject.projectTemplatesString} %\>
Add localVariables that allow easy call for default scenarios (isLinux, isMacOS, is Windows, isCordova, isNativeScript, isMobileWebsite).
In order to use it, just call:
\<% if (isCordova) %\>

http://teampulse.telerik.com/view#item/283188

### DIFF
--- a/commands/help.ts
+++ b/commands/help.ts
@@ -3,17 +3,23 @@
 
 import path = require("path");
 import util = require("util");
+import os = require("os");
 import commandParams = require("../command-params");
-
+import Future = require("fibers/future");
 export class HelpCommand implements ICommand {
 	constructor(private $logger: ILogger,
 		private $injector: IInjector,
 		private $errors: IErrors,
 		private $fs: IFileSystem,
-		private $staticConfig: Config.IStaticConfig) {}
+		private $staticConfig: Config.IStaticConfig,
+		private $microTemplateService: IMicroTemplateService) { }
 
 	public enableHooks = false;
-	public allowedParameters: ICommandParameter[] = [new commandParams.StringCommandParameter(this.$injector), new commandParams.StringCommandParameter(this.$injector)];
+	public canExecute(args: string[]): IFuture<boolean> {
+		return Future.fromResult(true);
+	}
+
+	public allowedParameters: ICommandParameter[] = [];
 
 	public execute(args: string[]): IFuture<void> {
 		return (() => {
@@ -22,8 +28,9 @@ export class HelpCommand implements ICommand {
 				topic = "";
 			}
 
-			if(args[1]) {
-				topic = util.format("%s|%s", args[0], args[1]);
+			var hierarchicalCommand = this.$injector.buildHierarchicalCommand(args[0], _.rest(args));
+			if(hierarchicalCommand) {
+				topic = hierarchicalCommand.commandName;
 			}
 
 			var helpContent = this.$fs.readText(this.$staticConfig.helpTextPath).wait();
@@ -35,19 +42,10 @@ export class HelpCommand implements ICommand {
 			if (match) {
 				var helpText = match[1].trim();
 
-				var substitutionPoint: any;
-				while (substitutionPoint = helpText.match(this.$injector.dynamicCallRegex)) {
-					this.$logger.trace(substitutionPoint);
-					var data = this.$injector.dynamicCall(substitutionPoint[0]).wait();
-
-					var pointStart = substitutionPoint.index;
-					var pointEnd = pointStart + substitutionPoint[0].length;
-					helpText = helpText.substr(0, pointStart) + data + helpText.substr(pointEnd);
-				}
-
-				this.$logger.out(helpText);
+				var outputText = this.$microTemplateService.parseContent(helpText);
+				this.$logger.out(outputText);
 			} else {
-				this.$errors.fail({ formatStr: "Unknown help topic '%s'", suppressCommandHelp: true }, topic);
+				this.$errors.failWithoutHelp("Unknown help topic '%s'", topic);
 			}
 		}).future<void>()();
 	}
--- a/declarations.d.ts
+++ b/declarations.d.ts
@@ -228,3 +228,18 @@ interface ITypeScriptCompilationService {
 	initialize(typeScriptFiles: string[]): void;
 	compileAllFiles(): IFuture<void>;
 }
+
+interface IDynamicHelpService {
+	isProjectType(...args: string[]): IFuture<boolean>;
+	isPlatform(...args: string[]): boolean;
+	getLocalVariables(): IFuture<IDictionary<any>>;
+}
+
+interface IDynamicHelpProvider {
+	isProjectType(args: string[]): IFuture<boolean>;
+	getLocalVariables(): IFuture<IDictionary<any>>;
+}
+
+interface IMicroTemplateService {
+	parseContent(data: string): string;
+}
--- a/services/commands-service.ts
+++ b/services/commands-service.ts
@@ -11,6 +11,10 @@ class CommandArgumentsValidationHelper {
 }
 
 export class CommandsService implements ICommandsService {
+	private static HIERARCHICAL_COMMANDS_DELIMITER = "|";
+	private static HIERARCHICAL_COMMANDS_DEFAULT_COMMAND_DELIMITER = "|*";
+	private static HOOKS_COMMANDS_DELIMITER = "-";
+
 	constructor(private $errors: IErrors,
 		private $logger: ILogger,
 		private $injector: IInjector,
@@ -33,9 +37,10 @@ export class CommandsService implements ICommandsService {
 				}
 				if(command.enableHooks === undefined || command.enableHooks === true) {
 					// Handle correctly hierarchical commands
-					var childrenCommandsNames = _.map(this.$injector.getChildrenCommandsNames(commandName), command => { return command.replace("*", ""); });
-					if(_.contains(childrenCommandsNames, commandArguments[0])) {
-						commandName = util.format("%s-%s", commandName, commandArguments[0]);
+					var hierarchicalCommandName = this.$injector.buildHierarchicalCommand(commandName, commandArguments);
+					if(hierarchicalCommandName) {
+						commandName = helpers.stringReplaceAll(hierarchicalCommandName.commandName, CommandsService.HIERARCHICAL_COMMANDS_DEFAULT_COMMAND_DELIMITER, CommandsService.HOOKS_COMMANDS_DELIMITER);
+						commandName = helpers.stringReplaceAll(commandName, CommandsService.HIERARCHICAL_COMMANDS_DELIMITER, CommandsService.HOOKS_COMMANDS_DELIMITER); 
 					}
 
 					this.$hooksService.initialize(commandName);
@@ -77,7 +82,6 @@ export class CommandsService implements ICommandsService {
 		return (() => {
 			var command = this.$injector.resolveCommand(commandName);
 			var beautifiedName = helpers.stringReplaceAll(commandName, "|", " ");
-
 			if(command) {
 				// If command wants to handle canExecute logic on its own.
 				if(command.canExecute) {
--- a/services/dynamic-help-service.ts
+++ b/services/dynamic-help-service.ts
@@ -1,0 +1,28 @@
+///<reference path="../../.d.ts"/>
+"use strict";
+
+import os = require("os");
+
+export class DynamicHelpService implements IDynamicHelpService {
+	constructor(private $dynamicHelpProvider: IDynamicHelpProvider) { }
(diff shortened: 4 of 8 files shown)


commands/help.ts
```diff
@@ -13,7 +14,11 @@ export class HelpCommand implements ICommand {
 		private $staticConfig: Config.IStaticConfig) {}
 
 	public enableHooks = false;
-	public allowedParameters: ICommandParameter[] = [new commandParams.StringCommandParameter(this.$injector), new commandParams.StringCommandParameter(this.$injector)];
+	public canExecute(args: string[]): IFuture<boolean> {
+		return Future.fromResult(true);
+	}
+
+	public allowedParameters: ICommandParameter[] = [];
```
teobugslayer: Imho, we should change our code so that this line should not be needed. But if we agree on this, it is a separate task.

commands/help.ts
```diff
@@ -39,15 +45,31 @@ export class HelpCommand implements ICommand {
 				while (substitutionPoint = helpText.match(this.$injector.dynamicCallRegex)) {
 					this.$logger.trace(substitutionPoint);
 					var data = this.$injector.dynamicCall(substitutionPoint[0]).wait();
-
 					var pointStart = substitutionPoint.index;
-					var pointEnd = pointStart + substitutionPoint[0].length;
-					helpText = helpText.substr(0, pointStart) + data + helpText.substr(pointEnd);
+					if(typeof data === "boolean") {
```
teobugslayer: The code is correct, but see https://lodash.com/docs#isBoolean . I do not insist on changing it.

commands/help.ts
```diff
@@ -39,15 +45,31 @@ export class HelpCommand implements ICommand {
...
+						var endToken = "!" + substitutionPoint[0];
```
teobugslayer: Magic constant "!" - can we declare it descriptively?

services/commands-service.ts
```diff
@@ -33,9 +33,10 @@ export class CommandsService implements ICommandsService {
 				}
 				if(command.enableHooks === undefined || command.enableHooks === true) {
 					// Handle correctly hierarchical commands
-					var childrenCommandsNames = _.map(this.$injector.getChildrenCommandsNames(commandName), command => { return command.replace("*", ""); });
-					if(_.contains(childrenCommandsNames, commandArguments[0])) {
-						commandName = util.format("%s-%s", commandName, commandArguments[0]);
+					var hierarchicalCommandName = this.$injector.buildHierarchicalCommand(commandName, commandArguments);
+					if(hierarchicalCommandName) {
+						commandName = helpers.stringReplaceAll(hierarchicalCommandName.commandName, "|*", "-");
```
teobugslayer: Magic constants, consider describing them.

services/dynamic-help-service.ts
```diff
@@ -0,0 +1,26 @@
...
+"use strict";
+
+import os = require("os");
+
+export class DynamicHelpService implements IDynamicHelpService {
+	constructor(private $injector: IInjector) { }
+	public isProjectType(...args: string[]): IFuture<boolean> {
+		return ((): boolean => {
+			// resolve here as Project.IProject is not part of common-lib
+			var $project: Project.IProject = this.$injector.resolve("project");
```
teobugslayer: Does TNS have Project.IProject?

services/dynamic-help-service.ts
```diff
@@ -0,0 +1,26 @@
...
+			if($project && $project.getProjectDir().wait()) {
+				var framework = $project.projectData.Framework;
+				return _.any(args, arg => arg.toLowerCase() === framework.toLowerCase());
```
teobugslayer: consider moving framework.toLowerCase outside the loop.

services/dynamic-help-service.ts
```diff
@@ -0,0 +1,26 @@
...
+			}
+
+			return true;
+		}).future<boolean>()();
+	}
+
+	public isPlatform(...args: string[]): boolean {
+		var platform = os.platform();
+		return _.any(args, arg => arg.toLowerCase() === platform.toLowerCase());
```
teobugslayer: Consider moving platform.toLowerCase() outside the loop.

services/dynamic-help-service.ts
```diff
@@ -0,0 +1,26 @@
...
+			if($project && $project.getProjectDir().wait()) {
+				var framework = $project.projectData.Framework;
+				return _.any(args, arg => arg.toLowerCase() === framework.toLowerCase());
+			}
+
+			return true;
+		}).future<boolean>()();
+	}
+
+	public isPlatform(...args: string[]): boolean {
```
teobugslayer: Why this function is in DynamicHelpService ? It just works with os.platform.

yok.ts
```diff
@@ -284,13 +301,17 @@ export class Yok implements IInjector {
 	}
 
 	public get dynamicCallRegex(): RegExp {
-		return /#{([^.]+)\.([^}]+)}/;
+		return /#{([^.]+)\.([^}]+?)(\((.+)\))*}/;
```
teobugslayer: I would not mind if a comment roughly explained the regexp.

services/dynamic-help-service.ts
Fatme: Tns doesn't have `IProject` abstraction. Yep, we're right this is AB CLI specific code.

yok.ts
```diff
@@ -167,8 +167,29 @@ export class Yok implements IInjector {
...
+			subCommandName = subCommandName ? [subCommandName, arg].join("|") : arg;
+			remainingArguments = _.rest(remainingArguments);
+			if(_.any(subCommands,(sc) => sc === subCommandName)) {
+				foundSubCommand = true;
+				return false;
+			} else if(_.any(subCommands, sc => sc === "*" + subCommandName)) {
+				subCommandName = "*" + subCommandName;
+				foundSubCommand = true;
+				return false;
+			}
```
Fatme: Can we combine `if else if` into one statement -> `if(_.any(subCommands, sc => sc === subCommandName || sc === "*" + subCommandName)) { foundSubCommand = true; return false; })`?

yok.ts
```diff
@@ -167,8 +167,29 @@ export class Yok implements IInjector {
...
+		});
+
+		if(foundSubCommand) {
+			return { commandName: [parentCommandName, subCommandName].join("|"), remainingArguments: remainingArguments };
```
Fatme: We use this `[parentCommandName, subCommandName].join("|")` on many places so I suggest to extract it to separate function.

services/dynamic-help-service.ts
rosen-vladimirov: My idea is to use dynamicHelpService in tns as well, because its help also has some os specific text. So instead of having DynamicHelpService on two places, I've decided to place it here and note that one of the methods is not usable for tns. In fact if we change this method this way:
```Typescript
public isProjectType(...args: string[]): IFuture<boolean> {
  var $project = this.$injector.resolve("project");
  if($project && $project.getProjectDir && $project.getProjectDir().wait()) {
	var framework = $project.projectData.Framework.toLowerCase();
        return _.any(args, arg => arg.toLowerCase() === framework);
  }

   return true;
}
```
it could be called inside tns without throwing error. It will just return true.

yok.ts
rosen-vladimirov: I need to know that the matched command has * in the beginning and add it to subCommandName. That's why they are separated.

services/dynamic-help-service.ts
Fatme: My suggestion is to add `dynamicHelpProvider` as a private filed of `dynamicHelpService`.
```Typescript
public class DynamicHelpService {
     constructor(private $dynamicHelpProvider: IDynamicHelpProvider) { }
     
    public isProjectType(...args: string[]): IFuture<boolean> {
        return this.$dynamicHelpProvider.isProjectType(args);
    }
}
```
The implementation of `dynamicHelpProvider` in tns would be:
```Typescript
public class DynamicHelpProvider {
   public isProjectType(...args: string[]): IFuture<boolean> {
       return Future.fromResult(true);
   }
}
```
And the implementation in icenium-cli would be:

```Typescript
public class DynamicHelpProvider {
    constructor(private $project: Project.IProject) { }

   public isProjectType(...args: string[]): IFuture<boolean> {
         var framework = $project.projectData.Framework.toLowerCase();
         return _.any(args, arg => arg.toLowerCase() === framework);
   }
}
```

commands/help.ts
rosen-vladimirov: :smiley_cat:  I totally agree

teobugslayer:
:+1:

rosen-vladimirov:
ping @teobugslayer , @Fatme , @ikoevska  - check only the latest commit - "Add microtemplate service" - I've used template engine in it and the whole logic is changed.

services/micro-templating-service.ts
```diff
@@ -0,0 +1,49 @@
...
+			.replace(lineFeedRegEx, '\r')
+			.replace(newLineRegEx, '\n');
+	}
+
+	// Simple JavaScript Templating
+	// John Resig - http://ejohn.org/ - MIT Licensed
+	// Check http://ejohn.org/blog/javascript-micro-templating/ for more information
+	public parseInput(data: string): any {
+	// Figure out if we're getting a template, or if we need to
+	// load the template - and be sure to cache the result.
```
rosen-vladimirov: maybe this command should be removed as we do not need the caching

teobugslayer:
:+1:

services/micro-templating-service.ts
```diff
@@ -0,0 +1,38 @@
...
+		// Use ( ) in order to use $1 to get whole expression later
+		this.dynamicCallRegex = new RegExp(util.format("(%s)", this.$injector.dynamicCallRegex.source), "g");
+	}
+
+	private initializeLocalVariables(): IFuture<void> {
+		return (() => {
+			this.localVariables = {};
+			this.localVariables["isMobileWebsite"] = this.$dynamicHelpService.isProjectType("mobileWebsite").wait();
+			this.localVariables["isCordova"] = this.$dynamicHelpService.isProjectType("cordova").wait();
+			this.localVariables["isNativeScript"] = this.$dynamicHelpService.isProjectType("nativescript").wait();
```
Fatme: This is appbuilder specific code and should be in `icenium-cli`

declarations.d.ts
```diff
@@ -228,3 +228,16 @@ interface ITypeScriptCompilationService {
 	initialize(typeScriptFiles: string[]): void;
 	compileAllFiles(): IFuture<void>;
 }
+
+interface IDynamicHelpService {
+	isProjectType(...args: string[]): IFuture<boolean>;
```
Fatme: Shouldn't this be `args: string`?

services/micro-templating-service.ts
```diff
@@ -0,0 +1,38 @@
...
+
+import util = require("util");
+
+export class MicroTemplateService implements IMicroTemplateService{
+	private localVariables: IDictionary<any>;
+	private dynamicCallRegex: RegExp;
+
+	constructor(private $dynamicHelpService: IDynamicHelpService,
+		private $injector: IInjector) {
+		this.initializeLocalVariables().wait();
```
Fatme: Do you need to initialize local variables in constructor? Can we do it in `parseContent` function?

declarations.d.ts
rosen-vladimirov: As this method should be called from help.txt in the following way:
<% if (#{dynamicHelpService.isProjectType(cordova)}) { %> help for cordova <% } %>
or 
<% if (#{dynamicHelpService.isProjectType(cordova, nativescript)}) { %> help for cordova and nativescript <% } %>
the number of parameters is not defined. And in injector's dynamicCall we call
```
module[parsed[2]].apply(module, args);
```
where args are the passed arguments. In case I define isProjectType with (args:string[]) the first parameter(cordova in the example above) will be passed to args and rest of arguments will be omitted.

services/micro-templating-service.ts
rosen-vladimirov: We can

Fatme:
:+1:

PR #210, opened 2015-02-24
Remove tns related code from common

Fatme:
I was not able to build so I rebased on master.
Actually only the last commit contains the changes.


rosen-vladimirov:
we can consider this as new release branch with all changes from master, right ?

Fatme:
Yes, these are all changes from master.

PR #214, opened 2015-02-27
Use --frameworkPath option with debug command

Fatme:


Fatme:
I am going to reopen against release

PR #215, opened 2015-02-27
Change $logger.error color to red

Mitko-Kerezov:
Note: This shade of red is a lot different than the shade of red displayed by $errors.fail
Tested in cmd and bash under windows


rosen-vladimirov:
:+1:

PR #216, opened 2015-02-27
Use --frameworkPath option with debug command

Fatme:


mobile/ios/ios-device.ts
```diff
@@ -486,8 +488,19 @@ export class IOSDeviceDebugging {
...
+			if(this.$fs.getFsStats(options.frameworkPath).wait().isFile()) {
+				this.$errors.fail("frameworkPath option must be path to directory which contains tns-ios framework");
+			}
+
+			tnsIosPackage = path.resolve(options.frameworkPath);
+		} else {
+			tnsIosPackage = this.$npm.install("tns-ios").wait();
+		}
+
+		var safariPath = path.join(tnsIosPackage, "WebInspectorUI/Chrome/Main.html");
```
rosen-vladimirov: Chrome? Shouldn't we use Safari?

rosen-vladimirov:
:+1:  from me, but as it is for nativescript release, I would like to ping @ErjanGavalji, @PanayotCankov  and @dtopuzov

PanayotCankov:
The "IOSDeviceDebugging" class from the common module belongs to the nativescript-cli. It will be moved soon. Most of the "options" in "ios-device" are used in that class and will be eliminated once it is moved. We need the "--frameworkPath" so that we can automate debugging tests.

Fatme:
This PR is merged in master

PR #219, opened 2015-03-09
Add method to check if analytics tracking is enabled

rosen-vladimirov:
Add isEnabled method which returns true only when analytics tracking is enabled. We need the value in order to pass it to simulator in appbuilder-cli.

http://teampulse.telerik.com/view#item/286925


teobugslayer:
:+1:

PR #220, opened 2015-03-09
Fix unzipping on node 0.12

teobugslayer:
Under node 0.12 we generate invalid command line on Windows.

Needed to fix http://teampulse.telerik.com/view#item/285421


rosen-vladimirov:
:+1:

PR #223, opened 2015-03-10
Fix hierarchical commands

rosen-vladimirov:
Fix code which should execute default hierarchical command which accepts arguments.
Split device commands in order to use different context for each of them,

### DIFF
--- a/commands/device/list-devices.ts
+++ b/commands/device/list-devices.ts
@@ -4,6 +4,7 @@
 import util = require("util");
 import options = require("./../../options");
 import commandParams = require("../../command-params");
+import MobileHelper = require("../../mobile/mobile-helper");
 
 export class ListDevicesCommand implements ICommand {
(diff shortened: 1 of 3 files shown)


commands/device/list-devices.ts
```diff
@@ -40,4 +41,34 @@ export class ListDevicesCommand implements ICommand {
...
+}
+$injector.registerCommand("device|android", ListAndroidDevicesCommand);
+
+class ListiOSDevicesCommand implements ICommand {
+	constructor(private $injector: IInjector) { }
+
+	allowedParameters: ICommandParameter[] = [];
+
+	public execute(args: string[]): IFuture<void> {
+		return (() => {
```
teobugslayer: We may remove the return ().future()() wrapping

teobugslayer:
:+1:

PR #226, opened 2015-03-12
Replace  appbuilder strings with CLIENT_NAME

rosen-vladimirov:
Common lib is used by several CLIs - appbuilder-cli, nativescript-cli, etc. 

- There are several places where appbuilder is used, while it should be staticConfig.CLIENT_NAME. Fix all of these.
- Replace MobileHelper's exported functions in a new class - MobileHelper. This way we can resolve it from injector and it can easily use staticConfig's CLIENT_NAME. 
- Remove DevicePlatforms enum and add DevicePlatformsConstants class which can be used instead of the enum. 
- Add SYS_REQUIREMENTS_LINK to StaticConfig in order to have different links for each CLI. 
- Add getClientName and getPrivacyPolicyLink methods to IAnalyticsSettingsService.

http://teampulse.telerik.com/view#item/287728

### DIFF
--- a/definitions/mobile.d.ts
+++ b/definitions/mobile.d.ts
@@ -228,4 +228,27 @@ declare module Mobile {
 		stdoutFilePath?: string;
 		appId?: string;
 	}
+
+	interface IPlatformsCapabilities {
+		getPlatformNames(): string[];
+		getAllCapabilities(): IDictionary<Mobile.IPlatformCapabilities>;
+	}
+
+	interface IMobileHelper {
+		platformNames: string[];
+		isAndroidPlatform(platform: string): boolean;
+		isiOSPlatform(platform: string): boolean;
+		isWP8Platform(platform: string): boolean;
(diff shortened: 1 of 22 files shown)


definitions/mobile.d.ts
```diff
@@ -228,4 +228,29 @@ declare module Mobile {
...
+	}
+
+	interface IMobileHelper {
+		platformNames: string[];
+		isAndroidPlatform(platform: string): boolean;
+		isiOSPlatform(platform: string): boolean;
+		isWP8Platform(platform: string): boolean;
+		normalizePlatformName(platform: string): string;
+		isPlatformSupported(platform: string): boolean;
+		generateWP8GUID(): string;
```
Fatme: This function seems related only to appbuilder

rosen-vladimirov: :crying_cat_face: I've missed it

teobugslayer:
:+1:

PR #227, opened 2015-03-12
Fix validation of device platform names

rosen-vladimirov:
In case the CLI doesn't support a mobile platform, for example WP8, but it is valid mobile platform according to normalizePlatformName(WP8 for example), we should break the execution.


Fatme:
:+1:

PR #229, opened 2015-03-13
Enable proxy using for http requests

rosen-vladimirov:
Rename the following config options:
 - FIDDLER_HOSTNAME to PROXY_HOSTNAME
 - PROXY_TO_FIDDLER to USE_PROXY

Add PROXY_PORT option to config with default value 8888. Add ConfigBase class which should be used as a base for CLI specific configs.

Required for https://github.com/NativeScript/nativescript-cli/issues/297 and https://github.com/NativeScript/nativescript-cli/issues/302

### DIFF
--- a/config-base.ts
+++ b/config-base.ts
@@ -1,0 +1,18 @@
+///<reference path="../.d.ts"/>
+"use strict";
+
+import path = require("path");
+import util = require("util");
+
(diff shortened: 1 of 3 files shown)


teobugslayer:
:+1:

PR #230, opened 2015-03-13
Remove duplicate -- options

rosen-vladimirov:
Yargs adds -- options to yargs.argv in two ways (--profile-dir is added as profile-dir and profileDir). In order to workaround this behavior, we had to add  all dashed options in both ways to our known options. Fix our validation logic, so we'll have to add only profile-dir.
Remove check for client name when validating -- options. This way options will be validated no matter if CLIENT_NAME( tns) or CLIENT_NAME_ALIAS (nativescript) is used on command line.

### DIFF
--- a/errors.ts
+++ b/errors.ts
@@ -167,24 +167,37 @@ export class Errors implements IErrors {
 		return this.executeAction(action);
 	}
 
-	public validateYargsArguments(parsed: any, knownOpts: any, shorthands: any, clientName?: string): void {
-		if(path.basename(process.argv[1]).indexOf(clientName) !== -1) {
-			_.each(_.keys(parsed), (opt) => {
-				var option = shorthands[opt] ? shorthands[opt] : opt;
-
-				if (option !== "_" && option !== "$0" && !knownOpts[option]) {
-					this.failWithoutHelp("The option '%s' is not supported. To see command's options, use '$ %s help %s'. To see all commands use '$ %s help'.", opt, clientName, process.argv[2], clientName);
-				} else if (knownOpts[option] !== Boolean && typeof (parsed[opt]) === 'boolean') {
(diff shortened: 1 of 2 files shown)


errors.ts
```diff
@@ -167,24 +167,36 @@ export class Errors implements IErrors {
...
+			return this.getYargsOriginalOption(secondaryPresentation);
 		}
+
+		return option;
+	}
+
+	public validateYargsArguments(parsed: any, knownOpts: any, shorthands: any, clientName?: string): void {
+		var knownOptionsKeys = _.keys(knownOpts);
+		_.each(_.keys(parsed), (opt) => {
+			var option: string = shorthands[opt] ? shorthands[opt] : opt;
```
teobugslayer: ```shorthands[opt] || opt```

errors.ts
```diff
@@ -167,24 +167,36 @@ export class Errors implements IErrors {
...
-				} else if (knownOpts[option] === String && helpers.isNullOrWhitespace(parsed[opt])) {
-					this.failWithoutHelp("The option '%s' requires non-empty value.", opt);
-				} else if (knownOpts[option] === Boolean && typeof (parsed[opt]) !== 'boolean') {
-					this.failWithoutHelp("The option '%s' does not accept values.", opt);
-				}
-			});
+	public getYargsOriginalOption(option: string): string {
+		var matchUpperCaseLetters = option.match(/(.+?)([A-Z])(.*)/);
+		if(matchUpperCaseLetters) {
+			// get here if profileDir option is specified
```
teobugslayer: we probably can smuggle some _.kebabCase in here :)

rosen-vladimirov: Unfortunately not. I've tried using kebabCase, but it cannot convert a-b-c-d-e-f-g in the same manner as yargs (check my last unit test).

teobugslayer:
:+1:

PR #232, opened 2015-03-13
Fix analytics OS tracking

rosen-vladimirov:
Set userAgentString to (osType). For example on Linux the agent will be (Linux).


dtopuzov:
:+1:

Fatme:
:+1:

PR #233, opened 2015-03-13
Fix ConfigBase parent

rosen-vladimirov:
ConfigBase must implement Config.IConfig, not IConfiguration. IConfiguration is specific for each CLI.


Fatme:
:+1:

PR #236, opened 2015-03-17
Move NativeScript debugging for iOS in the nativescript-cli, add emulator notifications to support debugging

PanayotCankov:


mobile/android/android-device.ts
```diff
@@ -308,7 +308,7 @@ export class AndroidDevice implements Mobile.IDevice {
             this.$childProcess.exec(clearDebugEnvironmentCommand).wait();
 
             var isRunning = false;
-            for (var i = 0; i < 30; i++) {
+            for (var i = 0; i < 60; i++) {
```
rosen-vladimirov: not for this PR, but we definitely have to use --timeout option here.

rosen-vladimirov:
Awesome work, @PanayotCankov , 
:+1:

mobile/android/android-device.ts
PanayotCankov: Log it. It fails on slow emulators.

PR #237, opened 2015-03-18
At post-install, ask the user to turn on command line autocompletion

teobugslayer:
Implements https://github.com/NativeScript/nativescript-cli/issues/350

### DIFF
--- a/commands/completion-prompt.ts
+++ b/commands/completion-prompt.ts
@@ -1,0 +1,30 @@
+///<reference path="../../.d.ts"/>
+"use strict";
+import helpers = require("../helpers");
+
+export class CompletionPromptCommand implements ICommand {
+
+	constructor(private $autoCompletionService: IAutoCompletionService,
+				private $logger: ILogger,
+				private $prompter: IPrompter) {
+	}
+
+	public disableAnalytics = true;
+	public allowedParameters: ICommandParameter[] = [];
+
+	public execute(args: string[]): IFuture<void> {
+		return (() => {
+			if (helpers.isInteractive()) {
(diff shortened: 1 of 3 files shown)


commands/completion-prompt.ts
```diff
@@ -0,0 +1,28 @@
...
+				private $logger: ILogger,
+				private $prompter: IPrompter) {
+	}
+
+	public disableAnalytics = true;
+	public allowedParameters: ICommandParameter[] = [];
+
+	public execute(args: string[]): IFuture<void> {
+		return (() => {
+			this.$logger.out("If you are using bash or zsh, you can enable command line auto completion.");
```
rosen-vladimirov: Maybe we should notify the user that we will write some data in bash_profile, zsh and bashrc

rosen-vladimirov:
:+1:  after considering my suggestion

commands/completion-prompt.ts
```diff
@@ -0,0 +1,28 @@
...
+			var message = "Enable it now?";
+
+			var autoCompetionStatus = this.$prompter.confirm(message, () => "y").wait();
```
Fatme: Shouldn't we prompt this message only when we have iteractive console?

commands/completion-prompt.ts
teobugslayer: Done - in the help file

commands/completion-prompt.ts
teobugslayer: Will restore the code (it used to be there)

Fatme:
:+1:

PR #239, opened 2015-03-23
Polished binary property list parsing

PanayotCankov:


mobile/ios/ios-core.ts
```diff
@@ -797,51 +809,74 @@ class PosixSocket implements Mobile.IiOSDeviceSocket {
...
+									this.state = ReadState.Plist;
+									break;
+								case ReadState.Plist:
+									try {
+										var plistBuffer = this.buffer.slice(0, this.length);
+										var message = bplistParser.parseBuffer(plistBuffer);
+										this.$logger.trace("MESSAGE RECEIVING");
+										this.$logger.trace(message);
+										try {
+											if (message && typeof (message) === "object" && message[0]) {
```
rosen-vladimirov: message is object and you are using message[0]. May I think message is array? If yes, instead of checking for object, you may use lodash's isArray

rosen-vladimirov:
May I ask you for short description of the problem you've found and the way you are fixing it. 
From the code I assume the problem is when we receive consecutive (or simultaneous) data events and our current code does not process them correctly as we do not expect to have more than one message in the data. Am I right?

PanayotCankov:
@rosen-vladimirov Yep. The code speaks for itself better than explanation :)

The "deploy" command freezes sometimes during the deploy step.

I need the "deploy" up and stable running for the "debug" tests, as the debug depends on successful deploy.

During investigation I've found out the binary plist sometimes get concatenated, and the "Complete" message gets in the "data" event along with the last progress report message. The deploy was successful, but the promise was never resolved, since only the progress message part of the data was parsed.

dtopuzov:
:+1:

mobile/ios/ios-core.ts
PanayotCankov: I've pretty much preserved this code. I do not want to know what the message is.

rosen-vladimirov:
@PanayotCankov   Thank you for the explanation. The change seems logic based on it, so I give you my :+1: 
Ping @Fatme to check the change as I believe she has written the old code (or at least understands it better than me :smile_cat: )

PR #240, opened 2015-03-23
Emit system information on every run of the CLI

teobugslayer:
Gather system information about the runtime environment.

### DIFF
--- a/commands/post-install.ts
+++ b/commands/post-install.ts
@@ -9,13 +9,13 @@ export class PostInstallCommand implements ICommand {
 	private static SEVEN_ZIP_ERROR_MESSAGE = "It looks like there's a problem with your system configuration. " +
 		"You can find all system requirements on %s"; 
 
-	constructor(private $autoCompletionService: IAutoCompletionService,
-		private $fs: IFileSystem,
+	constructor(private $fs: IFileSystem,
 		private $staticConfig: Config.IStaticConfig,
 		private $childProcess: IChildProcess,
-		private $errors: IErrors,
 		private $commandsService: ICommandsService,
-		private $htmlHelpService: IHtmlHelpService) {
+		private $htmlHelpService: IHtmlHelpService,
+		private $sysInfo: ISysInfo,
+		private $logger: ILogger) {
 	}
 
 	public disableAnalytics = true;
@@ -31,10 +31,50 @@ export class PostInstallCommand implements ICommand {
 				}
 			}
 
-			this.checkSevenZip().wait();
-			this.$commandsService.tryExecuteCommand("autocomplete", []).wait();
 			this.$htmlHelpService.generateHtmlPages().wait();
+
+			var sysInfo = this.$sysInfo.getSysInfo();
+			var isNativeScript = this.$staticConfig.CLIENT_NAME === "tns";
+			if (isNativeScript) {
+				this.printNSWarnings(sysInfo);
+			} else {
+				this.printAppBuilderWarnings(sysInfo);
+			}
+
+			this.checkSevenZip().wait();
+
+			this.$commandsService.tryExecuteCommand("autocomplete", []).wait();
 		}).future<void>()();
+	}
+
+	private printNSWarnings(sysInfo: ISysInfoData) {
+		if (!sysInfo.adbVer) {
+			this.$logger.warn("Cannot find adb in the path. The built-in one will be used, which may be incompatible with the latest Android SDK or Genymotion. Adjust system path to include adb from the latest Android SDK.");
+		}
+		if (!sysInfo.antVer) {
+			this.$logger.warn("Cannot find ant. You cannot build Android projects. Download and install ant as described in the Android SDK documentation.");
+		}
+		if (!sysInfo.javaVer) {
+			this.$logger.warn("Cannot find java. You cannot build Android projects and use Android Emulator. Install java as described in the Android SDK documentation.");
+		}
+		if (hostInfo.isDarwin() && !sysInfo.xcodeVer) {
+			this.$logger.warn("Cannot find Xcode. You cannot build iOS projects and use the iOS Simulator. Install Xcode from the App Store.");
+		}
+		if (!sysInfo.itunesInstalled) {
+			this.$logger.warn("iTunes is not installed. Commands for working with iOS devices will not work. Download and install iTunes from http://www.apple.com");
+		}
+	}
+
+	private printAppBuilderWarnings(sysInfo: ISysInfoData) {
+		if (!sysInfo.adbVer) {
+			this.$logger.warn("Cannot find adb in the path. The built-in one will be used, which may be incompatible with the latest Android SDK or Genymotion. Adjust system path to include adb from the latest Android SDK.");
+		}
+		if (!sysInfo.javaVer) {
+			this.$logger.warn("Cannot find java. You cannot use Android Emulator. Install java as described in the Android SDK documentation.");
+		}
+		if (!sysInfo.itunesInstalled) {
+			this.$logger.warn("iTunes is not installed. Commands for working with iOS devices will not work. Download and install iTunes from http://www.apple.com");
+		}
 	}
 
 	private checkSevenZip(): IFuture<void> {
--- a/declarations.d.ts
+++ b/declarations.d.ts
@@ -255,3 +255,44 @@ interface IHtmlHelpService {
 	getCommandLineHelpForCommand(commandName: string): IFuture<string>;
 	openHelpForCommandInBrowser(commandName: string): IFuture<void>;
 }
+
+interface ISysInfoData {
+	/** name and version of the CLI app itself */
+	procInfo: string;
+
+	// os stuff
+	/** os platform flavour, reported by os.platform */
+	platform: string;
+	/** Full os name, like `uname -a` on unix, registry query on win */
+	os: string;
+	/** .net version, applicable to windows only */
+	dotNetVer: string;
+	/** The command shell in use, usually bash or cmd */
+	shell: string;
+
+	// node stuff
+	/** node.js version, returned by `process.version` */
+	nodeVer: string;
+	/** npm version, returned by `npm -v` */
+	npmVer: string;
+	/** Process architecture, returned by `process.arch` */
+	procArch: string;
+	/** node-gyp version as returned by `node-gyp -v`*/
+	nodeGypVer: string;
+
+	// dependencies
+	/** Version string of java as returned by `java -version` */
+	javaVer: string;
+	/** version string of ant, as returned by `ant -version` */
+	antVer: string;
+	/** Xcode version string as returned by `xcodebuild -version`. Valid only on Mac */
+	xcodeVer: string;
+	/** Version string of adb, as returned by `adb version` */
+	adbVer: string;
+	/** Whether iTunes is installed on the machine */
+	itunesInstalled: boolean;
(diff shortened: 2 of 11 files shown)


host-info.ts
```diff
@@ -24,3 +26,43 @@ export function isLinux() {
...
+		try {
+			dotNetVersion(message || "An error occurred while reading the registry.").wait();
+			return true;
+		} catch (e) {
+			return false;
+		}
+	}).future<boolean>()();
+}
+
+export var hostCapabilities: { [key: string]: IHostCapabilities } = {
```
rosen-vladimirov: this var is already part of appbuilder-cli and it is not used by nativescript-cli. Is there a reason to add it here? I couldn't find any place where you've used it in nativescript-cli

sysinfo.ts
```diff
@@ -0,0 +1,99 @@
...
+			res.dotNetVer = hostInfo.isWindows() ? hostInfo.dotNetVersion(".Net is not installed").wait() : undefined;
+
+			// node stuff
+			res.nodeVer = process.version;
+			res.npmVer = this.$childProcess.exec("npm -v").wait().split("\n")[0];
+			res.procArch = process.arch;
+
+			// dependencies
+			try {
+				res.javaVer = this.$childProcess.spawnFromEvent("java", ["-version"], "exit").wait().stderr.split(os.EOL)[2];
```
rosen-vladimirov: have you tested this on linux and mac? Are all JAVA version writing to stderr instead of stdout? 
If JAVA is not configured properly, but it's part of the PATH, spliting by newLine the output of stderr will may show some data - do we consider this as working JAVA or not?

dispatchers.ts
```diff
@@ -21,6 +22,9 @@ export class CommandDispatcher implements ICommandDispatcher {
 				return;
 			}
 
+			this.$logger.trace("System information:");
+			this.$logger.trace(this.$sysInfo.getSysInfo());
```
rosen-vladimirov: If the log level is not trace, will the getSysInfo code be executed and just hidden from the output or the method will not be called at all?

commands/post-install.ts
```diff
@@ -32,6 +33,12 @@ export class PostInstallCommand implements ICommand {
 
 			this.checkSevenZip().wait();
 			this.$commandsService.tryExecuteCommand("autocomplete", []).wait();
+
+			var sysInfo = this.$sysInfo.getSysInfo();
+			if (!sysInfo.adbVer) { console.warn("Cannot find adb in the path. The built-in one will be used, which may be incompatible with the latest Android SDK or Genymotion"); }
+			if (!sysInfo.antVer) { console.warn("Cannot find ant. You cannot build Android projects."); }
```
rosen-vladimirov: appbuilder is also calling PostInstallCommand and these NativeScript specific messages will scare our users. They should be moved to nativescript-cli only. 
(or I'm totally missing something here)

declarations.d.ts
```diff
@@ -246,3 +246,30 @@ interface IDynamicHelpProvider {
...
 	parseContent(data: string): string;
 }
+
+interface ISysInfoData {
+
+	procInfo: string; // name and version of the CLI
+
+	// os stuff
+	platform: string; // os.platform
+	os: string; // full os name, like `uname -a` on unix, registry query on win
```
rosen-vladimirov: Just a suggestion, not a show stopper:
Use javascript api documentation comments
```JavaScript
/** Full os name, like `uname -a` on unix, registry query on win. */
os:string;
```
This way VisualStudio, WebStorm, etc. will show the message when you try using this property.

sysinfo.ts
```diff
@@ -0,0 +1,99 @@
...
+
+	private winVer(): string {
+		var future = new Future();
+		var Winreg = require("winreg");
+		var winver: string = "";
+		var regKey = new Winreg({
+			hive: Winreg.HKLM,
+			key:  '\\Software\\Microsoft\\Windows NT\\CurrentVersion'
+		});
+		regKey.get("ProductName", (err: Error, value: any) => {
```
rosen-vladimirov: I know we've discussed this, but I still prefer Future approach here:
```TypeScript
var winVer = util.format("%s %s.%s", this.getRegKey("ProductName").wait() , this.getRegKey("CurrentVersion").wait(), this.getRegKey("CurrentBuild").wait());

private getRegKey(key:string): IFuture<string> {
	var regKeyFuture = new Future<string>();
	regKey.get("ProductName", (err: Error, value: any) => {
		if (err) {
				future.throw(new Error());
		} else {
			future.return(value.value);
		}
	});
	
	return regKeyFuture;
}
```

sysinfo.ts
```diff
@@ -0,0 +1,99 @@
+///<reference path="../.d.ts"/>
+"use strict";
+
+import os = require("os");
+import osenv = require("osenv");
```
rosen-vladimirov: osenv does not have definition in common - we should move the definition file here.

commands/post-install.ts
```diff
@@ -32,6 +33,12 @@ export class PostInstallCommand implements ICommand {
...
+			if (!sysInfo.javaVer) { console.warn("Cannot find java. You cannot build Android projects."); }
+			if (hostInfo.isDarwin() && !sysInfo.xcodeVer) { console.warn("Cannot find Xcode. You cannot build iOS projects and use the iOS Simulator."); }
```
ligaz: Can we tweak the messages a little bit to provide next steps for action? For example: `You can download the latest version of Xcode from the App Store`.

dispatchers.ts
ligaz: I guess gathering system information could take time which will slow down each command execution. Can this be `if`ed so that we `getSysInfo` only when tracing.

sysinfo.ts
```diff
@@ -0,0 +1,99 @@
...
+			var res: ISysInfoData = Object.create(null);
+
+			var packageJson = require("../../package.json");
+			res.procInfo = packageJson.name + " / " + packageJson.version;
+
+			// os stuff
+			res.platform = os.platform();
+			res.os = hostInfo.isWindows() ? this.winVer() : this.unixVer();
+			res.shell = osenv.shell();
+			res.dotNetVer = hostInfo.isWindows() ? hostInfo.dotNetVersion(".Net is not installed").wait() : undefined;
```
ligaz: Is it better to move the `isWindows()` in the hostInfo implementation? In this case if we need this information on another place we will not have to do the ternary again.

dispatchers.ts
rosen-vladimirov: I agree with @ligaz , I've tried the code and getSysInfo is called no matter of the logging level. However if we implement our idea to have log level trace and output it to a file, no matter of the console logging level, such check will not be needed as we'll always need this information.

sysinfo.ts
```diff
@@ -0,0 +1,99 @@
...
+			res.procArch = process.arch;
+
+			// dependencies
+			try {
+				res.javaVer = this.$childProcess.spawnFromEvent("java", ["-version"], "exit").wait().stderr.split(os.EOL)[2];
+			} catch(e) {
+				res.javaVer = null;
+			} // if we got an error, assume not working
+
+			res.antVer = this.exec("ant -version").split(os.EOL)[0];
```
Fatme: Why do we need to split this? Can we add all output from `ant -version`?

commands/post-install.ts
teobugslayer: Yes, this is the intent. I'll add the missing information.

host-info.ts
teobugslayer: A copy-paste mistake, will return it to its rightful place.

sysinfo.ts
teobugslayer: Of course it writes on stderr - at least it **consistently** breaks the rules.

sysinfo.ts
teobugslayer: Good catch! Will move it

commands/post-install.ts
teobugslayer: A good catch. My intent was to kick-start the discussion about the best way to use this system. 
I will augment the PR to issue only relevant warnings.

sysinfo.ts
teobugslayer: Because the output of ant -version is actually 2 lines, and the second one is empty.

dispatchers.ts
teobugslayer: Yep, I will do this.

sysinfo.ts
teobugslayer: Yep, good catch. will do.

sysinfo.ts
teobugslayer: Sure, though I like the pyramid o' doom ;) (OK, I lied about this)

commands/post-install.ts
```diff
@@ -32,6 +33,28 @@ export class PostInstallCommand implements ICommand {
 
 			this.checkSevenZip().wait();
 			this.$commandsService.tryExecuteCommand("autocomplete", []).wait();
+
+			var isNS = require("../../package.json").name === "nativescript";
```
rosen-vladimirov: why don't you just check 
```TypeScript
this.$staticConfig.CLIENT_NAME === "tns"
```

dispatchers.ts
```diff
@@ -21,6 +22,12 @@ export class CommandDispatcher implements ICommandDispatcher {
 				return;
 			}
 
+			if (true) {
```
rosen-vladimirov: :fearful:

rosen-vladimirov:
I think it will be nice to have a new command that the users can call to check their system configuration, for exampe ```$ appbuilder check-config``` This will output the same as post-install does with the current implementation. If you do not have such option, you can install CLI, you'll see some warnings, for example that your adb is not configured correctly, and of course you decide to fix them. After you've fixed them, you'll probably want to check if you've configured them correctly based on CLI's requirements, but the only way to do it is to install it again (or try --log trace in case you know that it will print this info).
Ping @ligaz , @Fatme , what do you think about this?

commands/post-install.ts
```diff
@@ -31,12 +33,53 @@ export class PostInstallCommand implements ICommand {
 				}
 			}
 
+			this.$htmlHelpService.generateHtmlPages().wait();
+
+			var isNS = require("../../package.json").name === "nativescript";
```
rosen-vladimirov: this variable is not used anymore

commands/post-install.ts
```diff
@@ -31,12 +33,53 @@ export class PostInstallCommand implements ICommand {
...
+			var sysInfo = this.$sysInfo.getSysInfo();
+			var isNativeScript = require("../../../package.json").name === "nativescript";
```
rosen-vladimirov:
```TypeScript
this.$staticConfig.CLIENT_NAME === "tns"
```
but not a merge stopper

rosen-vladimirov:
:+1:  after removing the unused variable

ErjanGavalji:
:+1:

commands/post-install.ts
teobugslayer: Ah, an artifact from rebasing. Thanks for spotting.

teobugslayer:
@rosen-vladimirov , excellent idea, but perhaps for a second PR? This one is large enough as is.

rosen-vladimirov:
:+1:  for another PR

teobugslayer:
Created an issue: https://github.com/NativeScript/nativescript-cli/issues/382

PR #241, opened 2015-03-23
Show HTML Help

rosen-vladimirov:
## Use case:
CLIs should be able to show HTML help when using ```<cli name> help <command name>```
When using ```<cli name> <command name> --help``` console help should be shown.
When an error is raised in command, only console help is shown.

## Implementation
* ```HelpCommand``` is modified to show console help when ```options.help``` is specified and html help when it is not.
* Add ```HtmlHelpService``` which handles both console and html help by parsing markdown help files.
* Each commands help is placed in a separate file. The files are markdown and they are used for both html and console help.
* Html pages are generated from markdown on post-install. As post-install command could fail before generating html pages, when html help should be shown, we check if command name is valid and if the command exists, but its html help is missing, we generate all html pages.
* Modify ```CommandsService``` to show console help by specifying options.help before calling ```HelpCommand```
* Modify ```DynamicHelpService``` and ```DynamicHelpProvider``` to accept isHtml option. When ```isHtml``` is specified, all localVariables will be set to true. This way html help will show full help, no matter of the conditions.
* Add ```basic_page.html``` page and used three placeholders in it. The page is used to generate all html files, by replacing the placeholders with commands name, help content and path to css file.
* Add ```styles.css``` file which contains styles used in html pages.
* Add definitions for marked module.
* Add ```appendFile``` method to ```FileSystem```
* Fix issue in yok, which was not requiring the first hierarchical command if it was not default one (if the hierarchical command doesn't have default command to execute) - this was leading to incorrect behavior when checking if command is valid.

http://teampulse.telerik.com/view#item/259417

### DIFF
--- a/services/html-help-service.ts
+++ b/services/html-help-service.ts
@@ -1,0 +1,158 @@
+///<reference path="../../.d.ts"/>
+"use strict";
+
+import util = require("util");
+import Future = require("fibers/future");
+import path = require("path");
+import options = require("options");
+import marked = require("marked");
+var TerminalRenderer = require('marked-terminal');
+var chalk = require("chalk");
+
+export class HtmlHelpService implements IHtmlHelpService {
+	private static MARKDOWN_FILE_EXTENSION = ".md";
+	private static HTML_FILE_EXTENSION = ".html";
+	private static MAN_PAGE_NAME_PLACEHOLDER = "@MAN_PAGE_NAME@";
+	private static HTML_COMMAND_HELP_PLACEHOLDER = "@HTML_COMMAND_HELP@";
+	private static RELATIVE_PATH_TO_STYLES_CSS_PLACEHOLDER = "@RELATIVE_PATH_TO_STYLES_CSS@";
+
+	private pathToManPages: string;
+	private pathToHtmlPages: string;
+	private get pathToStylesCss(): string {
+		return path.join(this.$staticConfig.HTML_HELPERS_DIR, "styles.css");
+	}
+
+	private get pathToBasicPage(): string {
+		return path.join(this.$staticConfig.HTML_HELPERS_DIR, "basic-page.html");
+	}
+
+	constructor(private $logger: ILogger,
+		private $injector: IInjector,
+		private $errors: IErrors,
+		private $fs: IFileSystem,
+		private $staticConfig: Config.IStaticConfig,
+		private $microTemplateService: IMicroTemplateService,
+		private $opener: IOpener) {
+		this.pathToHtmlPages = this.$staticConfig.HTML_PAGES_DIR;
+		this.pathToManPages = this.$staticConfig.MAN_PAGES_DIR;
+	}
+
+	public generateHtmlPages(): IFuture<void> {
+		return (() => {
+			var mdFiles = this.$fs.enumerateFilesInDirectorySync(this.pathToManPages);
+			var basicHtmlPage = this.$fs.readFile(this.pathToBasicPage).wait().toString();
+			var futures = _.map(mdFiles, markdownFile => this.createHtmlPage(basicHtmlPage, markdownFile));
+			Future.wait(futures);
+			this.$logger.trace("Finished generating HTML files.");
+		}).future<void>()();
+	}
+
+	private createHtmlPage(basicHtmlPage: string, pathToMdFile: string): IFuture<void> {
+		return (() => {
+			var mdFileName = path.basename(pathToMdFile);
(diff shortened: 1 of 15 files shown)


services/html-help-service.ts
```diff
@@ -0,0 +1,161 @@
...
+var chalk = require("chalk");
+
+export class HtmlHelpService implements IHtmlHelpService {
+	private static MARKDOWN_FILE_EXTENSION = ".md";
+	private static HTML_FILE_EXTENSION = ".html";
+	private static MAN_PAGE_NAME_PLACEHOLDER = "@MAN_PAGE_NAME@";
+	private static HTML_COMMAND_HELP_PLACEHOLDER = "@HTML_COMMAND_HELP@";
+	private static RELATIVE_PATH_TO_STYLES_CSS_PLACEHOLDER = "@RELATIVE_PATH_TO_STYLES_CSS@";
+
+	private pathToManPages: string;
```
Fatme: Why do you need this local variable? You can use directly `staticConfig.HTML_PAGES_DIR`

rosen-vladimirov: The reasons is caching of the result.
 staticConfig is executing ```path.join``` when getting HTML_PAGES_DIR and MAN_PAGES_DIR, I do not want to execute it every time.

services/html-help-service.ts
```diff
@@ -0,0 +1,161 @@
...
+		return (() => {
+			var mdFileName = path.basename(pathToMdFile);
+			var htmlFileName = mdFileName.replace(HtmlHelpService.MARKDOWN_FILE_EXTENSION, HtmlHelpService.HTML_FILE_EXTENSION);
+			this.$logger.trace("Generating '%s' help topic.", htmlFileName);
+
+			var helpText = this.$fs.readText(pathToMdFile).wait();
+			var outputText = this.$microTemplateService.parseContent(helpText, { isHtml: true });
+			var htmlText = marked(outputText);
+
+			var filePath = pathToMdFile
```
Fatme: `var filePath = path.join(this.pathToHtmlPages, htmlFileName` is shorter.

rosen-vladimirov: yep, I'm not sure what I've been thinking here :smile_cat:

services/html-help-service.ts
```diff
@@ -0,0 +1,161 @@
...
+				.replace(path.basename(this.pathToManPages), path.basename(this.pathToHtmlPages))
+				.replace(mdFileName, htmlFileName);
+			this.$logger.trace("HTML file path for '%s' man page is: '%s'.", mdFileName, filePath);
+
+			var outputHtml = basicHtmlPage
+				.replace(HtmlHelpService.MAN_PAGE_NAME_PLACEHOLDER, mdFileName.replace(".md", ""))
```
Fatme: You can use `MARKDOWN_FILE_EXTENSION` constant here.

services/html-help-service.ts
```diff
@@ -0,0 +1,161 @@
...
+	private readManFileForCommand(commandName: string): IFuture<string> {
+		return ((): string => {
+			var mdFileName = this.convertCommandNameToFileName(commandName) + HtmlHelpService.MARKDOWN_FILE_EXTENSION;
+			this.$logger.trace("Reading help for command '%s'. FileName is '%s'.", commandName, mdFileName);
+			var helpText: string;
+
+			var markdownFile = _.find(this.$fs.enumerateFilesInDirectorySync(this.pathToManPages), file => path.basename(file) === mdFileName);
+			if(markdownFile) {
+				helpText = this.$fs.readText(markdownFile).wait();
+			} else {
```
Fatme: `else` is unneeded. Just `if(markdownFile) {
				return this.$fs.readText(markdownFile).wait(); 
}
this.$errors.failWithoutHelp("Unknown command '%s'. Try '$ %s help' for a full list of supported commands.", mdFileName.replace(".md", ""), this.$staticConfig.CLIENT_NAME.toLowerCase());
`

services/html-help-service.ts
rosen-vladimirov: Ahh, I've got it - as I want to preserve same directory structure, I need the replace method. If I use your suggestion, all html files will be placed inside html directory.

services/html-help-service.ts
```diff
@@ -0,0 +1,158 @@
...
+			this.$logger.trace("Found page to open: '%s'", pageToOpen);
+			this.$opener.open(pageToOpen);
+			return true;
+		} 
+		
+		this.$logger.trace("Unable to find file: '%s'", htmlPage);
+		return false;
+	}
+
+	private readManFileForCommand(commandName: string): IFuture<string> {
```
teobugslayer: Maybe the better name is "readMdFileForCommand", thoughts?

teobugslayer:
Cool, :+1:

PR #245, opened 2015-03-26
Supports for dynamic commands

Fatme:


definitions/commands-service.d.ts
```diff
@@ -3,4 +3,9 @@ interface ICommandsService {
 	tryExecuteCommand(commandName: string, commandArguments: string[]): IFuture<void>;
 	executeCommandUnchecked(commandName: string, commandArguments: string[]): IFuture<boolean>;
 	completeCommand(): IFuture<boolean>;
+}
+
+interface ICommandsServiceProvider {
+	allDynamicCommands(): IFuture<string[]>;
```
rosen-vladimirov: getDynamicCommands sounds better for me

services/commands-service.ts
```diff
@@ -104,6 +105,11 @@ export class CommandsService implements ICommandsService {
 
 				this.$errors.fail("Unable to execute command '%s'. Use '$ %s %s --help' for help.", beautifiedName, this.$staticConfig.CLIENT_NAME.toLowerCase(), beautifiedName);
 				return false;
+			} else {
+				if(_.any(this.$commandsServiceProvider.allDynamicCommands())) {
```
rosen-vladimirov: maybe this should be ```if(! _.any(this.$commandsServiceProvider.allDynamicCommands()))```

rosen-vladimirov: if you use my code, what will happen in case there aren't any dynamic commands :cat:

services/commands-service.ts
```diff
@@ -104,6 +105,11 @@ export class CommandsService implements ICommandsService {
 
 				this.$errors.fail("Unable to execute command '%s'. Use '$ %s %s --help' for help.", beautifiedName, this.$staticConfig.CLIENT_NAME.toLowerCase(), beautifiedName);
 				return false;
+			} else {
```
teobugslayer: You can simplify the code a bit writing ```else if``` (without nesting). Considering that everything in the if returns, the else is also not strictly needed, but I'd rather leave it so that future refactoring does not break easily.

PR #247, opened 2015-03-27
Use inquirer instead of prompter

Fatme:


definitions/prompt.d.ts
```diff
@@ -1,41 +1,25 @@
...
-}
-
-interface IPromptHistoryValue {
-	value: any;
+	type: string;
+	name: string;
+	message: any;
+	default?: any;
+	choices?: any[];
+	//validate?: (userInput: any) => any;
```
teobugslayer: Why is this commented?

prompter.ts
```diff
@@ -33,27 +33,22 @@ export class Prompter implements IPrompter {
...
 		var future = new Future;
-		prompt.get(schema, (err: any, result: any) => {
-			if (err) {
+		prompt.prompt(schema, (result: any) => {
+			if(result) {
+				future.return(result);
+			} else {
 				// the error is normally Ctrl-C:
 				// move the cursor after the prompt (so that the error is not printed right after the user input)
 				console.log("");
```
teobugslayer: Do we need this console log yet?

teobugslayer:
:+1:

jasssonpet:
I think, that the package.json should be updated too.

PR #248, opened 2015-03-30
Fix marked-terminal parsing

rosen-vladimirov:
Remove the hack when parsing markdown text that has to be show on the console. The problem was with code in table, but it had been fixed in version 1.3.0 of marked-terminal.

Part of http://teampulse.telerik.com/view#item/259417


teobugslayer:
:+1:

teobugslayer:
:+1:

PR #250, opened 2015-03-30
Logging

jasssonpet:
Introduce `--logging` flag for `emulate`, `run`, `debug` and `device run` commands that outputs the standard output/logcat to the running terminal.

### DIFF
--- a/mobile/android/android-device.ts
+++ b/mobile/android/android-device.ts
@@ -173,6 +173,10 @@ export class AndroidDevice implements Mobile.IAndroidDevice {
 		return (() => {
 			var startPackageCommand = this.composeCommand("shell am start -a android.intent.action.MAIN -n %s/%s", packageName, this.$staticConfig.START_PACKAGE_ACTIVITY_NAME);
 			this.$childProcess.exec(startPackageCommand).wait();
+
+			if (options.printAppOutput) {
+				this.openDeviceLogStream();
+			}
 		}).future<void>()();
 	}
 
--- a/mobile/ios/ios-core.ts
+++ b/mobile/ios/ios-core.ts
@@ -14,6 +14,10 @@ import net = require("net");
 import util = require("util");
 import Future = require("fibers/future");
 import bplistParser = require("bplist-parser");
+import string_decoder = require("string_decoder");
+import stream = require("stream");
+import options = require("../../options");
+import assert = require("assert");
 
 export class CoreTypes {
 	public static pointerSize = ref.types.size_t.size;
@@ -867,7 +871,7 @@ class PosixSocket implements Mobile.IiOSDeviceSocket {
 							}
 						}
 					} catch (e) {
-						this.$logger.trace("Exeption thrown!!!");
+						this.$logger.trace("Exception thrown: " + e);
 					}
 				} else if (this.format === CoreTypes.kCFPropertyListXMLFormat_v1_0) {
 					try {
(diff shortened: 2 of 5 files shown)


mobile/ios/ios-core.ts
```diff
@@ -980,30 +983,80 @@ export class PlistService implements Mobile.IiOSDeviceSocket {
 	}
 }
 
-export class GDBServer implements Mobile.IGDBServer {
-	private socket: Mobile.IiOSDeviceSocket  = null;
+util.inherits(GDBStandardOutputAdapter, stream.Transform);
+function GDBStandardOutputAdapter(opts?: stream.TransformOptions) {
```
Fatme: Can we do this with class - for example `class GDBStandardOutputAdapter extends stream.Transform`?

jasssonpet: No, the `inherits` function sets a `super_`, require by node.

mobile/ios/ios-core.ts
```diff
@@ -980,30 +983,80 @@ export class PlistService implements Mobile.IiOSDeviceSocket {
...
-		private $injector: IInjector) {
+	for (var i = 0; i < packet.length; i++) {
+		if (packet[i] === "$".charCodeAt(0)) {
+			var start = ++i;
 
-		if(hostInfo.isWindows()) {
-			this.socket = this.$injector.resolve(WinSocket, {service: this.service, format: 0});
-		} else if(hostInfo.isDarwin()) {
-			this.socket = this.$injector.resolve(PosixSocket, {service: this.service, format: CoreTypes.kCFPropertyListXMLFormat_v1_0});
+			while (packet[i] !== "#".charCodeAt(0)) {
```
Fatme: I suggest to extract this to separate function - `getUnicode(ch: string)` and can also assert the length of `ch`.

jasssonpet: Well, I need `i` to be moved also, so there should be an out parameter and I think it'll get more ugly/slow.

Fatme: You don't need out parameter. You can do:
`getUnicode(ch: string) { 
    return charCodeAt(0);
}`
`var unicode = this.getUnicode("#");
 while(packet[i] !== unicode) { .. }`

Even we can make `unicode` static variable.

jasssonpet: OK, I was thinking about something else. Where would be a good place to store this function?

Fatme: It should be accessable from `GDBStandardOutputAdapter` and `GDBSignalWatcher` so I suggest to make it internal for `ios-core` file

mobile/ios/ios-core.ts
```diff
@@ -867,7 +870,7 @@ class PosixSocket implements Mobile.IiOSDeviceSocket {
 							}
 						}
 					} catch (e) {
-						this.$logger.trace("Exeption thrown!!!");
+						this.$logger.trace("Exception thrown!!!");
```
teobugslayer: Consider showing the exception as well - this will aid in debugging and customer support.

teobugslayer: I suggest that we also log the exception.

mobile/ios/ios-core.ts
teobugslayer: Can't we call super_ in the constructor of the typescript class?

mobile/ios/ios-core.ts
```diff
@@ -980,30 +983,80 @@ export class PlistService implements Mobile.IiOSDeviceSocket {
...
-		var encodedArguments = _.map(argv, (arg, index) => util.format("%d,%d,%s", arg.length*2, index, new Buffer(arg).toString("hex"))).join(",");
-		this.send("A"+encodedArguments);
+		var encodedArguments = _.map(argv, (arg, index) => util.format("%d,%d,%s", arg.length * 2, index, new Buffer(arg).toString("hex"))).join(",");
+		this.send("A" + encodedArguments);
 
 		this.send("qLaunchSuccess");
-		this.send("vCont;c");
+
+		if(options.logging) {
+			this.socket.pipe(new (<any>GDBStandardOutputAdapter)()).pipe(process.stdout);
```
teobugslayer: Can we cast to something more specific? Like Stream or EventEmitter?

mobile/ios/ios-core.ts
jasssonpet: Made it to log the exception.

mobile/ios/ios-core.ts
jasssonpet: I have extracted the function.

mobile/android/android-device.ts
```diff
@@ -173,6 +173,10 @@ export class AndroidDevice implements Mobile.IAndroidDevice {
 		return (() => {
 			var startPackageCommand = this.composeCommand("shell am start -a android.intent.action.MAIN -n %s/%s", packageName, this.$staticConfig.START_PACKAGE_ACTIVITY_NAME);
 			this.$childProcess.exec(startPackageCommand).wait();
+
+			if (options.logging) {
```
teobugslayer: Do you think that it is possible for the users to confuse ```--log``` and ```--logging```? If so, can we find a better name for one of the options? For example, the user must write
```$ tns deploy ios --logging --log trace```

Comment (jasssonpet) [inline mobile/android/android-device.ts]: It's totally possible to confuse them. Lets come up with a better name.

Ionic uses `--consolelogs`.

File: mobile/ios/ios-core.ts (hunk 4 above)
Comment (jasssonpet) [inline mobile/ios/ios-core.ts]: Well, I tried all kinds of combinations (`Transform`, `Stream`, `typeof Transform`), but couldn't make it work. My TypeScript skills are not good, though :)

File: mobile/android/android-device.ts (hunk 5 above)
Comment (teobugslayer) [inline mobile/android/android-device.ts]: @ikoevska, opinions? On the top of my mind, ```--appstdio```, ```--stdout```, ```--appoutput```

File: mobile/ios/ios-core.ts [hunk 6]
```diff
@@ -980,44 +984,100 @@ export class PlistService implements Mobile.IiOSDeviceSocket {
...
+	return code;
+}
+
+util.inherits(GDBStandardOutputAdapter, stream.Transform);
+function GDBStandardOutputAdapter(opts?: stream.TransformOptions) {
+	stream.Transform.call(this, opts);
+	this.utf8StringDecoder = new string_decoder.StringDecoder("utf8");
+}
+GDBStandardOutputAdapter.prototype._transform = function(packet: Buffer, encoding: string, done: Function): void {
+	var result = "";
```
teobugslayer: This function throws (at least, assert.ok throws, but I assume other calls can throw as well). Should we add a try / catch and call ```done``` with the exception?

mobile/ios/ios-core.ts
```diff
@@ -980,44 +984,100 @@ export class PlistService implements Mobile.IiOSDeviceSocket {
...
+
+	for (var i = 0; i < packet.length; i++) {
+		if (packet[i] === getCharacterCodePoint("$")) {
```
teobugslayer: This comparison happens on plenty of places. What is your opinion about a helper function
``` codePointCompare(byte: number, char: string) => byte === getCharacterCodePoint(char)``` ?

File: mobile/ios/ios-core.ts (hunk 4 above)
Comment (teobugslayer) [inline mobile/ios/ios-core.ts]: I checked the type definitions, and it seems that pipe is declared ```pipe<T extends NodeJS.WritableStream>(destination: T, options?: { end?: boolean; }): T;``` I would try with \<NodeJS.WritableStream\>

Comment (jasssonpet) [inline mobile/ios/ios-core.ts]: I tried with `NodeJS.WritableStream`: TS2351: Cannot use 'new' with an expression whose type lacks a call or construct signature.

File: mobile/ios/ios-core.ts [hunk 8]
```diff
@@ -980,44 +984,100 @@ export class PlistService implements Mobile.IiOSDeviceSocket {
...
+
+util.inherits(GDBSignalWatcher, stream.Writable);
+function GDBSignalWatcher(opts?: stream.WritableOptions) {
+	stream.Writable.call(this, opts);
+}
+GDBSignalWatcher.prototype._write = function(packet: Buffer, encoding: string, callback: Function) {
+	for (var i = 0; i < packet.length - 2; i++) {
+		if (packet[i] === getCharacterCodePoint("$") && (packet[i + 1] === getCharacterCodePoint("T") || packet[i + 1] === getCharacterCodePoint("S"))) {
+			// SIGKILL
+			if (packet[i + 2] === getCharacterCodePoint("9")) {
```
teobugslayer: If i is packet.length -2, then this 'if' will read after the buffer end. Is this possible?

jasssonpet: Correct me if I'm wrong, but:
```
i < arr.length     - safe to access arr[i]
i < arr.length - 1 - safe to access arr[i + 1]
i < arr.length - 2 - safe to access arr[i + 2]
```

teobugslayer: I feel stupid, I was absolutely positively sure that the code read ```i <= packet.length -2``` I am not sure where my brain conjured that = from. Sorry for the noise.

mobile/ios/ios-core.ts
teobugslayer: I see - the issue is that TypeScript sees GDBStandardOutputAdapter as a function but it expects a class. You are hacking its brains! Unfortunately for me, \<any\> will stay, unless we convert GDBStandardOutputAdapter to a proper class

mobile/android/android-device.ts
ikoevska: `--appLog`, `--appOutput`, `--outputAppLog`, `--streamAppLog`, `--streamAppOutput` are all viable options.

I am more inclined towards one of those: `--outputAppLog`, `--streamAppLog`, `--streamAppOutput` as they are more intuitive and self-explanatory => minimal room for confusion.

mobile/ios/ios-core.ts
jasssonpet: Added a `try/catch`.

mobile/ios/ios-core.ts
```diff
@@ -980,44 +984,104 @@ export class PlistService implements Mobile.IiOSDeviceSocket {
...
 
-		if(hostInfo.isWindows()) {
-			this.socket = this.$injector.resolve(WinSocket, {service: this.service, format: 0});
-		} else if(hostInfo.isDarwin()) {
-			this.socket = this.$injector.resolve(PosixSocket, {service: this.service, format: CoreTypes.kCFPropertyListXMLFormat_v1_0});
+	return code;
+}
+
+util.inherits(GDBStandardOutputAdapter, stream.Transform);
+function GDBStandardOutputAdapter(opts?: stream.TransformOptions) {
```
teobugslayer: We just verified with @fatme, and it seems that indeed, this can be written as a class. You need to use this code:

```
class GDBStandardOutputAdapter extends stream.Transform {
	constructor(opts?: stream.TransformOptions) {
		super(opts);
		util.inherits(GDBStandardOutputAdapter, stream.Transform);

		this.utf8StringDecoder = new string_decoder.StringDecoder("utf8");
	}

	_transform(packet: any, encoding: string, done: Function): void {
		// your code for _transform here
                // note: _transform could be called with either Buffer or string and you need to either add a check or convert to the type you expect
	}

```

mobile/android/android-device.ts
jasssonpet: I like `--printAppLog`.

mobile/ios/ios-core.ts
jasssonpet: Made `GDBStandardOutputAdapter` a proper class.

mobile/android/android-device.ts
jasssonpet: Changed it to `printAppOutput`.

PR #251, opened 2015-03-31
Fix livesync

Fatme:
http://teampulse.telerik.com/view#item/287903

Post notification logic has been changed. With current implementation we try to receive all data from socket before sending the data is started. And this is ok on `OSX` but on `Windows` `receive` function is blocking and livesync command is not able to complete.
I've introduced two functions - one for debugger purposes and one that only sends the notification.

### DIFF
--- a/mobile/ios/ios-proxy-services.ts
+++ b/mobile/ios/ios-proxy-services.ts
@@ -220,7 +220,7 @@ export class InstallationProxyClient {
 }
 $injector.register("installationProxyClient", InstallationProxyClient);
 
-export class NotificationProxyClient {
+export class NotificationProxyClient implements Mobile.INotificationProxyClient {
 
 	private plistService: Mobile.IiOSDeviceSocket = null;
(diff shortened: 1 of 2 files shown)


mobile/ios/ios-proxy-services.ts
```diff
@@ -301,6 +301,14 @@ export class NotificationProxyClient {
 		}
 	}
 
+	private postNotificationCore(notificationName: string): void {
+		this.plistService.sendMessage({
```
rosen-vladimirov: is it possible plistService to be null here? For example when calling postNotificationAndAttachForData who will initialize plistService?

Fatme: No, it is not possible. When `postNotificationAndAttachForData` is called we initialize `plistService` inside `openSocket` function - https://github.com/telerik/mobile-cli-lib/pull/251/files#diff-b2b4fa0509e169323e6cd005ec98dc9bL234

rosen-vladimirov:
:+1:

PR #252, opened 2015-03-31
Android logging

jasssonpet:


mobile/android/LogcatHelper.ts
```diff
@@ -0,0 +1,59 @@
+///<reference path="../../../.d.ts"/>
+"use strict";
+import byline = require("byline");
+
+function getConsoleLogFromLine(lineText:String):any {
+	var acceptedTags = ["chromium", "Web Console", "JS"];
+
+	//sample line is "I/Web Console(    4438): Received Event: deviceready at file:///storage/emulated/0/Icenium/com.telerik.TestApp/js/index.js:48"
+	var match = lineText.match(/.\/(.+?)\s*\(\s*(\d+?)\): (.*)/);
```
jasssonpet: Added `\s*` to handle `JS     (` case.

teobugslayer: I think we could move the RegExp out so that it evaluates only once.

mobile/android/LogcatHelper.ts
```diff
@@ -0,0 +1,59 @@
+///<reference path="../../../.d.ts"/>
+"use strict";
+import byline = require("byline");
+
+function getConsoleLogFromLine(lineText:String):any {
```
teobugslayer: Is there a reason for moving this function out of the class?

mobile/android/LogcatHelper.ts
```diff
@@ -0,0 +1,59 @@
...
+	else if(_.any(acceptedTags, (tag:string) => {
+			return lineText.indexOf(tag) !== -1;
+		})) {
+		return {message: match[3]};
+	}
+
+	return null;
+}
+
+class LogcatHelper {
```
teobugslayer: Consider creating an interface ILogcatHelper and implement it with this class. resolutions through the injector should deal with this interface only.

mobile/android/LogcatHelper.ts
jasssonpet: Moved the regex to a field.

mobile/android/LogcatHelper.ts
jasssonpet: Moved the function in the class.

mobile/android/LogcatHelper.ts
jasssonpet: Created `ILogcatHelper` which the injector uses.

mobile/android/LogcatHelper.ts
```diff
@@ -0,0 +1,59 @@
...
+			}
+		} else if(_.any(acceptedTags, (tag: string) => { return lineText.indexOf(tag) !== -1; })) {
+			return {message: match[3]};
+		}
+
+		return null;
+	}
+}
+
+export = LogcatHelper;
```
Fatme: You don't need this. You can use `export class LogcatHelper {`

mobile/android/android-device.ts
```diff
@@ -4,11 +4,11 @@ import util = require("util");
 import Future = require("fibers/future");
 import path = require("path");
 import temp = require("temp");
-import byline = require("byline");
 import helpers = require("../../helpers");
 import os = require("os");
 import options = require("./../../options");
 import Fiber = require("fibers");
+import LogcatHelper = require("./LogcatHelper");
```
Fatme: `It is better to register `logcatHelper` to `injector`.
You need to call `$injector.require("logcatHelper", LogcatHelper)` at the end of `LogcatHelper.ts` file
and also you need to register it in bootstarap -> `$injector.register("logcatHelper", "./common/mobile/android/LogcatHelper.ts")`

mobile/android/android-device.ts
```diff
@@ -64,7 +64,8 @@ export class AndroidDevice implements Mobile.IAndroidDevice {
 		private $errors: IErrors,
 		private $staticConfig: Config.IStaticConfig,
 		private $opener: IOpener,
-		private $devicePlatformsConstants: Mobile.IDevicePlatformsConstants) {
+		private $devicePlatformsConstants: Mobile.IDevicePlatformsConstants,
+		private $injector: IInjector) {
```
Fatme: You can use `private $logcatHelper: ILogcatHelper) {`

mobile/android/LogcatHelper.ts
jasssonpet: Changed.

mobile/android/android-device.ts
jasssonpet: Moved it to `$injector`. Thanks for the exhaustive help :)

mobile/android/android-device.ts
jasssonpet: Replaced `$injector` with `$logcatHelper`.

bootstrap.ts
```diff
@@ -61,6 +61,7 @@ $injector.require("iOSDeviceDiscovery", "./common/mobile/mobile-core/device-disc
 $injector.require("androidDeviceDiscovery", "./common/mobile/mobile-core/device-discovery");
 $injector.require("iOSDevice", "./common/mobile/ios/ios-device");
 $injector.require("androidDevice", "./common/mobile/android/android-device");
+$injector.register("logcatHelper", "./common/mobile/android/logcat-helper");
```
Fatme: This should be `require` :smile_cat:

jasssonpet: nice catch

definitions/mobile.d.ts
```diff
@@ -38,6 +38,10 @@ declare module Mobile {
 		startService(serviceName: string): number;
 	}
 
+	interface ILogcatHelper {
+		start(deviceIdentifier: string, adbPath: string): any;
```
Fatme: Hm, it's a little bit strange to have `adbPath` as parameter. `start` function can resolve it internal. We can get it from `helpers.getPathToAdb()`

Fatme: This is just a suggestion. I don't insist to implement it in this PR.

Fatme:
:+1:

PR #253, opened 2015-04-01
Fix installation and uninstall of CLI

rosen-vladimirov:
Add dev-preuninstall command which is used to execute preuninstall actions. It should be referenced from each CLI in its preuninstall script. Fix KillSwitches dir to be under our profile-dir directory.

### DIFF
--- a/commands/preuninstall.ts
+++ b/commands/preuninstall.ts
@@ -1,0 +1,37 @@
+﻿///<reference path="../../.d.ts"/>
+"use strict";
+
+import options = require("../options");
+import path = require("path");
+import util = require("util");
+
+export class PreUninstallCommand implements ICommand {
+	private static ADB_RELATIVE_PATH = "../resources/platform-tools/android/%s/adb";
+
+	constructor(private $fs: IFileSystem,
+		private $childProcess: IChildProcess,
+		private $logger: ILogger) { }
+	public disableAnalytics = true;
+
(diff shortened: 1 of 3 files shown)


commands/preuninstall.ts
```diff
@@ -0,0 +1,33 @@
...
+	public disableAnalytics = true;
+
+	public allowedParameters: ICommandParameter[] = [];
+
+	public execute(args: string[]): IFuture<void> {
+		return (() => {
+			var relativeAdbPath = util.format(PreUninstallCommand.ADB_RELATIVE_PATH, process.platform);
+			var adbPath = path.join(__dirname, relativeAdbPath);
+			var executableAdbPath = process.platform === "win32" ? adbPath + ".exe" : adbPath;
+			if(!this.$fs.exists(executableAdbPath).wait()) {
```
teobugslayer: I think this check is not needed. Consider deleting it.

commands/preuninstall.ts
```diff
@@ -0,0 +1,33 @@
...
+				// in some older installations adb was not in common
+				adbPath = path.join(__dirname, "..", "..", relativeAdbPath);
+			}
+
+			var killAdbServerCommand = util.format("\"%s\" kill-server", adbPath);
```
teobugslayer: I think different clean-up operation should be guarded in try/catch, so if one of them fails, others have chance to execute

Fatme:
:+1:

PR #254, opened 2015-04-01
Fix warning for missing androis sdk on mac

teobugslayer:
The issue is that `android -h` returns exit code 1, which is unconventional and inconvenient way of working for a unix command-line tool

### DIFF
--- a/sysinfo.ts
+++ b/sysinfo.ts
@@ -2,13 +2,15 @@
 "use strict";
 
 import os = require("os");
+import child_process = require("child_process");
 import osenv = require("osenv");
 import Future = require("fibers/future");
 import hostInfo = require("./host-info");
 
 export class SysInfo implements ISysInfo {
 	constructor(private $childProcess: IChildProcess,
-				private $iTunesValidator: Mobile.IiTunesValidator) { }
+				private $iTunesValidator: Mobile.IiTunesValidator,
+				private $logger: ILogger) { }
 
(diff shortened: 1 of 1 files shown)


sysinfo.ts
```diff
@@ -64,6 +66,26 @@ export class SysInfo implements ISysInfo {
 		} // if we got an error, assume not working
 	}
 
+	// `android -h` returns exit code 1 on successful invocation. Therefore, we cannot use $childProcess
+	private execAndroidH(): string {
```
rosen-vladimirov: can you change the return type to IFuture<string> and return the future you create on the next line? I suggest this only because we stick to the idea each `.wait` to be wrapped in IFuture :)

sysinfo.ts
```diff
@@ -64,6 +66,26 @@ export class SysInfo implements ISysInfo {
 		} // if we got an error, assume not working
 	}
 
+	// `android -h` returns exit code 1 on successful invocation. Therefore, we cannot use $childProcess
```
rosen-vladimirov: Is this unexpected behavior detected on MacOS only? I've tried to reproduce it on Windows, but without success.

rosen-vladimirov:
One suggestion, anyway :+1:  Great finding!

sysinfo.ts
teobugslayer: Yes, it happens on Mac. I do not know whether Google will make it on other OSes so that I do not make OS check.

PR #264, opened 2015-04-06
Fix errors print in debug

rosen-vladimirov:
In case you are using Node 0.12.x and in config.json "debug" is set to true, the real error message is not printed to the output. When we are in debug mode, we print only the callstack and in Node 0.10.x , when fibers throw exception, the top row of the callstack is the error message. With Node 0.12.x, when error is thrown from fibers, the message is not part of the callstack for some reason. Workaround this behavior by checking the callstack and add the message in case it is missing.

### DIFF
--- a/errors.ts
+++ b/errors.ts
@@ -9,8 +9,8 @@ import helpers = require("./helpers");
 function Exception() {}
 Exception.prototype = new Error();
 
-function resolveCallStack(stack: string): string {
-	var stackLines: string[]= stack.split("\n");
+function resolveCallStack(error: Error): string {
+	var stackLines: string[]= error.stack.split("\n");
 	var parsed = _.map(stackLines, (line: string): any => {
 		var match = line.match(/^\s*at ([^(]*) \((.*?):([0-9]+):([0-9]+)\)$/);
 		if (match) {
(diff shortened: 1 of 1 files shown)


rosen-vladimirov:
Maybe I should include this fix in release branch? What do you think? @teobugslayer , @Fatme

errors.ts
```diff
@@ -54,14 +54,20 @@ function resolveCallStack(stack: string): string {
 		return util.format("    at %s (%s:%s:%s)", functionName, source, sourcePos.line, sourcePos.column);
 	});
 
-	return remapped.join("\n");
+	var outputMessage = remapped.join("\n");
+	if(outputMessage && outputMessage.indexOf(message) === -1) {
```
ligaz: Is there a need for the `&&`? You can just use the condition with `indexOf`.

errors.ts
```diff
@@ -9,7 +9,7 @@ import helpers = require("./helpers");
 function Exception() {}
 Exception.prototype = new Error();
 
-function resolveCallStack(stack: string): string {
+function resolveCallStack(stack: string, message: string): string {
```
ligaz: Why not pass the `Error` object here instead of two parameters?

Fatme:
I think that we should include the fix in release branch.

Fatme:
:+1:

PR #267, opened 2015-04-06
Support Google API images in Android Emulator.

teobugslayer:
See https://github.com/NativeScript/nativescript-cli/issues/175

### DIFF
--- a/mobile/android/android-emulator-services.ts
+++ b/mobile/android/android-emulator-services.ts
@@ -365,7 +365,11 @@ class AndroidEmulatorServices implements Mobile.IEmulatorPlatformServices {
 		var platform = target.replace('android-', '');
 		var platformNumber = +platform;
 		if (isNaN(platformNumber)) {
-			if (platform === "L") {
+			// this might be a google image
+			platform = target.split(":")[2];
+			if (platform) {
+				platformNumber = +platform;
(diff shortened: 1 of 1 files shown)


mobile/android/android-emulator-services.ts
```diff
@@ -365,7 +365,11 @@ class AndroidEmulatorServices implements Mobile.IEmulatorPlatformServices {
 		var platform = target.replace('android-', '');
 		var platformNumber = +platform;
 		if (isNaN(platformNumber)) {
-			if (platform === "L") {
+			// this might be a google image
+			platform = target.split(":")[2];
+			if (platform) {
+				platformNumber = +platform;
```
rosen-vladimirov: What will happen in case this is some strange image and after splitting by ":" we get string or undefined for platform. +ing it will set platformNumber ot NaN and the code will not work

teobugslayer: Well, if is a strange image, the code won't work anyway. Right now it returns undefined and the maps() after that does not find a suitable image and we give a message "no image found". I can make it throw an exception but  this ultimately would give the same message to the user, right?

rosen-vladimirov: Ok, I didn't check the rest of the code, sorry.

rosen-vladimirov:
:+1:  awesome :smile_cat:

PR #270, opened 2015-04-07
Improve autocompletion

rosen-vladimirov:
Move autocompletion code to a new shell file. In .bash_profile, .bashrc and .zshrc add code that executes this code when we want to enable autocompletion.
Autocomplete is now hierarchical command. The subcommands and their behavior is:
* `autocomplete` - checks if autocompletion is enabled. In case it is not, ask the user for confirmation. In case autocompletion is enabled, but it is the old one (where we've written some data to .bash_profile, etc.), silently remove the old code and use the new approach. In case the user chooses to not use autocompletion, call disableAutoCompletion in order to remove our code from all shell profiles.
* `autocomplete enable` - enables autocompletion in case it is not already enabled.
* `autocomplete disable` - disables autocompletion in case it is not already disabled.
* `autocomplete status` - prints the current status of autocompletion.


services/auto-completion-service.ts
```diff
@@ -3,33 +3,176 @@
...
+
+			return false;
+		}).future<boolean>()();
+	}
+
+	private addAutoCompletionToShellScript(fileName: string): IFuture<void> {
 		return (() => {
 			try {
-				this.updateShellScriptCore(fileName).wait();
+				if(!this.isNewAutoCompletionEnabledInFile(fileName).wait() || this.isObsoleteAutoCompletionEnabledInFile(fileName).wait()) {
```
teobugslayer: This can be simplified to a single return

services/auto-completion-service.ts
```diff
@@ -3,33 +3,176 @@
...
+			}
+
+			return false;
+		}).future<boolean>()();
+	}
+
+	private isObsoleteAutoCompletionEnabledInFile(fileName: string): IFuture<boolean> {
+		return (() => {
+			var text = this.$fs.readText(fileName).wait();
+			if(text.match(this.getTabTabObsoleteRegex(this.$staticConfig.CLIENT_NAME)) || text.match(this.getTabTabObsoleteRegex(this.$staticConfig.CLIENT_NAME))) {
```
teobugslayer: This can be simplified to a single return

teobugslayer:
:+1: this is so awesome! This must be bolded and italicized in the release notes. And mentioned twice, for good measure!

PR #271, opened 2015-04-07
Fix commands parameters validation

Fatme:
"Returns - An array containing the DELETED element" - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/splice. 
We need to remove the element and are not interested in deleted element :smile_cat:


rosen-vladimirov:
:+1:

PR #276, opened 2015-04-08
Add getString method to the prompter

Mitko-Kerezov:
In most cases one only needs to prompt the user for a single input.
I exposed a getString method similar to getPassword for facilitation.

promptForChoice on the other hand asks the user to choose from a list of options.

### DIFF
--- a/declarations.d.ts
+++ b/declarations.d.ts
@@ -183,8 +183,6 @@ interface IDictionary<T> {
 	[key: string]: T
 }
 
-interface IStringDictionary extends IDictionary<string> { }
-
 interface IAnalyticsService {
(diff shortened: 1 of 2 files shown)


declarations.d.ts
```diff
@@ -183,8 +183,6 @@ interface IDictionary<T> {
 	[key: string]: T
 }
 
-interface IStringDictionary extends IDictionary<string> { }
```
rosen-vladimirov: Why?

Mitko-Kerezov: It's duplicated :point_up:

teobugslayer:
:+1:

PR #283, opened 2015-04-15
Add generic remove function to helpers

Mitko-Kerezov:
Used to remove elements from an array.
Contrary to `_.remove` removes only one element matching the predicate (if not explicitly specified otherwise)


teobugslayer:
:+1:

PR #286, opened 2015-04-16
Use info-zip based unzip on windows, and OS supplied one on linux

teobugslayer:
and darwin

Because info-zip cannot find the zip file case-insensitively, we have to manually find the correct one and feed it to it.

Remove obsoleted $staticConfig.sevenZipFilePath


rosen-vladimirov:
:+1:

PR #295, opened 2015-04-20
Fix wrong message when livesyncing and remove unneeded function

Fatme:
http://teampulse.telerik.com/view#item/290210


rosen-vladimirov:
:+1:

teobugslayer:
We agreed to add a custom message

PR #297, opened 2015-04-20
Fix markdown links in console

rosen-vladimirov:
Links in markdown files should be stripped in console help. Fix the incorrect regex, so now the correct text between [ and ] will be shown.
Fixes: http://teampulse.telerik.com/view#item/290105


teobugslayer:
:+1:

teobugslayer:
I could not find tests for the htmlHelpService - maybe we should write some.

rosen-vladimirov:
You are correct, its in my TODO list, but I postponed writing tests as we had some higher priority tasks

PR #302, opened 2015-04-23
Enable autocompletion for dynamic commands

Fatme:
http://teampulse.telerik.com/view#item/290527

### DIFF
--- a/services/commands-service.ts
+++ b/services/commands-service.ts
@@ -25,8 +25,8 @@ export class CommandsService implements ICommandsService {
 		private $commandsServiceProvider: ICommandsServiceProvider) {
 	}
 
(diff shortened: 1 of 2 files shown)


services/commands-service.ts
```diff
@@ -239,7 +239,11 @@ export class CommandsService implements ICommandsService {
 				}
 
 				if(data.words === 1) {
-					return tabtab.log(this.allCommands(false), data);
+					var allCommands = this.allCommands(false);
```
teobugslayer: Can we change false to consts / enum? In TypeScript 1.4, there is 'const enum' which is quite less painful than the full enum.

rosen-vladimirov:
:+1:  from me :)

PR #304, opened 2015-04-23
Show application output by default

Fatme:
Until now the application output was printed only if the --printAppOutput option was specified. This PR changes this behavior and the application output is printed by default. The new option - justlaunch is introduced and if it is specified the application output is not printed. 

Also this PR removes the filtration of android logcat for projects created with nativescript-cli.

### DIFF
--- a/mobile/android/logcat-helper.ts
+++ b/mobile/android/logcat-helper.ts
@@ -1,16 +1,17 @@
 ///<reference path="../../../.d.ts"/>
 "use strict";
 import byline = require("byline");
+import util = require("util");
 
 export class LogcatHelper implements Mobile.ILogcatHelper {
-	private lineRegex = /.\/(.+?)\s*\(\s*(\d+?)\): (.*)/;
-
 	constructor(private $childProcess: IChildProcess,
-		    private $logger: ILogger) {
+		    private $logcatPrinter: Mobile.ILogcatPrinter,
+			private $logger: ILogger){
 
 	}
 
 	public start(deviceIdentifier: string, adbPath: string): any {
+		this.$childProcess.exec(util.format("%s -s %s logcat -c", adbPath, deviceIdentifier)).wait();
 		var adbLogcat = this.$childProcess.spawn(adbPath, ["-s", deviceIdentifier, "logcat"]);
 		var lineStream = byline(adbLogcat.stdout);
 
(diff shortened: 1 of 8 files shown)


fealebenpae:
:+1:

mobile/android/logcat-helper.ts
```diff
@@ -1,16 +1,17 @@
...
-
 	constructor(private $childProcess: IChildProcess,
-		    private $logger: ILogger) {
+		    private $logcatPrinter: Mobile.ILogcatPrinter,
+			private $logger: ILogger){
 
 	}
 
 	public start(deviceIdentifier: string, adbPath: string): any {
+		this.$childProcess.exec(util.format("%s -s %s logcat -c", adbPath, deviceIdentifier)).wait();
```
rosen-vladimirov: why not using the below variable- adbLogcat?

Fatme: `adbLogcat` is childProcess object. Here we need just a string.

options.ts
```diff
@@ -31,7 +31,7 @@ var knownOpts: any = {
 	"start": Boolean,
 	"stop": Boolean,
 	"ddi": String, // the path to developer  disk image
-	"print-app-output": Boolean
+	"justlaunch": Boolean
```
rosen-vladimirov: why isn't it just-launch?

Fatme: good question! :smile:

mobile/android/logcat-helper.ts
rosen-vladimirov: can you write a comment why we have two different ways to spawn/exec adb logcat

mobile/ios/ios-proxy-services.ts
```diff
@@ -392,13 +391,17 @@ export class IOSSyslog {
...
+		var shouldLog = false;
+		setTimeout(() => shouldLog = true, 2500);
+
 		var printData = (data: NodeBuffer) => {
-			var output = ref.readCString(data, 0);
-			if(this.matchRegex.test(output)) {
-				this.$logger.out(output);
+			if(shouldLog) {
+				var output = ref.readCString(data, 0);
+				process.stdout.write(output);
```
rosen-vladimirov: why using process.stdout.write?

options.ts
Fatme: @ikoevska, @teobugslayer, @DimitarTodorov What do you think?

mobile/ios/ios-proxy-services.ts
Fatme: Because if we use `logger.out` we produce two empty lines on console. Actually `output` ends with `\n` and `logger.out` adds one more `\n`. `process.stdout.write` doesn't add additional empty line. :smile:

rosen-vladimirov: You can use `$logger.write`
https://github.com/telerik/mobile-cli-lib/blob/f1a6b5d4c1c4b3d632d0f95a1b4c053e1768eaa4/logger.ts#L78

teobugslayer:
:+1: as well

rosen-vladimirov:
:+1:

options.ts
ikoevska: justLaunch is probably the way to go. I've gone through our nativescript cli options and for the most part we are using camelCase for our options - e.g. availableDevices keyStore*, frameworkPath, forDevice

Two of the few exceptions to this rule I've found are debug-brk and no-client

ikoevska:
Where are the docs?

Fatme:
The docs is in nativescript-cli repo - https://github.com/NativeScript/nativescript-cli/commit/b813667e20bde9e7f4f45c02dc1382b027bf1582

PR #309, opened 2015-04-28
Fixes validation of hierarchical commands with too many arguments

teobugslayer:
Fixes http://teampulse.telerik.com/view#item/290977


rosen-vladimirov:
:+1:

PR #311, opened 2015-04-28
Fix command line validation logic

teobugslayer:
This prevents "swallowing" of arguments, due to Array.splice()

See http://teampulse.telerik.com/view#item/290977


rosen-vladimirov:
:+1:

Fatme:
:+1:

PR #312, opened 2015-04-29
Prevent doubling of typescript definitions

rosen-vladimirov:
CLI has its own typescript definition files, which are used when we build typescript projects. Make sure to exclude them in case the project directory already has files with such name. This is the scenario with NativeScript TypeScript projects, which have android17.d.ts files in the templates and we are adding them again on build.

Fixes http://teampulse.telerik.com/view#item/290964


teobugslayer:
:+1:

PR #314, opened 2015-04-30
send debug break to engine

Plamen5kov:


blagoev:
This fixes the 3s wait time when no debugger is going to attach on this run. 
It creates a new file "debugbrake" which is deleted by the engine if found on the files system.
If found the engine will wait for the debugger to connect for 3s otherwise engine will enable js debugging only.

mobile/android/android-device.ts
```diff
@@ -259,6 +259,9 @@ export class AndroidDevice implements Mobile.IAndroidDevice {
         var clearDebugEnvironmentCommand = this.composeCommand('shell rm "%s"', envDebugOutFullpath);
         this.$childProcess.exec(clearDebugEnvironmentCommand).wait();
 
+		var setDebugBreakEnvironmentCommand = this.composeCommand('shell echo "" > "%s"', "debugbreak");
```
teobugslayer: Can you use ``` touch %s``` instead?

Also, it looks like there are mixed tabs / spaces, please fix them.

blagoev: can you justify touch %s over echo. Will fix mixed spaces/tabs

teobugslayer: ```'shell echo "" > "%s"', "debugbreak"```
vs
```"shell touch debugbreak"```

Looks much more readable to me. Furthermore, man touch explicitly says: "If any file does not exist, it is created ..." while using echo is a hack.

My justification is that I suggest using the right tool for the purpose, which as a side benefit produces shorter and easier to read and support code. 

Of course, your code may depend on the created file containing an empty line, in which case, echo is the correct approach and you can disregard my comment.

blagoev:
Hi Teo,
Thanks for the clarification still we choose to merge it like so, since its more consistent with the rest of the code.
Are you going to merge this or you are waiting for us to do that.

teobugslayer:
:+1: please merge

PR #315, opened 2015-04-30
Totev/onprem config

teobugslayer:


teobugslayer:
Wrong branch, will open another PR against the correct one

PR #316, opened 2015-04-30
Introduce ON_PREM configuration setting

teobugslayer:
Introduce ON_PREM configuration setting
Introduce ICommand.isEnabled property with semantics:
 - when explicitly false, do not execute command
 - when true or undefined, execute it

See http://teampulse.telerik.com/view#item/288143

### DIFF
--- a/definitions/commands.d.ts
+++ b/definitions/commands.d.ts
@@ -2,6 +2,8 @@ interface ICommand extends ICommandOptions {
 	execute(args: string[]): IFuture<void>;
 	allowedParameters: ICommandParameter[];
 
+	isDisabled?: boolean;
+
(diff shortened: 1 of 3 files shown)


rosen-vladimirov:
Maybe we should change the variable to be isDisabled. This way the code will be simplified, as the check will be 
```TypeScript
// Verify command is enabled
if (command.isDisabled) {
...
```

Also isEnabled sounds that if you do not set it to true, all other commands will have it as false and will not be executed. This will simplify the commands when they have to disable the command:
Instead of:
```TypeScript
get isEnabled() {
	return !this.$config.ON_PREM;
}
```
they will just have:
```TypeScript
isEnabled = this.$config.ON_PREM;
```

rosen-vladimirov:
:+1:

PR #317, opened 2015-05-04
Switch the debugging on emulator for iOS to use the iOSEmulatorService

jasssonpet:
This is the respective missing commit in `lib/common` of https://github.com/NativeScript/nativescript-cli/pull/423

### DIFF
--- a/mobile/ios/ios-emulator-services.ts
+++ b/mobile/ios/ios-emulator-services.ts
@@ -85,6 +85,10 @@ class IosEmulatorServices implements Mobile.IEmulatorPlatformServices {
 			opts = opts.concat("--device", options.device);
 		}
 
+		if (emulatorOptions.args) {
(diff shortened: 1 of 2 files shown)


mobile/ios/ios-emulator-services.ts
```diff
@@ -85,6 +85,10 @@ class IosEmulatorServices implements Mobile.IEmulatorPlatformServices {
 			opts = opts.concat("--device", options.device);
 		}
 
+		if (emulatorOptions.args) {
+			opts.push(util.format("--args=\"%s\"", emulatorOptions));
```
Fatme: Shouldn't be emulatorOptions.args?

rosen-vladimirov:
:+1:

Fatme:
:+1:

PR #318, opened 2015-05-04
Encode passwords with asterisks

Mitko-Kerezov:
So that they are not outputted when --log trace option is provided
Fixes http://teampulse.telerik.com/view#item/289692

### DIFF
--- a/logger.ts
+++ b/logger.ts
@@ -9,6 +9,10 @@ import Future = require("fibers/future");
 
 export class Logger implements ILogger {
 	private log4jsLogger: log4js.ILogger = null;
+	private encodeRequestPaths: string[] = ['/appbuilder/api/itmstransporter/applications?username='];
+	private encodeBody: boolean = false;
+	private passwordRegex = /[Pp]assword=(.*?)(['&,]|$)|\"[Pp]assword\":\"(.*?)\"/;
+	private requestBodyRegex = /^\"(.*?)\"$/;
 	
 	constructor($config: Config.IConfig) {
 		var appenders: log4js.IAppender[] = [];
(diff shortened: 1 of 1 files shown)


logger.ts
```diff
@@ -109,5 +113,39 @@ export class Logger implements ILogger {
 
 		return printMsgFuture;
 	}
+
+	private getPasswordEncodedArguments(args: string[]): string[] {
+		return _.map(args, a => {
+			var passwordRegex = /[Pp]assword=(.*?)(['&,]|$)|\"[Pp]assword\":\"(.*?)\"/,
```
teobugslayer: The regexp creation should happen only once. Consider moving it outside the map lambda.

logger.ts
```diff
@@ -109,5 +115,40 @@ export class Logger implements ILogger {
...
+					this.encodeBody = a.indexOf(path) > -1;
+					break;
+				}
+			}
+
+			return a;
+		});
+	}
+
+	private getEncodedPassword(password: string, originalString: string) {
```
teobugslayer: The name of this method does not reflect its function. It does not get the password, it replaces it in the string. Perhaps something similar to ```hidePassword```. Note, the password is not encoded, it is removed.

logger.ts
```diff
@@ -109,5 +115,40 @@ export class Logger implements ILogger {
 
 		return printMsgFuture;
 	}
+
+	private getPasswordEncodedArguments(args: string[]): string[] {
+		return _.map(args, a => {
```
rosen-vladimirov: can we change `a` to `arg` or `argument`

logger.ts
```diff
@@ -109,5 +115,40 @@ export class Logger implements ILogger {
...
+			}
+
+			if (this.encodeBody) {
+				var bodyMatch = this.requestBodyRegex.exec(a);
+				if (bodyMatch) {
+					return this.getEncodedPassword(bodyMatch[1], a);
+				}
+			}
+
+			for (var i = 0; i < this.encodeRequestPaths.length; ++i) {
```
rosen-vladimirov: why not using _.each

teobugslayer:
:+1:

PR #323, opened 2015-05-06
Cherrypick more commits for ns 1.0.1 release

Fatme:
This prevents "swallowing" of arguments, due to Array.splice()

See http://teampulse.telerik.com/view#item/290977


teobugslayer:
:+1:

PR #327, opened 2015-05-08
Fix hierarchical commands' case-sensitivity

Mitko-Kerezov:
In addition declare ErrorCodes enum as const
to conform with TypeScript 1.5
Fixes http://teampulse.telerik.com/view#item/291385


teobugslayer:
:+1:

PR #329, opened 2015-05-14
Check if project file is correct when looking for project dir

Fatme:


project-helper.ts
```diff
@@ -55,5 +56,15 @@ export class ProjectHelper implements IProjectHelper {
 		let sanitizedName = _.filter(appName.split(""), (c) => /[a-zA-Z0-9]/.test(c)).join("");
 		return sanitizedName;
 	}
+	
+	private isProjectFileCorrect(projectFilePath: string): boolean {
+		if(this.$staticConfig.CLIENT_NAME_KEY_IN_PROJECT_FILE) {
+			var fileContent = this.$fs.readJson(projectFilePath).wait();
```
teobugslayer: This code could throw, should we catch it here and return false, or let uplevel  code handle it?

project-helper.ts
```diff
@@ -55,5 +56,15 @@ export class ProjectHelper implements IProjectHelper {
...
+			var clientSpecificData = fileContent[this.$staticConfig.CLIENT_NAME_KEY_IN_PROJECT_FILE];
+			return clientSpecificData !== null && clientSpecificData !== undefined;
```
teobugslayer: consider writing it shortly like

```JavaScript
return !!clientSpecificData
```

project-helper.ts
Fatme: Actually when we use this function we are sure that the file exists - https://github.com/telerik/mobile-cli-lib/pull/329/files#diff-2fc55d29fa5bf8690e0a4ae30a71effdR25

teobugslayer: It exists, but the content may not be a JSON.

teobugslayer:
:+1:

PR #333, opened 2015-05-15
Detect mono version

teobugslayer:
At install, warn if there is problem with mono installation

Implements http://teampulse.telerik.com/view#item/291901

### DIFF
--- a/sysinfo.ts
+++ b/sysinfo.ts
@@ -12,6 +12,7 @@ export class SysInfo implements ISysInfo {
 				private $iTunesValidator: Mobile.IiTunesValidator,
 				private $logger: ILogger) { }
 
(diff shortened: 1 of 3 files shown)


sysinfo.ts
```diff
@@ -12,6 +12,7 @@ export class SysInfo implements ISysInfo {
 				private $iTunesValidator: Mobile.IiTunesValidator,
 				private $logger: ILogger) { }
 
+	private static monoVerRegExp = /version (\d+[.]\d+[.]\d+) /gm;
```
rosen-vladimirov: is the last space of the regex required?

rosen-vladimirov:
:+1:  (one minor comment)

sysinfo.ts
teobugslayer: absolutely

PR #339, opened 2015-05-19
Add a constant containing all provision types

Mitko-Kerezov:


teobugslayer:
It is a good idea to state why this change is needed (I guess https://github.com/Icenium/icenium-cli/pull/1044 needs it). Otherwise, :+1:

mobile/constants.ts
```diff
@@ -5,6 +5,20 @@ export class ProvisionType {
...
+		for(let key in keys) {
+			if (keys[key] === 'AllProvisions') {
+				continue;
+			}
+
+			result.push(keys[key]);
+		}
+
+		return result;
+	}
```
teobugslayer: Doesn't this function boils down to

```
_.filter(ProvisionTypes, _.isString);
```
?

Mitko-Kerezov:
Closing this PR as there is no sensible way of getting the **values** of the properties of this class instead of its **keys**.
Instead wherever this is needed `_.values(ProvisionTypes)` may be used

PR #342, opened 2015-05-22
Work correctly with option's aliases

Fatme:


options.ts
```diff
@@ -26,7 +26,16 @@ export class OptionsBase {
 
 		yargs.options(this.options);
 		this.argv = yargs.argv;
-
+		
+		// Check if some of the given options is alias and add its original name to argv
+		_.each(_.keys(this.argv), item => {
+			_.each(this.optionNames, optionName => {
```
rosen-vladimirov: I think you can use ```_.each(this.optionNames, (optionName, optionValue) => { ```
which will simplify the if below to:
```TypeScript
if(optionValue.alias === item)
```

rosen-vladimirov:
:+1:  Just one idea for improvement, but nothing critical :)

options.ts
Fatme: Actually the `optionValue` is the first parameter :smile:

rosen-vladimirov: argh, ignore my comment 
bear_polar_bear_facepalm

PR #343, opened 2015-05-22
Fix autocompletion on Linux

rosen-vladimirov:
Linux has case-sensitive file system and autocompletion is unable to find correct appbuilder.js/nativescript.js file as it is searching for AppBuilder.js/NativeScript.js


Fatme:
:+1:

PR #344, opened 2015-05-25
Send Darwin notifications in iOS Emulator

fealebenpae:


mobile/ios/ios-emulator-services.ts
```diff
@@ -93,3 +106,5 @@ class IosEmulatorServices implements Mobile.IEmulatorPlatformServices {
 	}
 }
 $injector.register("iOSEmulatorServices", IosEmulatorServices);
+
+export = IosEmulatorServices;
```
rosen-vladimirov: you can set export before the class declaration:
```TypeScript
export class IosEmulatorService
```

mobile/ios/ios-emulator-services.ts
```diff
@@ -39,6 +39,19 @@ class IosEmulatorServices implements Mobile.IEmulatorPlatformServices {
...
+		let iosSimPath = require.resolve("ios-sim-portable");
+		let nodeCommandName = process.argv[0];
+
+		let opts = [ "notify-post", notification ];
+
+		if (this.$options.device) {
+			opts.push("--device", this.$options.device);
+		}
+
+		return this.$childProcess.exec(`${nodeCommandName} ${iosSimPath} ${opts.join(' ')}`);
```
rosen-vladimirov: Will this code work in case this.$options.device contains whitespace (is it possible in fact to have whitespace for this option :smirk_cat: )

rosen-vladimirov: We've discussed the code with @Fatme and it seems the option cannot have whitespace, so ignore my comment.

rosen-vladimirov:
:+1:  (just one minor suggestion for "export")

mobile/ios/ios-emulator-services.ts
fealebenpae: Wouldn't this be equivalent to `module.exports.IosEmulatorService = IosEmulatorService`, whereas I need it to be `module.exports = IosEmulatorService`?

rosen-vladimirov: Yes, this will be equivalent of ```module.exports.IosEmulatorService = IosEmulatorService```.
I've checked your other PR in order to see why you need the export to be this way and it look like you just need the type. My suggestion is to use 
```TypeScript
export class IosEmulatorService implements Mobile.IiOSEmulatorPlatformServices
```
where IiOSEmulatorPlatformService is :
```TypeScript
interface IiOSEmulatorPlatformService extends IEmulatorPlatformServices {
    postDarwinNotification(notification: string): IFuture<void>;
}
```
And in https://github.com/NativeScript/nativescript-cli/pull/479/files#diff-3ea8e9be6fce663fc286a4462786a297R48 you can use 
```TypeScript
private $iOSEmulatorServices: Mobile.IiOSEmulatorPlatformService
```
Maybe I'm missing something why you need the export to be ```module.exports = IosEmulatorService```

fealebenpae:
@rosen-vladimirov I've extracted a new interface like you suggested.

rosen-vladimirov:
awesome :+1:  :+1: :+1:

PR #345, opened 2015-05-26
If none of these environmental variables exist; this will crash.  

Nathanaela:
Create a workable fallback.

### DIFF
--- a/services/cancellation.ts
+++ b/services/cancellation.ts
@@ -58,7 +58,12 @@ class CancellationService implements ICancellationService {
 	}
 
 	private static get killSwitchDir(): string {
-		return path.join(os.tmpdir(), process.env.SUDO_USER || process.env.USER || process.env.USERNAME, "KillSwitches");
+        var tmp = process.env.SUDO_USER || process.env.USER || process.env.USERNAME;
+        if (tmp) {
+            return path.join(os.tmpdir(), tmp, "KillSwitches");
+        } else {
+            return path.join(os.tmpdir(), "KillSwitches");
+        }
 	}
 
 	private static makeKillSwitchFileName(name: string): string {
--- a/services/cancellation.ts
+++ b/services/cancellation.ts
@@ -58,12 +58,7 @@ class CancellationService implements ICancellationService {
 	}
 
 	private static get killSwitchDir(): string {
-        var tmp = process.env.SUDO_USER || process.env.USER || process.env.USERNAME;
-        if (tmp) {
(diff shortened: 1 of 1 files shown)


services/cancellation.ts
```diff
@@ -58,7 +58,12 @@ class CancellationService implements ICancellationService {
 	}
 
 	private static get killSwitchDir(): string {
-		return path.join(os.tmpdir(), process.env.SUDO_USER || process.env.USER || process.env.USERNAME, "KillSwitches");
+        var tmp = process.env.SUDO_USER || process.env.USER || process.env.USERNAME;
+        if (tmp) {
+            return path.join(os.tmpdir(), tmp, "KillSwitches");
+        } else {
+            return path.join(os.tmpdir(), "KillSwitches");
+        }
```
teobugslayer: Isn't this equivalent to
```JavaScript
var tmp = process.env.SUDO_USER || process.env.USER || process.env.USERNAME || "";
return path.join(os.tmpdir(), tmp, "KillSwitches");
```

teobugslayer:
Hello @Nathanaela,
The issue with this code is that if you have machine with two simultaneously logged in users, it would either break due to missing write permissions, or one user would stop the running CLI process of the other user. What could be done is such case?

services/cancellation.ts
Nathanaela: Interesting.   I just tested path.join and your understanding is correct.  I had expected path join to do two slashes in the middle of the path...   I'll repush a commit.

Nathanaela:
Are you sure it isn't running.  I just installed NativeScript yesterday and this code file is present on my box...

teobugslayer:
@Nathanaela,

the code is instantiated only on Windows - see https://github.com/telerik/mobile-cli-lib/commit/d1230fadf8f3dbabde8593662100cf3912e07e95#diff-dc9cf03aafbce6404bc646fee9d733c0R82

You are correct that this code can potentially fail on Windows, though I believe it is highly unlike for USERNAME to miss on Windows.

Maybe the correct fix is:

```
private static get killSwitchDir(): string {
		return path.join(os.tmpdir(), process.env.USERNAME || "", "KillSwitches");
	}
```

Nathanaela:
It is unlikely that it would now fail on windows; as username is set by the os and someone would have to actively remove it from the environment.    But it might be worth doing the patch so that in the future if it is changed to be active on linux/osx it won't fail again.

teobugslayer:
I agree. Thank you for your effort!

PR #346, opened 2015-05-26
Stop using Cancellation on unix OSes

teobugslayer:


teobugslayer:
:+1:  from @rosen-vladimirov

PR #349, opened 2015-05-28
Handle bad file descriptors in iOSDevice.connectToPort

fealebenpae:
Whenever `USBMuxConnectByPort` fails to connect to a port on a device, either because the device is no longer there, or the connection is refused, it returns a negative value file descriptor. This upsets Node's `net.Socket` constructor that accepts a file descriptor and causes an assertion failure.

This patch handles bad file descriptors by creating a pristine `net.Socket` and emitting the `error` event on it.


fealebenpae:
ping @Fatme

teobugslayer:
:+1:

teobugslayer:
:+1: from @fatme as well

PR #361, opened 2015-06-11
Implement doctor command in AppBuilder CLI

teobugslayer:
See https://github.com/NativeScript/nativescript-cli/issues/480 and http://teampulse.telerik.com/view#item/294290


rosen-vladimirov:
:+1:

PR #362, opened 2015-06-12
Add Object type to Options

rosen-vladimirov:
Yargs creates objects for options passed with dots, for example --var.debug.a 2 will create the object
```JavaScript
{
    "debug": 2
}
```
Add the Object type as available for our options and replace failWithoutHelp with fail as the message says "Check the help below", but there's no help when `failWithoutHelp` is used.

Part of http://teampulse.telerik.com/view#item/292387

### DIFF
--- a/options.ts
+++ b/options.ts
@@ -9,7 +9,8 @@ export enum OptionType {
 	String,
 	Boolean,
 	Number,
-	Array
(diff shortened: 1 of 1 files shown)


teobugslayer:
:+1:

PR #363, opened 2015-06-15
Add general track method

rosen-vladimirov:
When feature tracking is enabled, we need an easy way to track new features that will go in separate section on the feature page. Add such track method. When you want to track new information, just pass the feature name and its value to track method.


ligaz:
:+1:

PR #364, opened 2015-06-17
Fix analytics feature tracking

rosen-vladimirov:
Feature tracking is checking incorrect option - client.  As this option has different meaning in our CLIs, add new option - analyticsClient and use it instead of client when tracking features.


ErjanGavalji:
:+1:

Fatme:
:+1:

PR #365, opened 2015-06-18
Chain options before getting argv

Mitko-Kerezov:
This is the [right way](https://www.npmjs.com/package/yargs#options-key-opt) to use the `.options()` function of yargs.


rosen-vladimirov:
:+1:  after green build in icenium-cli
and make sure to create PR in nativescript-cli

Fatme:
:+1:

PR #366, opened 2015-06-19
Fix markdown links in the console

rosen-vladimirov:
Fix the regex for markdown links, so `[\*\*](http://....)` can be shown correctly as \*\*


teobugslayer:
Tried quite hard to break your regex. The only difference against Github's one I could find is that yous does not accept ":" in the link text part. I can live with that, :+1: and awesome work!

PR #367, opened 2015-06-22
Add dashedOptions for each command

rosen-vladimirov:
Each command can use specific dashed options (`-- avd <name>` for example is valid for one command, but not for other). Add dashedOptions parameter to ICommand interface and make sure to use it for validation when it exists. In case there's no dashedOptions paramter, use the common options.

Use correct types for yargs - "string", "boolean", etc. instead of 0, 1, 2 which were the result of our enum. Replace the enum with class and remove code, which we do not need anymore as yargs is doing its work as expected.
After yargs is updated to latest version, some fixes have to be applied - now when `yargs.options(opts).argv` is passed, it returns all members of `opts`, but the ones that are not part of process.argv have value undefined.

### DIFF
--- a/declarations.d.ts
+++ b/declarations.d.ts
@@ -328,7 +328,7 @@ interface Error {
 
 interface ICommonOptions {
 	argv: IYargArgv;
-	validateOptions(): void;
+	validateOptions(commandSpecificDashedOptions?: IDictionary<IDashedOption>): void;
 	options: IDictionary<any>;
 	shorthands: string[];
 
@@ -362,6 +362,65 @@ interface ICommonOptions {
 interface IYargArgv extends IDictionary<any> {
 	_: string[];
 	$0: string;
+}
+
+/**
+ * Describes dashed option (starting with --) passed on the command line.
+ * @interface
+ */
+interface IDashedOption {
+	/**
+	 * Type of the option. It can be string, boolean, Array, etc.
+	 */
+	type: string;
+	/**
+	 * Shorthand option passed on the command line with `-` sign, for example `-v`
+	 */
+	alias?: any;
+	/**
+	 * Defines if the options is mandatory or the number of mandatory arguments.
+	 */
+	demand?: any;
+	/**
+	 * @see demand
+	 */
+	required?: any;
+	/**
+	 * @see demand
+	 */
+	require?: any;
+	/**
+	 * Sets default value of the -- option if it is NOT passed on the command line.
+	 */
+	default?: any;
+	/**
+	 * Interpret the value as boolean, even if value is passed for it.
+	 */
+	boolean?: any;
+	/**
+	 * Interpret the value as string, especially useful when you have to preserve numbers leading zeroes.
+	 */
+	string?: any;
+	/**
+	 * Returns the count of the dashed options passed on the command line.
+	 */
+	count?: any;
+	/**
+	 * Describes the usage of option.
+	 */
+	describe?: any;
+	/**
+	 * No information about this option. Keep it here for backwards compatibility, but use describe instead.
+	 */
+	description?: any;
+	/**
+	 * @see describe
+	 */
+	desc?: any;
+	/**
+	 * Specifies either a single option key (string), or an array of options that must be followed by option values.
+	 */
+	requiresArg?: any;
 }
 
 /**
--- a/options.ts
+++ b/options.ts
@@ -5,52 +5,37 @@ import util = require("util");
 import helpers = require("./helpers");
 import yargs = require("yargs");
 
-export enum OptionType {
-	String,
-	Boolean,
-	Number,
-	Array,
-	Object
+export class OptionType {
+	public static String = "string";
+	public static Boolean = "boolean";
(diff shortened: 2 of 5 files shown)


declarations.d.ts
```diff
@@ -365,6 +365,69 @@ interface IYargArgv extends IDictionary<any> {
...
 
 /**
+ * Describes dashed option (starting with --) passed on the command line.
+ * @interface
+ */
+interface IDashedOption {
+	/**
+	 * Type of the option. It can be string, boolean, Array, etc.
+	 */
+	type?: any;
```
teobugslayer: is the type really **any** or it can be a **string**?

declarations.d.ts
```diff
@@ -365,6 +365,69 @@ interface IYargArgv extends IDictionary<any> {
...
+	/**
+	 * Interpret the value as boolean, even if value is passed for it.
+	 */
+	boolean?: any;
+	/**
+	 * Interpret the value as string, especially useful when you have to preserve numbers leading zeroes.
+	 */
+	string?: any;
+	/**
+	 * Returns the count of option passed on the command line.
```
teobugslayer: Perhaps option -> options?

options.ts
```diff
@@ -138,9 +131,14 @@ export class OptionsBase {
 		return _.contains(this.optionNames, secondaryOptionName) ? secondaryOptionName : optionName;
 	}
 	
-	private getOptionType(optionName: string) {
-		var option = this.options[optionName] || this.tryGetOptionByAliasName(optionName);
-		 return option ? option.type : ""
+	private getOptionType(optionName: string): any {
```
teobugslayer: Doesn't this method always return a string?

rosen-vladimirov: nope, option.type is number in cases when we use our enum

declarations.d.ts
rosen-vladimirov: This interface describes one option, that's why I've used option (without 's').

declarations.d.ts
rosen-vladimirov: sometimes it is number

teobugslayer: can we use "string|number" instead?

rosen-vladimirov: I prefer to keep it any, as we have string, number, array and object

options.ts
```diff
@@ -4,13 +4,14 @@ import path = require("path");
...
 import helpers = require("./helpers");
 import yargs = require("yargs");
-
-export enum OptionType {
-	String,
-	Boolean,
-	Number,
-	Array,
-	Object
+let assert = require("assert");
```
teobugslayer: can we import it?

options.ts
```diff
@@ -84,11 +55,20 @@ export class OptionsBase {
 			"stop": { type: OptionType.Boolean },
 			"ddi": { type: OptionType.String }, // the path to developer  disk image
 			"justlaunch": { type: OptionType.Boolean },
-			"file": { type: OptionType.String },
-			"analyticsClient": {type: OptionType.String}
+			"file": { type: OptionType.String }
 		}
 	}
 
+	private globalOptions: IDictionary<IDashedOption> = {
```
teobugslayer: can we make this static?

declarations.d.ts
rosen-vladimirov: I'm stupid - it is always string, I'll fix it and make it mandatory

options.ts
```diff
@@ -98,36 +78,42 @@ export class OptionsBase {
...
 				if (_.isArray(optionValue) && optionType !== OptionType.Array) {
 					this.$errors.fail("You have set the %s option multiple times. Check the correct command syntax below and try again.", originalOptionName);
-				} else if (this.doesOptionRequireValue(optionType, parsedOptionType)) {
-					this.$errors.failWithoutHelp("The option '%s' requires a value.", originalOptionName);
 				} else if (optionType === OptionType.String && helpers.isNullOrWhitespace(optionValue)) {
 					this.$errors.failWithoutHelp("The option '%s' requires non-empty value.", originalOptionName);
-				} else if (optionType === OptionType.Boolean && parsedOptionType !== 'boolean') {
-					this.$errors.failWithoutHelp("The option '%s' does not accept values.", originalOptionName);
+				} else if(optionType === OptionType.Array && optionValue.length === 0) {
+					this.$errors.failWithoutHelp("The option '%s' requires non-empty array.", originalOptionName);
```
teobugslayer: What is a "non-empty array" from user's point of view? Can we say "at least two arguments" or something similar?

rosen-vladimirov: one argument is enough, the code here is for situations like:
`$ appbuilder appmanager upload android --groups`
as groups is defined as array, yargs sets its value to [] . But we need at least one value

teobugslayer: Then perhaps "the option ${xxx} requires additional parameters"?

teobugslayer: ping @ikoevska for the message

ikoevska: The option '%s' requires one or more values.

teobugslayer:
:+1:

PR #369, opened 2015-06-28
Error-reporting command

rosen-vladimirov:
Add error-reporting command that should be used in a similar manner as feature-usage-tracking command. The difference is that during installation, the exceptions tracking will be enabled without prompt (keep current behavior that we are tracking exceptions). It can be disabled later by using `$ appbuilder error-reporting disable`.
Refactor analytics-service in order to "hide" some methods from the interface and to simplify adding of new analytics commands and variables. 

**Rename feature-usage-tracking to usage-reporting**

### DIFF
--- a/commands/analytics.ts
+++ b/commands/analytics.ts
@@ -1,52 +1,75 @@
 ///<reference path="../../.d.ts"/>
 "use strict";
 
-import util = require("util");
-
-export class AnalyticsCommand implements ICommand {
+class AnalyticsCommand implements ICommand {
 	constructor(private $analyticsService: IAnalyticsService,
 		private $logger: ILogger,
-		private $errors: IErrors) { }
+		private $errors: IErrors,
+		private $options: IOptions,
+		private $staticConfig: IStaticConfig,
+		private settingName: string,
+		private humanReadableSettingName: string) { }
+		
+	public disableAnalytics = true;
+	public allowedParameters = [new AnalyticsCommandParameter(this.$errors)];
 
-	execute(args: string[]): IFuture<void> {
+	public execute(args: string[]): IFuture<void> {
 		return(() => {
-			let arg = args[0];
-			switch(arg) {
+			let arg = args[0] || "";
+			switch(arg.toLowerCase()) {
 				case "enable":
-					this.$analyticsService.setAnalyticsStatus(true).wait();
-					this.$logger.info("Feature usage tracking is now enabled.");
+					this.$analyticsService.setStatus(this.settingName, true).wait();
+					this.$logger.info(`${this.humanReadableSettingName} is now enabled.`);
 					break;
 				case "disable":
-					this.$analyticsService.disableAnalytics().wait();
-					this.$logger.info("Feature usage tracking is now disabled.");
+					this.$analyticsService.setStatus(this.settingName, false).wait();
+					this.$logger.info(`${this.humanReadableSettingName} is now disabled.`);
 					break;
 				case "status":
-				case undefined:
-					this.$logger.out(this.$analyticsService.getStatusMessage().wait());
+				case "":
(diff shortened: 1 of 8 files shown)


services/analytics-service.ts
```diff
@@ -21,13 +21,16 @@ export class AnalyticsService implements IAnalyticsService {
 		private $prompter: IPrompter,
 		private $userSettingsService: UserSettings.IUserSettingsService,
 		private $analyticsSettingsService: IAnalyticsSettingsService,
-		private $options: IOptions) { }
+		private $options: IOptions) {
+			this.settingsNames = [this.$staticConfig.TRACK_FEATURE_USAGE_SETTING_NAME, this.$staticConfig.ERROR_REPORT_SETTING_NAME];
+			this.settingsNames.forEach(settingName => this.getStatus(settingName).wait());
```
rosen-vladimirov: This should not be here as it is causing cyclic dependencies error in some cases

commands/analytics.ts
```diff
@@ -6,23 +6,64 @@ import util = require("util");
...
+				case "disable":
+					this.$analyticsService.setStatus(settingName, false).wait();
+					this.$logger.info("Usage reporting is now disabled.");
+					break;
+				case "status":
+				case "":
+					this.$logger.out(this.$analyticsService.getStatusMessage(settingName, this.$options.json, "Usage reporting").wait());
+					break;
+				default:
+					this.$errors.fail(util.format("The value '%s' is not valid. Valid values are 'enable', 'disable' and 'status'.", arg));
```
teobugslayer: Consider using template strings instead.

commands/analytics.ts
```diff
@@ -32,18 +73,19 @@ export class AnalyticsCommand implements ICommand {
...
 	validate(validationValue: string): IFuture<boolean> {
 		return (() => {
-			switch(validationValue) {
+			let val = validationValue || "";
+			switch(val.toLowerCase()) {
 				case "enable":
 				case "disable":
 				case "status":
-				case undefined:
+				case "":
```
teobugslayer: We can use template string on line 91 :)

services/analytics-service.ts
```diff
@@ -186,70 +183,75 @@ export class AnalyticsService implements IAnalyticsService {
...
+			let status = this.getStatus(settingName).wait();
 			let enabled = status === AnalyticsStatus.notConfirmed ? null : status === AnalyticsStatus.disabled ? false : true;
 			return JSON.stringify({ enabled: enabled });
 		}).future<string>()();
 	}
+
+	private isEverythingDisabled(): boolean {
+		let statuses = _(this.analyticsStatuses)
+						.values()
+						.groupBy(p => p)
```
teobugslayer: or perhaps ```.groupBy(_.identity) ```

commands/analytics.ts
```diff
@@ -1,52 +1,88 @@
...
+		private $options: IOptions,
+		private $staticConfig: IStaticConfig) { }
+
+	public disableAnalytics = true;
+
+	execute(args: string[]): IFuture<void> {
+		return(() => {
+			let arg = args[0] || "";
+			let settingName = this.$staticConfig.ERROR_REPORT_SETTING_NAME;
+			switch(arg.toLowerCase()) {
```
Fatme: This code is duplicated with the code in AnalyticsCommand. Consider reusing it.

commands/analytics.ts
rosen-vladimirov: done :)

rosen-vladimirov:
ping @Fatme  - I've changed the commands as you have suggested in order to reuse the code, could you take a look
@teobugslayer I've fixed the code based on your comments

teobugslayer:
:+1:

PR #370, opened 2015-06-29
Exposes read function from properties parser class

Fatme:


properties-parser.ts
```diff
@@ -10,6 +10,19 @@ export class PropertiesParser implements IPropertiesParser {
 	public parse(text: string): any {
 		return propertiesParser.parse(text);
 	}
+	
+	public read(filePath: string): IFuture<IStringDictionary> {
+		let future = new Future<any>();
```
teobugslayer: The return type declaration of the function and the return variable are different - which is the correct one?

teobugslayer:
:+1:

PR #372, opened 2015-06-30
Use package.json's buildVersion property

rosen-vladimirov:
In case the package.json has buildVersion property, concatenate it with the version from staticConfig and show it to user when `$ appbuilder --version` is invoked.
Add PATH_TO_PACKAGE_JSON to staticConfig in order to allow each CLI to specify where is its package.json.

### DIFF
--- a/dispatchers.ts
+++ b/dispatchers.ts
@@ -13,13 +13,13 @@ export class CommandDispatcher implements ICommandDispatcher {
 		private $commandsService: ICommandsService,
 		private $staticConfig: Config.IStaticConfig,
 		private $sysInfo: ISysInfo,
-		private $options: IOptions) { }
+		private $options: IOptions,
+		private $fs: IFileSystem) { }
 
 	public dispatchCommand(): IFuture<void> {
 		return(() => {
 			if (this.$options.version) {
-				this.$logger.out(this.$staticConfig.version);
-				return;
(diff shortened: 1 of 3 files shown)


dispatchers.ts
```diff
@@ -13,12 +13,19 @@ export class CommandDispatcher implements ICommandDispatcher {
...
+		private $options: IOptions,
+		private $fs: IFileSystem) { }
 
 	public dispatchCommand(): IFuture<void> {
 		return(() => {
 			if (this.$options.version) {
-				this.$logger.out(this.$staticConfig.version);
+				let version = this.$staticConfig.version;
+
+				let json = this.$fs.readJson(this.$staticConfig.PATH_TO_PACKAGE_JSON).wait();
```
teobugslayer: You may require package.json, which has been already cached by $staticConfig.version but not a merge stopper.

dispatchers.ts
```diff
@@ -13,12 +13,19 @@ export class CommandDispatcher implements ICommandDispatcher {
...
+				if(json && json.buildVersion) {
+					version = `${version}-${json.buildVersion}`;
+				}
+				this.$logger.out(version);
```
teobugslayer: I would extract this code in a separate function and call it here.

teobugslayer:
:+1:

PR #376, opened 2015-07-07
LiveSync support

Fatme:


definitions/mobile.d.ts
```diff
@@ -6,49 +6,92 @@ declare module Mobile {
...
-		runApplication(applicationId: string): IFuture<void>;
-		uninstallApplication(applicationId: string): IFuture<void>;
-		listFiles(devicePath: string): IFuture<void>;
-		getFile(deviceFilePath: string): IFuture<void>;
-		putFile(localFilePath: string, deviceFilePath: string): IFuture<void>;
+	}
+	
+	interface IAndroidDevice extends IDevice {
+		adb: Mobile.IAndroidDebugBridge;		
+		sendBroadcastToDevice(action: string, extras: IStringDictionary): IFuture<number>;
```
ligaz: You can move the broadcast sending to the `IAndroidDebugBridge` interface.

definitions/mobile.d.ts
```diff
@@ -6,49 +6,92 @@ declare module Mobile {
...
 	}
 
-	interface IAppIdentifier {
+	interface IIOSDevice extends IDevice {
```
ligaz: The second `I` should be lowercased because the name is `iOS`.

definitions/mobile.d.ts
```diff
@@ -6,49 +6,92 @@ declare module Mobile {
...
+		startService(serviceName: string): number;
+		mountImage(): IFuture<void>;
```
ligaz: Shouldn't you pass the path to the image here?

definitions/mobile.d.ts
```diff
@@ -6,49 +6,92 @@ declare module Mobile {
...
+		tryToExecuteFunction<TResult>(func: () => TResult): TResult;
```
ligaz: You can simplify the name here to `tryExecute(...)`

definitions/mobile.d.ts
```diff
@@ -6,49 +6,92 @@ declare module Mobile {
...
-		deviceProjectPath: string;
-		liveSyncFormat: string;
-		encodeLiveSyncHostUri(hostUri: string): string;
-		isLiveSyncSupported(device: any): IFuture<boolean>;
-		getLiveSyncNotSupportedError(device: any): string;
 	}
 
-	interface IAndroidDevice extends IDevice {
-		debug(packageFile: string, packageName: string, debuggerSetup?: any): IFuture<void>;
+	interface IDeviceAppData extends IDeviceAppDataBase {
```
ligaz: Consider removing `Data` suffix from the name here because this is not DTO. It has a method: `isLiveSyncSupported(...)`

definitions/mobile.d.ts
```diff
@@ -6,49 +6,92 @@ declare module Mobile {
...
 	}
 
-	interface IAppIdentifier {
+	interface IIOSDevice extends IDevice {
+		startService(serviceName: string): number;
+		mountImage(): IFuture<void>;
+		tryToExecuteFunction<TResult>(func: () => TResult): TResult;
+	}
+	
+	interface IDeviceAppDataBase {
```
ligaz: Do we really need this interface? Can `IDeviceAppDataBase` and `IDeviceAppData` be merged?

definitions/mobile.d.ts
```diff
@@ -6,49 +6,92 @@ declare module Mobile {
...
 
 	interface ILogcatPrinter {
 		print(line: string): void;
 	}
+	
+	interface IDeviceApplicationManager {
+		getInstalledApplications(): IFuture<string[]>;
+		installApplication(packageFilePath: string): IFuture<void>;
+		uninstallApplication(appIdentifier: string): IFuture<void>; 
+		runApplication(appIdentifier: string): IFuture<void>;
```
ligaz: Consider either renaming this method to `startApplication` or rename the `stopApplication` to something else. The main issue is that when you have such pair of methods they should have common prefixes: `start`/`stop`, `begin`/`end`, etc.

definitions/mobile.d.ts
```diff
@@ -6,49 +6,92 @@ declare module Mobile {
...
+		stopApplication(appIdentifier: string): IFuture<void>;
+		restartApplication(applicationId: string): IFuture<void>;		
+	}
+	
+	interface IDeviceFileSystem {
+		listFiles(devicePath: string): IFuture<void>;
+		getFile(deviceFilePath: string): IFuture<void>;
+		putFile(localFilePath: string, deviceFilePath: string): IFuture<void>;
+		transferCollection(appIdentifier: string, localToDevicePaths: Mobile.ILocalToDevicePathData[]): IFuture<void>;
```
ligaz: There is no need to have `Collection` suffix here. Maybe `Files` will be better.

definitions/mobile.d.ts
```diff
@@ -6,49 +6,92 @@ declare module Mobile {
...
+		transferFile?(localFilePath: string, deviceFilePath: string): IFuture<void>;
+	}
+	
+	interface IAndroidDebugBridge {
+		composeCommand(...args: string[]): IFuture<string>;
```
ligaz: Is there a need for this method to be part of the interface?

definitions/mobile.d.ts
```diff
@@ -69,7 +112,7 @@ declare module Mobile {
 	interface IDevicesServices {
 		hasDevices: boolean;
 		deviceCount: number;
-		execute(action: (device: Mobile.IDevice) => IFuture<void>, canExecute?: (dev: Mobile.IDevice) => boolean, options?: {allowNoDevices?: boolean}): IFuture<void>;
+		execute(action: (device: Mobile.IDevice, deviceApplicationManager: Mobile.IDeviceApplicationManager, deviceFileSystem: Mobile.IDeviceFileSystem) => IFuture<void>, canExecute?: (dev: Mobile.IDevice) => boolean, options?: {allowNoDevices?: boolean}): IFuture<void>;
```
ligaz: Shouldn't the device be sufficient here?

mobile/android/android-application-livesync-service.ts
```diff
@@ -0,0 +1,63 @@
...
+		return "SyncFiles \r";
+	}
+
+	public static RefreshCurrentViewCommand(): string {
+		return "RefreshCurrentView \r";
+	}
+}
+
+export class AndroidApplicationLiveSyncService implements Mobile.IAndroidLiveSyncService {
+	private static COMMANDS_FILE = "telerik.livesync.commands";
```
ligaz: Those can be made `const`s.

mobile/android/android-application-livesync-service.ts
```diff
@@ -0,0 +1,63 @@
...
+	public static SyncFilesCommand(): string {
+		return "SyncFiles \r";
+	}
+
+	public static RefreshCurrentViewCommand(): string {
+		return "RefreshCurrentView \r";
+	}
+}
+
+export class AndroidApplicationLiveSyncService implements Mobile.IAndroidLiveSyncService {
```
ligaz: What is the need to have `Application` in the class name here?

mobile/android/android-application-livesync-service.ts
```diff
@@ -0,0 +1,63 @@
...
+	private static COMMANDS_FILE = "telerik.livesync.commands";	
+	private static LIVESYNC_BROADCAST_NAME = "com.telerik.LiveSync";
+	
+	constructor(protected device: Mobile.IAndroidDevice,
```
ligaz: Should those be `protected`?

mobile/android/android-application-manager.ts
```diff
@@ -0,0 +1,49 @@
...
+		private $staticConfig: Config.IStaticConfig) { }
+	
+	public getInstalledApplications(): IFuture<string[]> {
+		return (() => {
+			if (!this._installedApplications) {
+				let result = this.adb.executeShellCommand(this.identifier, "pm list packages").wait();
+				this._installedApplications = _.map(result.split(os.EOL), (packageString: string) => {
+					let match = packageString.match(/package:(.+)/);
+					return match ? match[1] : null;
+				}).filter(parsedPackage => parsedPackage != null);
```
ligaz: You can use `_.compact()` here.

mobile/android/android-application-manager.ts
```diff
@@ -0,0 +1,49 @@
...
+			return this._installedApplications;
+		}).future<string[]>()();
+	}
+	
+	public installApplication(packageFilePath: string): IFuture<void> {
+		return this.adb.executeCommand(`install -r "${packageFilePath}"`);
+	}
+
+	public uninstallApplication(appIdentifier: string): IFuture<void> {
+		return (() => { }).future<void>()();
```
ligaz: You can use `_.noop()` or just return a `future.fromResult()`.

mobile/ios/ios-application-manager.ts
```diff
@@ -0,0 +1,130 @@
...
+		return (() => {
+			if(this.$hostInfo.isWindows && (this.$staticConfig.enableDeviceRunCommandOnWindows === null || this.$staticConfig.enableDeviceRunCommandOnWindows === undefined)) {
+				this.$errors.fail("$%s device run command is not supported on Windows for iOS devices.", this.$staticConfig.CLIENT_NAME.toLowerCase());
+			}
+			
+			this.validateApplicationId(applicationId);
+			this.device.mountImage().wait();
+			
+			this.runApplicationCore(applicationId).wait();
+			this.$logger.info("Successfully run application %s on device with ID %s", applicationId, this.device.deviceInfo.identifier);
```
ligaz: Consider using template string literal here.

mobile/ios/ios-core.ts
```diff
@@ -1084,33 +1090,71 @@ export class GDBServer implements Mobile.IGDBServer {
...
+			this.send("QEnvironmentHexEncoded:");
+			this.send("QSetDisableASLR:1");
+		}).future<void>()();
+	}
+	
+	public run(argv: string[]): IFuture<void> {
+		return (() => {
+			this.init().wait();
+			
+			let encodedArguments = _.map(argv, (arg, index) => util.format("%d,%d,%s", arg.length * 2, index, new Buffer(arg).toString("hex"))).join(",");
```
ligaz: I see new `Buffer(...).toString("hex")` repeated a couple of times. Consider extracting it into a method.

definitions/mobile.d.ts
Fatme: `mountImage` internally finds the path to developer image - https://github.com/telerik/mobile-cli-lib/pull/376/files#diff-3efdd98dbb45e19bb229436517332d9dL310 or respects the `ddi` option. I think this is a better approach, because we actually don't know where the image is located. All the logic that finds the best image match is encapsulated in this function https://github.com/telerik/mobile-cli-lib/blob/fatme/ns-livesync/mobile/ios/ios-device.ts#L175

definitions/mobile.d.ts
Fatme: I wanted to have separate interface for base class https://github.com/telerik/mobile-cli-lib/pull/376/files#diff-6e983c3aecd632cd680b54d7aa27a9abR8 but actually I can merge the both interfaces... :smile_cat:

commands/device/stop-application.ts
```diff
@@ -0,0 +1,36 @@
...
+			}
+
+			let action = (device: Mobile.IDevice) =>  { return (() => device.applicationManager.stopApplication(args[0]).wait()).future<void>()(); };
+			this.$devicesServices.execute(action).wait();
+		}).future<void>()();
+	}
+	
+	canExecute(args: string[]): IFuture<boolean> {
+		return (() => {
+			return true;
```
rosen-vladimirov: you can use 
```TypeScript
return Future.fromResult(true);
```

commands/device/stop-application.ts
```diff
@@ -0,0 +1,36 @@
...
+		private $options: IOptions) { }
+
+	allowedParameters: ICommandParameter[] = [this.$stringParameter];
+
+	public execute(args: string[]): IFuture<void> {
+		return (() => {
+			this.$devicesServices.initialize({ deviceId: this.$options.device, skipInferPlatform: true }).wait();
+
+			if (this.$devicesServices.deviceCount > 1) {
+				this.$errors.fail("More than one device found. Specify device explicitly with --device option.To discover device ID, use $%s device command.", this.$staticConfig.CLIENT_NAME.toLowerCase());
```
rosen-vladimirov: Shouldn't it be `failWithoutHelp`, the message gives enough information how to fix the problem

commands/device/stop-application.ts
```diff
@@ -0,0 +1,36 @@
...
+
+export class StopApplicationOnDeviceCommand implements ICommand {
+
+	constructor(private $devicesServices: Mobile.IDevicesServices,
+		private $errors: IErrors,
+		private $stringParameter: ICommandParameter,
+		private $staticConfig: Config.IStaticConfig,
+		private $options: IOptions) { }
+
+	allowedParameters: ICommandParameter[] = [this.$stringParameter];
```
rosen-vladimirov: stringParameter will be disregarded, as this command implements `canExecute` method. Are we sure we want this?

commands/device/stop-application.ts
rosen-vladimirov: Also, does this command accepts from 0 to infinity parameters? You can remove canExecute and use command parameters or please add some logic here to check the number of passed args (in case there is some logic that can be added).

mobile/android/android-debug-bridge.ts
```diff
@@ -0,0 +1,54 @@
...
+
+			let result = this.executeShellCommand(broadcastCommand).wait();
+			this.$logger.trace(`Broadcast result ${result} from ${broadcastCommand}`);
+			
+			let match = result.match(/Broadcast completed: result=(\d+)/);
+			if (match) {
+				return +match[1];
+			}
+
+			this.$errors.fail("Unable to broadcast to android device:\n%s", result);
```
rosen-vladimirov: `failWithoutHelp` maybe, but I'm not sure

mobile/android/android-debug-bridge.ts
```diff
@@ -0,0 +1,54 @@
...
+		}).future<number>()();
+	}
+	
+	private composeCommand(args: string[]): IFuture<string> {
```
rosen-vladimirov: return type should be `string`, not `IFuture<string>`

mobile/android/android-debug-bridge.ts
```diff
@@ -0,0 +1,54 @@
...
+		private $childProcess: IChildProcess,
+		private $errors: IErrors,	
+		private $logger: ILogger,	
+		private $staticConfig: Config.IStaticConfig) { }
+	
+
+	public executeCommand(...args: string[]): IFuture<any> {
+		return (() => {
+			let command = this.composeCommand(args).wait();
+			this.$childProcess.exec(command).wait();
```
rosen-vladimirov: return type is `IFuture<any>`, but you are not returning anything, maybe missing return before this childProcess exec or incorrect return type

mobile/android/android-device-file-system.ts
```diff
@@ -0,0 +1,36 @@
...
+		return (() => { }).future<void>()();
+	}
+
+	public putFile(localFilePath: string, deviceFilePath: string): IFuture<void> {
+		return (() => { }).future<void>()();
+	}
+	
+	public transferFiles(appIdentifier: string, localToDevicePaths: Mobile.ILocalToDevicePathData[]): IFuture<void> {
+		return (() => {
+		_.each(localToDevicePaths, (localToDevicePathData) => this.transferFile(localToDevicePathData.getLocalPath(), localToDevicePathData.getDevicePath()).wait());
```
rosen-vladimirov: indent me

mobile/android/android-device-file-system.ts
```diff
@@ -0,0 +1,36 @@
...
+		}).future<void>()();
+	}
+
+	public transferFile(localPath: string, devicePath: string): IFuture<void> {
+		return (() => {
+			this.$logger.trace(`Transfering ${localPath} to ${devicePath}..`);
```
rosen-vladimirov: two `.` at the end

mobile/android/android-livesync-service.ts
```diff
@@ -0,0 +1,63 @@
...
+"use strict";
+
+import helpers = require("../../helpers");
+import path = require("path");
+import temp = require("temp");
+import util = require("util");
+
+class LiveSyncCommands {
+	public static DeployProjectCommand(liveSyncUrl: string): string {
+		return util.format("DeployProject %s \r", liveSyncUrl);
```
rosen-vladimirov: You can use template strings and remove util require. Also what's the idea of `\r`?

mobile/ios/ios-application-manager.ts
```diff
@@ -0,0 +1,130 @@
...
+	// NOTE: This function works only on OSX
+	public stopApplication(applicationId: string): IFuture<void> {
+		let application = this.getApplicationById(applicationId);		
+		let gdbServer = this.createGdbServer();
+		return gdbServer.kill(application.CFBundleExecutable);
+	}
+	
+	public restartApplication(applicationId: string): IFuture<void> {
+		return (() => {
+			this.$childProcess.exec(`tns device stop ${applicationId}`).wait();
```
rosen-vladimirov: tns device stop?
You should use some service, not tns device stop, as the chance to have tns installed globally is not so big.

mobile/android/android-debug-bridge.ts
Fatme: We have `.wait()` inside the function - https://github.com/telerik/mobile-cli-lib/pull/376/files#diff-416bc024035578187882887a05fe7bfaR50

rosen-vladimirov: correct, sorry :)

mobile/ios/ios-application-manager.ts
Fatme: I am not able to use service here, because I need to execute the command in separate process.

teobugslayer: Then use the process.argv to construct the proper command line - this way you will always have a working command line regardless the way tns is invoked.

teobugslayer: Wait, why the common lib invokes tns? This code does not belong here at all.

commands/device/stop-application.ts
teobugslayer: Shouldn't this check be performed in canExecute?

mobile/android/android-application-manager.ts
```diff
@@ -0,0 +1,50 @@
...
+		private identifier: string,
+		private $options: IOptions,
+		private $staticConfig: Config.IStaticConfig) { }
+	
+	public getInstalledApplications(): IFuture<string[]> {
+		return (() => {
+			if (!this._installedApplications) {
+				let result = this.adb.executeShellCommand("pm list packages").wait();
+				this._installedApplications = _.map(result.split(os.EOL), (packageString: string) => {
+					let match = packageString.match(/package:(.+)/);
```
teobugslayer: Consider moving the regexp outside the lambda.

mobile/android/android-livesync-service.ts
```diff
@@ -0,0 +1,63 @@
...
+	
+	public createCommandsFileOnDevice(commandsFileDevicePath: string, commands: string[]): IFuture<void> {
+		return (() => {
+			let hostTmpDir = this.getTempDir();
+			let commandsFileHostPath = path.join(hostTmpDir, AndroidLiveSyncService.COMMANDS_FILE);
+			this.$fs.writeFile(commandsFileHostPath, commands.join("\n")).wait();
+			
+			// copy it to the device
+			this.device.fileSystem.transferFile(commandsFileHostPath, commandsFileDevicePath).wait();
+			this.device.adb.executeShellCommand(`chmod 0777 ${commandsFileDevicePath}`).wait();
```
teobugslayer: Is there a reason for this file to be executable? Can we narrow down the permissions to the essential ones?

mobile/ios/ios-application-manager.ts
```diff
@@ -0,0 +1,130 @@
...
+			} finally {
+			}
+
+			this.$logger.trace("Application %s has been uninstalled successfully.", applicationId);
+		}).future<void>()();
+	}	
+	
+	public startApplication(applicationId: string): IFuture<void> {
+		return (() => {
+			if(this.$hostInfo.isWindows && (this.$staticConfig.enableDeviceRunCommandOnWindows === null || this.$staticConfig.enableDeviceRunCommandOnWindows === undefined)) {
```
teobugslayer: Is there a reason not to use the shorter !this.$staticConfig.enableDeviceRunCommandOnWindows?

mobile/ios/ios-application-manager.ts
```diff
@@ -0,0 +1,130 @@
...
+			
+			this.validateApplicationId(applicationId);
+			this.device.mountImage().wait();
+			
+			this.runApplicationCore(applicationId).wait();
+			this.$logger.info(`Successfully run application ${applicationId} on device with ID ${ this.device.deviceInfo.identifier}.`);
+		}).future<void>()();
+	}
+	
+	// NOTE: This function works only on OSX
```
teobugslayer: Just write an assert inside the function instead / in addition to this comment.

mobile/ios/ios-device-file-system.ts
```diff
@@ -0,0 +1,106 @@
...
+
+			let afcClient = this.resolveAfc();
+
+			let walk = (root:string, indent:number) => {
+				this.$logger.info(util.format("%s %s", Array(indent).join(" "), root));
+				let children:string[] = [];
+				try {
+					children = afcClient.listDir(root);
+				} catch (e) {
+					children = [];
```
teobugslayer: This code is not needed, the aray is already an `[`]

teobugslayer:
This code is highly non-trivial - can we write some unit tests?

Fatme:
It is not trivial to write unit tests for this code because it is device related.

mobile/ios/ios-application-manager.ts
Fatme: Yes, when `enableDeviceRunCommandOnWindows` is false.

commands/device/stop-application.ts
Fatme: I've removed canExecute function.

mobile/android/android-livesync-service.ts
Fatme: Livesync plugin requires executable permissions.

mobile/device-app-data/device-app-data-base.ts
```diff
@@ -0,0 +1,36 @@
...
+			return broadcastResult !== 0;
+		}).future<boolean>()();
+	}
+}
+
+export class CompanionDeviceAppDataBase extends DeviceAppDataBase {
+	public isLiveSyncSupported(device: Mobile.IDevice): IFuture<boolean> {
+		return (() => {
+			let applications = device.applicationManager.getInstalledApplications().wait();
+			return _.contains(applications, this.appIdentifier);
```
rosen-vladimirov: Do we want case-sensitive match here?

mobile/ios/ios-application-manager.ts
```diff
@@ -0,0 +1,129 @@
...
+			installationProxy.closeSocket();
+		}).future<void>()();
+	}
+
+	public uninstallApplication(applicationId: string): IFuture<void> {
+		return (() => {
+			let afc = this.device.startService(iOSProxyServices.MobileServices.INSTALLATION_PROXY);
+			try {
+				let result = this.$mobileDevice.deviceUninstallApplication(afc, this.$coreFoundation.createCFString(applicationId), null, this.uninstallApplicationCallbackPtr);
+				if(result !== 0) {
```
rosen-vladimirov: in case result is always number, you can use
```TypeScript
if(result)
```

mobile/device-app-data/device-app-data-base.ts
Fatme: Yes, we want case-sensitive match here.

commands/device/stop-application.ts
rosen-vladimirov: nice :)

mobile/ios/ios-application-manager.ts
```diff
@@ -0,0 +1,129 @@
...
+					this.$errors.fail("AMDeviceUninstallApplication returned '%d'.", result);
```
rosen-vladimirov: failWithoutHelp is better. The help will not give any benefit in this situation

mobile/ios/ios-application-manager.ts
```diff
@@ -0,0 +1,129 @@
...
+				}
+			} finally {
+			}
```
rosen-vladimirov: What's the idea of the empty finally block?

mobile/ios/ios-application-manager.ts
rosen-vladimirov: in case `enableDeviceRunCommandOnWindows` is false, the code will continue its execution, is this expected? I assume in case `enableDeviceRunCommandOnWindows` is false, the command will not be supported on Windows, am I wrong?

mobile/ios/ios-application-manager.ts
```diff
@@ -0,0 +1,129 @@
...
+	
+	public stopApplication(applicationId: string): IFuture<void> {
+		let application = this.getApplicationById(applicationId);		
+		let gdbServer = this.createGdbServer();
+		return gdbServer.kill(application.CFBundleExecutable);
+	}
+	
+	public restartApplication(applicationId: string): IFuture<void> {
+		return (() => {
+			this.$childProcess.exec(`${process.argv[0]} device stop ${applicationId}`).wait();
```
rosen-vladimirov: `process.argv[0]` should be node, so you'll try executing `node device stop goshko.telerik.com`
Also spawning another process for this is not really expected for me, can't we just execute the logic behind `device stop` command instead of spawning? Also this code will fail in case we have local installation of CLI.

mobile/ios/ios-application-manager.ts
```diff
@@ -0,0 +1,129 @@
...
+			this.runApplicationCore(applicationId).wait();
+		}).future<void>()();
+	}
+	
+	private lookupApplications(): IDictionary<Mobile.IDeviceApplication> {
+		let func = () => {
+			let dictionaryPointer = ref.alloc(CoreTypes.cfDictionaryRef);
+			let result = this.$mobileDevice.deviceLookupApplications(this.devicePointer, 0, dictionaryPointer);
+			if(result !== 0) {
```
rosen-vladimirov: `if(result)` maybe

mobile/ios/ios-application-manager.ts
```diff
@@ -0,0 +1,129 @@
...
+
+		return this.device.tryExecuteFunction<IDictionary<Mobile.IDeviceApplication>>(func);
+	}
+	
+	private validateApplicationId(applicationId: string): Mobile.IDeviceApplication {
+		let applications = this.lookupApplications();
+		let application = applications[applicationId];
+		if(!application) {
+			let sortedKeys = _.sortBy(_.keys(applications));
+			this.$errors.fail("Invalid application id: %s. All available application ids are: %s%s ", applicationId, os.EOL, sortedKeys.join(os.EOL));
```
rosen-vladimirov: `failWithoutHelp`

mobile/ios/ios-core.ts
```diff
@@ -593,6 +595,10 @@ export class MobileDevice implements Mobile.IMobileDevice {
 	public afcDirectoryCreate(afcConnection: NodeBuffer, path: string): number {
 		return this.mobileDeviceLibrary.AFCDirectoryCreate(afcConnection, path);
 	}
+	
+	public afcFileInfoOpen(afcConnection: NodeBuffer, path: string, afcDirectory: NodeBuffer): number {
```
rosen-vladimirov: this method is not doing anything, is there a reason to have it here? The code which calls it, can call `this.mobileDeviceLibrary.AFCFileInfoOpen` directly.

mobile/ios/ios-core.ts
```diff
@@ -1084,33 +1090,71 @@ export class GDBServer implements Mobile.IGDBServer {
...
+			this.send("A" + encodedArguments);
+	
+			this.send("qLaunchSuccess");
+	
+			if(this.$hostInfo.isWindows) {
 				this.send("vCont;c");
 			} else {
-				// Disconnecting the debugger closes the socket and allows the process to quit
-				this.send("D");
+				if (!this.$options.justlaunch) {
```
rosen-vladimirov: not a merge stopper, but can we invert the if, so it will be something like:
```TypeScript
if(this.$options.justlaunch) {
```

mobile/ios/ios-proxy-services.ts
```diff
@@ -145,6 +146,13 @@ export class AfcClient implements Mobile.IAfcClient {
 
 		return entries;
 	}
+	
+	public close(): void {
+		let result = this.$mobileDevice.afcConnectionClose(this.afcConnection);
+		if (result !== 0) {
+			this.$errors.fail(`Unable to close apple file connection: ${result}`);
```
rosen-vladimirov: apple -> Apple maybe
failWithoutHelp maybe

mobile/ios/ios-application-manager.ts
Fatme: `device stop` command must be in separate process because sometimes `ddb` crashes and the cli process crashes too.

services/usb-livesync-service-base.ts
```diff
@@ -0,0 +1,125 @@
...
+interface IProjectFileInfo {
+	fileName: string;
+	onDeviceName: string;
+	shouldIncludeFile: boolean;
+}
+
+export class UsbLiveSyncServiceBase implements IUsbLiveSyncServiceBase {
+	constructor(protected $devicesServices: Mobile.IDevicesServices,
+		protected $mobileHelper: Mobile.IMobileHelper,
+		private $localToDevicePathDataFactory: Mobile.ILocalToDevicePathDataFactory,
```
rosen-vladimirov: can you order the dependencies based on their privacy... just move this private dependency next to other private dependencies if it is possible

mobile/ios/ios-core.ts
rosen-vladimirov: after personal discussion, we have decided to keep the code as is for consistency

mobile/ios/ios-application-manager.ts
teobugslayer: @rosen-vladimirov, process.argv[0] is node, except when it is `nodejs` or `iojs` :)

teobugslayer: This *may be* a problem. Proton does not have a command line executable. Instead, it links to node.dll. Is this code expected to be called by Proton, if so, it needs to be re-written.

mobile/ios/ios-application-manager.ts
Fatme: yes @rosen-vladimirov  :smile_cat:  you're right. I mean that the reason we cannot use `!this.$staticConfig.enableDeviceRunCommandOnWindows` is when enableDeviceRunCommandOnWindows is false.

mobile/ios/ios-application-manager.ts
Fatme: This function will not be used in Proton because it needs ddi image which is unavailable on windows.

rosen-vladimirov:
:+1:

PR #377, opened 2015-07-08
Mobile-cli modifications

Mitko-Kerezov:
* Replace BOM from the header of the file if it exists
* Throw in prompter if console isn't interactive
* Introduce trimSymbol function

### DIFF
--- a/file-system.ts
+++ b/file-system.ts
@@ -226,7 +226,8 @@ export class FileSystem implements IFileSystem {
 		return (() => {
 			let data = this.readText(filename, encoding).wait();
 			if(data) {
-				return JSON.parse(data);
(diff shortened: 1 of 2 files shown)


rosen-vladimirov:
:+1:

mobile/constants.ts
```diff
@@ -7,4 +7,8 @@ export class ProvisionType {
 	static Enterprise = "Enterprise";
 }
 
+export class ImageConstants {
```
Fatme: This should be in `icenium-cli`

Mitko-Kerezov:
Ping @Fatme @rosen-vladimirov @teobugslayer for a quick review

PR #383, opened 2015-07-15
Fail in prompter when console is not interactive

Mitko-Kerezov:
When the console is not interactive there's no point in using the prompter at all.
Part of [#297090](http://teampulse.telerik.com/view#item/297090)'s implementation.

### DIFF
--- a/prompter.ts
+++ b/prompter.ts
@@ -41,13 +41,17 @@ export class Prompter implements IPrompter {
 
 	public get(schema: IPromptSchema[]): IFuture<any> {
 		let future = new Future;
-		prompt.prompt(schema, (result: any) => {
-			if(result) {
-				future.return(result);
-			} else {
-				future.throw(result);
-			}
-		});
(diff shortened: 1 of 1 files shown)


rosen-vladimirov:
:+1:

prompter.ts
```diff
@@ -41,13 +41,17 @@ export class Prompter implements IPrompter {
...
 		let future = new Future;
-		prompt.prompt(schema, (result: any) => {
-			if(result) {
-				future.return(result);
-			} else {
-				future.throw(result);
-			}
-		});
+		if (!helpers.isInteractive() && _.any(schema, s => !s.default)) {
+			future.throw('Console is not interactive');
```
teobugslayer: From user point of view, this message is completely meaningless and not actionable. What about `You are using ${staticConfig.CLIENT_NAME_ALIAS} in non-interactive mode (for example, in a script or a pipe). Please file a issue on ${staticConfig.FeedbackUrl} containing the full details. As a work-around, try running it interactively from the command shell.`

PR #384, opened 2015-07-16
 Initial preparation of common as separate npm package which includes:

rosen-vladimirov:
## Initial preparation of common as separate npm package which includes:
* Add package.json
* Update reference paths to .d.ts
* Add Gruntfile.js

## Provide a way to expose Promisified public API

Use Typescript decorators in order to provide an easy way to expose public API.
Created two decorators:
* `register` - class decorator that can be used instead of `$injector.register`. Usage is `@decorators.register(<module_name>)`
* `promisify` - creates a wrapper around method in order to return promise. The method itself is kept untouched, but the wrapped is added to the publicApi object. Usage is `@decorators.promisify(<module_name>)`

Add requirePublic method in yok in order to create public api. When a module has some methods that must be added to public API, it has to be required with "requirePublic".

Set `getFileSize` of fs module to be promisified in order to allow testing.


rosen-vladimirov:
In order to test the code, you should:
* Get this branch
* Execute `npm install` in the directory
* Execute `grunt pack`
* Get the produced tgz and install it somewhere `npm i mobile-cli-lib-0.0.1-undefined.tgz` - this way it will be placed in node_modules.
* Use some JS, for example:

```JavaScript:
var common = require("mobile-cli-lib");
common.fs.getFileSize("D:\\Work\\t.txt")
    .then(function (a) {
    console.log("in first then");
    console.log(a);
    return a;
})
    .then(function (a) {
    console.log("in second then");
    console.log(a);
})
    .catch(function (err) {
    console.log("in catch");
    console.log(err);
    console.log("stack = ");
    console.log(err.stack);
});
```

definitions/yok.d.ts
```diff
@@ -1,6 +1,8 @@
 interface IInjector extends IDisposable {
 	require(name: string, file: string): void;
 	require(names: string[], file: string): void;
+	requirePublic(name: string, file: string): void;
+	requirePublic(names: string[], file: string): void;
```
ligaz: Can a union type be used here `string | string[]`?

file-system.ts
```diff
@@ -7,7 +6,9 @@ import path = require("path");
 import util = require("util");
 import rimraf = require("rimraf");
 import minimatch = require("minimatch");
+import decorators = require("./decorators");
 
+@decorators.register("fs")
```
ligaz: From a pure API perspective I think it will be better to move the decorators into the injector file itself. It will look like:
```ts
@injector.register("fs")
```

file-system.ts
```diff
@@ -132,7 +133,8 @@ export class FileSystem implements IFileSystem {
 
 		return future;
 	}
-
+	
+	@decorators.promisify("fs")
```
ligaz: Is the parameter `"fs"` really needed here? Can't it be resolved from the class level decorator?

yok.ts
```diff
@@ -29,7 +29,7 @@ let FN_NAME_AND_ARGS = /^function\s*([^\(]*)\(\s*([^\)]*)\)/m;
 let FN_ARG_SPLIT = /,/;
 let FN_ARG = /^\s*(_?)(\S+?)\1\s*$/;
 let STRIP_COMMENTS = /((\/\/.*$)|(\/\*[\s\S]*?\*\/))/mg;
-
+var Promise = require("bluebird");
```
ligaz: Is this needed?

yok.ts
```diff
@@ -137,9 +137,29 @@ export class Yok implements IInjector {
...
+			this.requireOne(name, file);
+			this.resolvePublicApi(name, file);
+		});
+	}
+
+	private resolvePublicApi(name: string, file: string): void {
+		Object.defineProperty(this.publicApi, name, {
+			get: () => {
+				let obj: any = {};
+				let originalModule = this.resolve(name);
```
ligaz: Is something missing here? I cannot see where we populate the `_publicApi` field.

Gruntfile.js
```diff
@@ -0,0 +1,166 @@
...
+
+	grunt.initConfig({
+		jobName: process.env["JOB_NAME"] || defaultEnvironment,
+		buildNumber: process.env["BUILD_NUMBER"] || "non-ci",
+		dateString: now.substr(0, now.indexOf("T")),
+
+		pkg: grunt.file.readJSON("package.json"),
+
+		ts: {
+			options: {
```
ligaz: Do we need those options as there is `tsconfig.json` file as well?

yok.ts
rosen-vladimirov: when this.resolve is called, the injector requires the file. During this process, the decorators are executed and in `promisify` decorator we populate _publicApi.

file-system.ts
rosen-vladimirov: I couldn't find a way to skip the "fs" parameter. The problem is that method decorators are executed before class decorators, so at the moment when the method is promisified, our injector still does not know anything about "fs"

yok.ts
ligaz: Indeed, I saw this after I posted my comment. Having that in mind I would suggest to rename the `promisify` decorator to `public` or `export` to better donate what is is doing under the hood.

definitions/yok.d.ts
```diff
@@ -24,6 +26,8 @@ interface IInjector extends IDisposable {
 	isValidHierarchicalCommand(commandName: string, commandArguments: string[]): boolean;
 	getChildrenCommandsNames(commandName: string): string[];
 	buildHierarchicalCommand(parentCommandName: string, commandLineArguments: string[]): any;
+	publicApi: any;
+	_publicApi: any;
```
Fatme: Do we need this to be part of interface?

tsconfig.json
```diff
@@ -0,0 +1,12 @@
+{
```
Fatme: Is this file needed?

Gruntfile.js
```diff
@@ -0,0 +1,166 @@
...
+				}
+			},
+
+			build_package: {
+				command: "npm pack",
+				options: {
+					execOptions: {
+						env: (function() {
+							var env = shallowCopy(process.env);
+							env["APPBUILDER_SKIP_POSTINSTALL_TASKS"] = "1";
```
Fatme: Is this name correct `APPBUILDER_SKIP_POSTINSTALL_TASKS` ?

Gruntfile.js
```diff
@@ -0,0 +1,166 @@
...
+							return env;
+						})()
+					}
+				}
+			}
+		},
+
+		clean: {
+			src: ["test/**/*.js*", "**/*.js*", "!vendor/*.js", "!hooks/**/*.js", "*.tgz"]
```
Fatme: `hooks` folder is not needed here

common-lib.ts
```diff
@@ -0,0 +1,9 @@
+///<reference path=".d.ts"/>
+
```
teobugslayer: `"use strict";`

decorators.ts
```diff
@@ -0,0 +1,29 @@
+///<reference path=".d.ts"/>
+
```
teobugslayer: `"use strict";`

fiber-bootstrap.ts
```diff
@@ -0,0 +1,13 @@
+import Fiber = require("fibers");
```
teobugslayer: `"use strict";`

tsconfig.json
rosen-vladimirov: in fact no

teobugslayer: I would prefer to introduce tsconfig and remove the code from the gruntfile

fiber-bootstrap.ts
rosen-vladimirov: done

definitions/yok.d.ts
rosen-vladimirov: yes, it is filled in the decorator, so I need it public

Gruntfile.js
rosen-vladimirov: yes, tsconfig.json is not respected from grunt :(

ligaz: Crap, we should look for alternatives :smile:

definitions/yok.d.ts
ligaz: Can we just use a private field in the publicApi itself? For example: `publicApi.__modules__`

rosen-vladimirov: sounds good and I think it will work :) 
I'll change it, thanks for the idea :)

ligaz:
:+1:

PR #385, opened 2015-07-17
Do not call default command when there's no default command

rosen-vladimirov:
In case command does not have default command and arguments are not passed, we are trying to execute default command, which is not expected. Make sure not to return true in this method, when there's no default command.


teobugslayer:
:+1:

PR #386, opened 2015-07-20
Parse command line values as strings

rosen-vladimirov:
Yargs will parse all non-hyphenated values as strings in case we tell we want this. So, let's declare we want this and remove our hack, that's doing it. This will also fix the case when numbers like 8.0 are passed on the console. Yargs is giving them to us as "8", but we need the full value. With this fix, we'll receive "8.0".


teobugslayer:
:+1:

PR #387, opened 2015-07-20
Fix not able to parse JSON files with UTF-8 with BOM encoding

ispasov:


ispasov:
:+1:  from @rosen-vladimirov

PR #388, opened 2015-07-20
Ensure that all directories have permissions

Fatme:
On some devices  adb push removes write permission of others group. This is problem when trying to move files to /data/data because run-as is always executed with different user from the user who pushes the files to /data/local/tmp.

Fixes https://github.com/NativeScript/nativescript-cli/issues/654


rosen-vladimirov:
:+1:

PR #391, opened 2015-07-20
Fix livesync ios when having both iOS and Android devices connected at the same time

Fatme:
Fixes https://github.com/telerik/mobile-cli-lib

Fixes the second part of this issue https://github.com/NativeScript/nativescript-cli/issues/654

### DIFF
--- a/commands/device/stop-application.ts
+++ b/commands/device/stop-application.ts
@@ -13,11 +13,11 @@ export class StopApplicationOnDeviceCommand implements ICommand {
 		private $staticConfig: Config.IStaticConfig,
 		private $options: IOptions) { }
 
-	allowedParameters: ICommandParameter[] = [this.$stringParameter];
+	allowedParameters: ICommandParameter[] = [this.$stringParameter, this.$stringParameter];
 
 	public execute(args: string[]): IFuture<void> {
 		return (() => {
-			this.$devicesServices.initialize({ deviceId: this.$options.device, skipInferPlatform: true }).wait();
+			this.$devicesServices.initialize({ deviceId: this.$options.device, skipInferPlatform: true, platform: args[1] }).wait();
 
 			if (this.$devicesServices.deviceCount > 1) {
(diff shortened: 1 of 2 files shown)


commands/device/stop-application.ts
```diff
@@ -13,11 +13,11 @@ export class StopApplicationOnDeviceCommand implements ICommand {
 		private $staticConfig: Config.IStaticConfig,
 		private $options: IOptions) { }
 
-	allowedParameters: ICommandParameter[] = [this.$stringParameter];
+	allowedParameters: ICommandParameter[] = [this.$stringParameter, this.$stringParameter];
 
 	public execute(args: string[]): IFuture<void> {
 		return (() => {
-			this.$devicesServices.initialize({ deviceId: this.$options.device, skipInferPlatform: true }).wait();
+			this.$devicesServices.initialize({ deviceId: this.$options.device, skipInferPlatform: true, platform: args[1] }).wait();
```
rosen-vladimirov: why is this change required? this way you will not use the first passed argument, for example:
`$ appbuilder device stop android ios`
will omit android platform

rosen-vladimirov: and in case you pass only one platform, it will be skipped, for example:
`$ appbuilder device stop android`

Fatme: This command is not publicly exposed. We use it only as "dev" command because we need to exec stop application logic in separate process.

rosen-vladimirov: Ok, I've got it now, the first argument is the app id, the second one is the platform :)

rosen-vladimirov:
:+1:

commands/device/stop-application.ts
Fatme: This change is required because when livesyncing to ios device and android device is connected at the same time we're initializing and starting device discovery without platform. This triggers android discovery and thus throws an error. Actually initialing device service again is not correct (we've already initialized it at the begging of livesync command) and the change actually passed the specified platform to device service.

PR #395, opened 2015-07-22
Don't fail when more than one ios devices are attached

Fatme:
If you connect two iOS devices and execute `tns livesync` the following exception is thrown: 
`Multiple exceptions were thrown:
Command failed: /bin/sh -c node ~/Work/nativescript-cli/bin/nativescript.js device stop org.nativescript.debuggerApp iOS`


teobugslayer:
:+1:

PR #401, opened 2015-07-27
Fix grunt tasks

rosen-vladimirov:
Fix `grunt clean` task in order to remove only .js and .js.map files from common lib. Un-gitignore bin/common-lib.js file.
Fix rename package to work only when there's specific build number.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -112,7 +112,7 @@ module.exports = function(grunt) {
 		},
 
 		clean: {
-			src: ["test/**/*.js*", "**/*.js*", "!vendor/*.js", "*.tgz"]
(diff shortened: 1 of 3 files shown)


Gruntfile.js
```diff
@@ -112,7 +112,7 @@ module.exports = function(grunt) {
 		},
 
 		clean: {
-			src: ["test/**/*.js*", "**/*.js*", "!vendor/*.js", "*.tgz"]
+			src: ["test/**/*.js*", "**/*.js*", "!**/*.json", "!Gruntfile.js", "!node_modules/**/*", "!bin/common-lib.js",  "!vendor/*.js", "*.tgz"]
```
rosen-vladimirov: maybe it is better to specify the `.js` and `.js.map` directly  instead of using `"**/*.js*"`

teobugslayer: I am fine as is

teobugslayer:
:+1:

teobugslayer:
:+1:

PR #402, opened 2015-07-27
Print connected devices as a table

teobugslayer:
Rewrote device discovery as an EventEmitter

Implements https://github.com/NativeScript/nativescript-cli/issues/702


rosen-vladimirov:
:+1:

PR #403, opened 2015-07-28
Add getFileShasum method

rosen-vladimirov:
The getFileShasum method must return the shasum of a provided file.

### DIFF
--- a/file-system.ts
+++ b/file-system.ts
@@ -7,6 +7,7 @@ import path = require("path");
 import util = require("util");
 import rimraf = require("rimraf");
 import minimatch = require("minimatch");
+import crypto = require('crypto');
 
 export class FileSystem implements IFileSystem {
 	constructor(private $injector: IInjector,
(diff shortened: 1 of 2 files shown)


file-system.ts
```diff
@@ -443,5 +444,28 @@ export class FileSystem implements IFileSystem {
...
 		return foundFiles;
 	}
+
+	public getFileShasum(fileName: string, encoding?: string): IFuture<string> {
+		let future = new Future<string>();
+		encoding = encoding || "sha1";
+		let logger: ILogger = this.$injector.resolve("logger");
+		let shasumData = crypto.createHash(encoding);
+		let fileStream = this.createReadStream(fileName);
+		fileStream.on('data',(data: any) => {
```
teobugslayer: Is the type 'any' or rather 'NodeBuffer'?

file-system.ts
```diff
@@ -443,5 +444,28 @@ export class FileSystem implements IFileSystem {
 		}
 		return foundFiles;
 	}
+
+	public getFileShasum(fileName: string, encoding?: string): IFuture<string> {
+		let future = new Future<string>();
+		encoding = encoding || "sha1";
+		let logger: ILogger = this.$injector.resolve("logger");
```
teobugslayer: I prefer to name variables from injector with leading $, so that searching manually with text editor finds them more easily.

file-system.ts
```diff
@@ -443,5 +444,28 @@ export class FileSystem implements IFileSystem {
...
+			shasumData.update(data);
+		});
+
+		fileStream.on('end', () => {
+			let shasum: string = shasumData.digest('hex');
+			logger.trace(`Shasum of file ${fileName} is ${shasum}`);
+			future.return(shasum);
+		});
+
+		fileStream.on('error', (err: any) => {
```
teobugslayer: Is the type any or Error?

file-system.ts
rosen-vladimirov: https://nodejs.org/api/stream.html#stream_event_data
according to this it could be buffer or string, so I prefer to keep it any

teobugslayer:
:+1:

Fatme:
:+1:

PR #411, opened 2015-08-03
Integrate livesync logic from android runtime

Fatme:
https://github.com/NativeScript/nativescript-cli/issues/742

### DIFF
--- a/declarations.d.ts
+++ b/declarations.d.ts
@@ -264,11 +264,11 @@ interface IHtmlHelpService {
 
 interface IUsbLiveSyncServiceBase {
 	initialize(platform: string): IFuture<string>;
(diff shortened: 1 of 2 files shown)


teobugslayer:
Why do we need this PR? Can we get a link to the issue resolved?

rosen-vladimirov:
I believe this is to fix issues with some Android devices which have "intersting" logic of permissions when starting the app (not compatible with other Android devices). For example this one https://github.com/NativeScript/nativescript-cli/issues/732 will be fixed with these changes

Fatme:
Here is the issue - https://github.com/NativeScript/nativescript-cli/issues/742

teobugslayer:
:+1:

PR #425, opened 2015-08-10
Fix "An error was encountered processing the command (code=146): Exception encountered connecting to CoreSimulatorBridge: Unable to connect to CoreSimulatorBridge"

Fatme:
On some machines sometimes the following error was thrown: "An error was encountered processing the command (code=146): Exception encountered connecting to CoreSimulatorBridge: Unable to connect to CoreSimulatorBridge". This fix also speed up the livesync command on iOS simulator.

### DIFF
--- a/mobile/ios/ios-emulator-services.ts
+++ b/mobile/ios/ios-emulator-services.ts
@@ -202,15 +202,16 @@ class IosEmulatorServices implements Mobile.IiOSSimulatorService {
 			
 			let runningSimulatorId = this.getRunningSimulatorId(appIdentifier).wait();
 			let applicationPath = this.getApplicationPath(appIdentifier, runningSimulatorId).wait();
+			let applicationName = path.basename(applicationPath);
 			syncAction(applicationPath);
 		
 			try {
-				this.$childProcess.exec("killall launchd_sim").wait();
-				this.$childProcess.exec(`xcrun simctl launch ${runningSimulatorId} ${appIdentifier}`).wait();				
+				this.$childProcess.exec(`killall ${applicationName.split(".")[0]}`).wait();					
 			} catch(e) {
 				this.$logger.trace("Unable to kill simulator: " + e);
 			}
 			
+			this.$childProcess.exec(`xcrun simctl launch ${runningSimulatorId} ${appIdentifier}`).wait();				
 		}).future<void>()();
 	}
 }


justcodebuilduser:
:broken_heart:

vchimev:
:+1:

mobile/ios/ios-emulator-services.ts
```diff
@@ -202,15 +202,16 @@ class IosEmulatorServices implements Mobile.IiOSSimulatorService {
...
 		
 			try {
-				this.$childProcess.exec("killall launchd_sim").wait();
-				this.$childProcess.exec(`xcrun simctl launch ${runningSimulatorId} ${appIdentifier}`).wait();				
+				this.$childProcess.exec(`killall ${applicationName.split(".")[0]}`).wait();					
 			} catch(e) {
 				this.$logger.trace("Unable to kill simulator: " + e);
 			}
 			
+			this.$childProcess.exec(`xcrun simctl launch ${runningSimulatorId} ${appIdentifier}`).wait();
```
teobugslayer: This exec needs to be try - catched as well.

Fatme: Why? I think it is better to fail and show the real error instead silently to fail.

mobile/ios/ios-emulator-services.ts
```diff
@@ -202,15 +202,16 @@ class IosEmulatorServices implements Mobile.IiOSSimulatorService {
...
 			let applicationPath = this.getApplicationPath(appIdentifier, runningSimulatorId).wait();
+			let applicationName = _.last(applicationPath.split(path.sep));
 			syncAction(applicationPath);
 		
 			try {
-				this.$childProcess.exec("killall launchd_sim").wait();
-				this.$childProcess.exec(`xcrun simctl launch ${runningSimulatorId} ${appIdentifier}`).wait();				
+				this.$childProcess.exec(`killall ${applicationName.split(".")[0]}`).wait();					
 			} catch(e) {
 				this.$logger.trace("Unable to kill simulator: " + e);
```
rosen-vladimirov: If killall fails, we'll go in the strange and inconsistent state when the button will have multiple contents from different running processes. Shouldn't we just fail directly here?

mobile/ios/ios-emulator-services.ts
```diff
@@ -202,15 +202,16 @@ class IosEmulatorServices implements Mobile.IiOSSimulatorService {
 			
 			let runningSimulatorId = this.getRunningSimulatorId(appIdentifier).wait();
 			let applicationPath = this.getApplicationPath(appIdentifier, runningSimulatorId).wait();
+			let applicationName = _.last(applicationPath.split(path.sep));
```
teobugslayer: Isn't this `path.basename`?

mobile/ios/ios-emulator-services.ts
rosen-vladimirov: or maybe not, we do not have -KILL flag, so the process will not be zombified (I think), so we can safely disregard killall failure :)

mobile/ios/ios-emulator-services.ts
teobugslayer: I see, you are correct.

mobile/ios/ios-emulator-services.ts
Fatme: We shouldn't fail directly because If the process is not running, we'll fail and will not sync successfully.

rosen-vladimirov:
:+1:  after using path.basename if possible :)

justcodebuilduser:
:broken_heart:

PR #428, opened 2015-08-11
Fix android device-related commands

teobugslayer:
Generate proper command line for strating adb
Change public API for adb-related functions
Use ES6 imports
When there's no content in http body , print [no content] instead of [nothing]

Fixes http://teampulse.telerik.com/view#item/299038

### DIFF
--- a/definitions/mobile.d.ts
+++ b/definitions/mobile.d.ts
@@ -85,8 +85,8 @@ declare module Mobile {
 	}
 
 	interface IAndroidDebugBridge {
-		executeCommand(...args: string[]): IFuture<any>;
-		executeShellCommand(...args: string[]): IFuture<any>;
+		executeCommand(args: string[]): IFuture<any>;
+		executeShellCommand(args: string[]): IFuture<any>;
 		sendBroadcastToDevice(action: string, extras?: IStringDictionary): IFuture<number>;
 	}
 
--- a/mobile/android/android-application-manager.ts
+++ b/mobile/android/android-application-manager.ts
@@ -1,20 +1,20 @@
 ///<reference path="../../.d.ts"/>
 "use strict";
-import os = require("os");
+import * as os from "os";
 import Future = require("fibers/future");
 
 export class AndroidApplicationManager implements Mobile.IDeviceApplicationManager {
 	private _installedApplications: string[];
-	
+
 	constructor(private adb: Mobile.IAndroidDebugBridge,
 		private identifier: string,
 		private $options: ICommonOptions,
 		private $staticConfig: Config.IStaticConfig) { }
-	
+
 	public getInstalledApplications(): IFuture<string[]> {
 		return (() => {
 			if (!this._installedApplications) {
-				let result = this.adb.executeShellCommand("pm list packages").wait();
+				let result = this.adb.executeShellCommand(["pm", "list", "packages"]).wait();
 				let regex = /package:(.+)/;
 				this._installedApplications = _.map(result.split(os.EOL), (packageString: string) => {
 					let match = packageString.match(regex);
@@ -25,27 +25,27 @@ export class AndroidApplicationManager implements Mobile.IDeviceApplicationManag
 			return this._installedApplications;
 		}).future<string[]>()();
 	}
-	
+
 	public installApplication(packageFilePath: string): IFuture<void> {
-		return this.adb.executeCommand(`install -r "${packageFilePath}"`);
+		return this.adb.executeCommand(["install", "-r", `${packageFilePath}`]);
 	}
 
 	public uninstallApplication(appIdentifier: string): IFuture<void> {
-		return this.adb.executeShellCommand(`pm uninstall "${appIdentifier}"`);
+		return this.adb.executeShellCommand(["pm", "uninstall", `${appIdentifier}`]);
 	}
-	
+
 	public startApplication(appIdentifier: string): IFuture<void> {
-		return this.adb.executeShellCommand(`am start -a android.intent.action.MAIN -n ${appIdentifier}/${this.$staticConfig.START_PACKAGE_ACTIVITY_NAME} -c android.intent.category.LAUNCHER`);
+		return this.adb.executeShellCommand(["am", "start", "-a", "android.intent.action.MAIN", "-n", `${appIdentifier}/${this.$staticConfig.START_PACKAGE_ACTIVITY_NAME}`, "-c", "android.intent.category.LAUNCHER"]);
 	}
-	
+
 	public stopApplication(appIdentifier: string): IFuture<void> {
-		return this.adb.executeShellCommand(`am force-stop ${appIdentifier}`);
+		return this.adb.executeShellCommand(["am", "force-stop", `${appIdentifier}`]);
 	}
-	
+
 	public restartApplication(appIdentifier: string): IFuture<void> {
 		return (() => {
 			this.stopApplication(appIdentifier).wait();
 			this.startApplication(appIdentifier).wait();
 		}).future<void>()();
-	}	
+	}
 }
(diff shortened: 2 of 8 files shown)


justcodebuilduser:
:heart:

definitions/mobile.d.ts
```diff
@@ -85,8 +85,8 @@ declare module Mobile {
 	}
 
 	interface IAndroidDebugBridge {
-		executeCommand(...args: string[]): IFuture<any>;
-		executeShellCommand(...args: string[]): IFuture<any>;
+		executeCommand(args: string[]): IFuture<any>;
+		executeShellCommand(args: string[]): IFuture<any>;
```
Fatme: Why do you need to change this?

teobugslayer: If i leave it as-was, the compiler complains with `error TS2345: Argument of type 'string[]' is not assignable to parameter of type 'string'.` I am open to suggestions how else to fix it.

mobile/android/android-application-manager.ts
```diff
@@ -25,27 +25,27 @@ export class AndroidApplicationManager implements Mobile.IDeviceApplicationManag
 			return this._installedApplications;
 		}).future<string[]>()();
 	}
-	
+
 	public installApplication(packageFilePath: string): IFuture<void> {
-		return this.adb.executeCommand(`install -r "${packageFilePath}"`);
+		return this.adb.executeCommand(["install", "-r", `${packageFilePath}`]);
```
rosen-vladimirov: if packageFilePath contains spaces, we may still fail, maybe we should verify this and use something like:
`"${packageFilePath}"`

mobile/android/android-debug-bridge.ts
```diff
@@ -11,35 +11,34 @@ interface IComposeCommandResult {
...
+			return this.$childProcess.spawnFromEvent(shellCommand.command, shellCommand.args, "close", undefined, {throwError: false}).wait().stdout;
 		}).future<any>()();
 	}
-	
+
 	public sendBroadcastToDevice(action: string, extras: IStringDictionary = {}): IFuture<number> {
 		return (() => {
-			let broadcastCommand = `am broadcast -a "${action}"`;
-			_.each(extras, (value,key) => broadcastCommand += ` -e "${key}" "${value}"`);
+			let broadcastCommand = ["am", "broadcast", "-a", `${action}`];
```
rosen-vladimirov: is there any chance to have space in "action", as far as I see from the previous code, it was intended to have "", so consider using `"${action}"`

mobile/android/android-device-file-system.ts
```diff
@@ -20,17 +20,17 @@ export class AndroidDeviceFileSystem implements Mobile.IDeviceFileSystem {
...
 		return (() => { }).future<void>()();
 	}
-	
+
 	public transferFiles(appIdentifier: string, localToDevicePaths: Mobile.ILocalToDevicePathData[]): IFuture<void> {
 		return (() => {
 			_(localToDevicePaths)
 				.filter(localToDevicePathData => this.$fs.getFsStats(localToDevicePathData.getLocalPath()).wait().isFile())
-				.each(localToDevicePathData => this.adb.executeCommand(`push "${localToDevicePathData.getLocalPath()}" "${localToDevicePathData.getDevicePath()}"`).wait())
+				.each(localToDevicePathData => this.adb.executeCommand(["push", `${localToDevicePathData.getLocalPath()}`, `${localToDevicePathData.getDevicePath()}`]).wait())
```
rosen-vladimirov: we have to add "" around `localToDevicePathData.getLocalPath()` and `localToDevicePathData.getDevicePath()`

mobile/android/android-device-file-system.ts
```diff
@@ -20,17 +20,17 @@ export class AndroidDeviceFileSystem implements Mobile.IDeviceFileSystem {
...
 				.value();
-			
+
 			_(localToDevicePaths)
 				.filter(localToDevicePathData => this.$fs.getFsStats(localToDevicePathData.getLocalPath()).wait().isDirectory())
-				.each(localToDevicePathData => this.adb.executeShellCommand(`chmod 0777 "${localToDevicePathData.getDevicePath()}"`).wait())
+				.each(localToDevicePathData => this.adb.executeShellCommand(["chmod", "0777", `${localToDevicePathData.getDevicePath()}`]).wait())
```
rosen-vladimirov: same as above - "" around `localToDevicePathData.getDevicePath()`

mobile/android/android-device-file-system.ts
```diff
@@ -40,10 +40,10 @@ export class AndroidDeviceFileSystem implements Mobile.IDeviceFileSystem {
 			this.$logger.trace(`Transfering ${localPath} to ${devicePath}`);
 			let stats = this.$fs.getFsStats(localPath).wait();
 			if(stats.isDirectory()) {
-				this.adb.executeShellCommand(`mkdir "${path.dirname(devicePath)}"`).wait();
+				this.adb.executeShellCommand(["mkdir", `${path.dirname(devicePath)}`]).wait();
```
rosen-vladimirov: still not sure about this, but maybe we need "" as well

definitions/mobile.d.ts
ligaz: Are you both using the same version of the TypeScript compiler?

Fatme: Yes, we use the same version of the TypeScript compiler. The problem actually comes after this change https://github.com/telerik/mobile-cli-lib/commit/aa5e00d3e8c0113837aa8828a746f3d0ea2446bb#diff-416bc024035578187882887a05fe7bfaR22. We've replaced `exec` with `spawnFromEvent`.

mobile/android/android-debug-bridge.ts
```diff
@@ -11,35 +11,34 @@ interface IComposeCommandResult {
...
+			_.each(extras, (value,key) => broadcastCommand.concat(["-e", `${key}`, `${value}`]));
```
rosen-vladimirov: is it possible to have spaces in key or value?

mobile/android/android-device-file-system.ts
```diff
@@ -40,10 +40,10 @@ export class AndroidDeviceFileSystem implements Mobile.IDeviceFileSystem {
...
 			} else {
-				this.adb.executeCommand(`push "${localPath}" "${devicePath}"`).wait(); 
+				this.adb.executeCommand(["push", `${localPath}`, `${devicePath}`]).wait();
```
rosen-vladimirov: and same here

mobile/android/android-livesync-service.ts
```diff
@@ -25,39 +24,39 @@ class LiveSyncCommands {
...
 		return (() => {
 			let hostTmpDir = this.getTempDir();
 			let commandsFileHostPath = path.join(hostTmpDir, AndroidLiveSyncService.COMMANDS_FILE);
 			this.$fs.writeFile(commandsFileHostPath, commands.join("\n")).wait();
-			
+
 			// copy it to the device
 			this.device.fileSystem.transferFile(commandsFileHostPath, commandsFileDevicePath).wait();
-			this.device.adb.executeShellCommand(`chmod 0777 ${commandsFileDevicePath}`).wait();
+			this.device.adb.executeShellCommand(["chmod", "0777", `${commandsFileDevicePath}`]).wait();
```
rosen-vladimirov: maybe "${commandsFileDevicePath}"

mobile/android/android-debug-bridge.ts
teobugslayer: No problem, spawn will add double quotes.

rosen-vladimirov:
Ok, after discussion, we have decided that my comments are incorrect as spawn guarantees that an argument, which has spaces in it, will be sent correctly to the command. So you have my :+1:

teobugslayer:
:+1: from @fatme as well.

PR #431, opened 2015-08-12
Improve packing of common lib

rosen-vladimirov:
Set correct version of ios-sim-portable. Move typescript to devDependencies as it is not required for end users of common lib. Also npmignore some files.


teobugslayer:
:+1:

Mitko-Kerezov:
:+1:

justcodebuilduser:
:heart:

PR #433, opened 2015-08-13
Small code cleanups

teobugslayer:


justcodebuilduser:
:heart:

services/commands-service.ts
```diff
@@ -3,8 +3,7 @@
 
 let jaroWinklerDistance = require("../vendor/jaro-winkler_distance");
 import helpers = require("../helpers");
-import util = require("util");
-import os = require("os");
+import { EOL } from "os";
```
rosen-vladimirov: maybe it is a good idea to have this as global variable, but this is not a merge stopper ;)

rosen-vladimirov:
:+1:

PR #445, opened 2015-08-27
Fix tns doctor command when javac is not added to PATH

Fatme:


justcodebuilduser:
:heart:

Fatme:
:+1: from @teobugslayer

PR #451, opened 2015-09-07
Introduce $logger.printMarkdown

Mitko-Kerezov:
This logic is used when printing help on the console, as well as for command suggestions.

### DIFF
--- a/logger.ts
+++ b/logger.ts
@@ -5,6 +5,9 @@ import * as log4js from "log4js";
 import * as util from "util";
 import * as stream from "stream";
 import Future = require("fibers/future");
+import marked = require("marked");
+let TerminalRenderer = require("marked-terminal");
+let chalk = require("chalk");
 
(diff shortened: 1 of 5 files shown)


justcodebuilduser:
:heart:

logger.ts
```diff
@@ -5,6 +5,9 @@ import * as log4js from "log4js";
 import * as util from "util";
 import * as stream from "stream";
 import Future = require("fibers/future");
+import marked = require("marked");
+let TerminalRenderer = require("marked-terminal");
```
teobugslayer: Can we introduce a .d.ts file and use import instead?

Mitko-Kerezov: Noted, however this module seems to be missing in [DefinatelyTyped](https://github.com/borisyankov/DefinitelyTyped) so writing down a .d.ts file should probably come as a latter PR

teobugslayer:
:+1:

PR #462, opened 2015-09-16
Livesync iOS simulator restart fixes

hdeshev:
1. We are now batching files and restarting just once.
2. No longer failing to restart the app after killing it since it hasn't already died by the time we try to start it.

I also pulled the batching/timeout code to a `SyncBatch` class, so we can share it with the android batch sync operation. Tested and working on an android emulator as well.

### DIFF
--- a/mobile/ios/ios-emulator-services.ts
+++ b/mobile/ios/ios-emulator-services.ts
@@ -208,7 +208,12 @@ class IosEmulatorServices implements Mobile.IiOSSimulatorService {
 				this.$logger.trace("Unable to kill simulator: " + e);
 			}
 
-			this.$childProcess.exec(`xcrun simctl launch ${runningSimulatorId} ${appIdentifier}`).wait();
+			setTimeout(() => {
+				// Killall doesn't always finish immediately, and the subsequent
+				// start fails since the app is already running.
+				// Give it some time to die before we attempt restarting.
+				this.$childProcess.exec(`xcrun simctl launch ${runningSimulatorId} ${appIdentifier}`);
+			}, 500);
 		}).future<void>()();
 	}
 }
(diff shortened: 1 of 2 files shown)


justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

services/usb-livesync-service-base.ts
```diff
@@ -134,38 +183,56 @@ export class UsbLiveSyncServiceBase implements IUsbLiveSyncServiceBase {
...
-	protected preparePlatformForSync(platform: string) {
-        //Overridden in platform-specific services.
-	}
+	private batchSimulatorLiveSync(
+		appIdentifier: string,
+		projectFilesPath: string,
+		filePath: string,
+		notRunningiOSSimulatorAction: () => IFuture<void>,
+		iOSSimulatorRelativeToProjectBasePathAction:(projectFile: string) => string): void {
+			if (this.batch == null || !this.batch.syncPending) {
```
teobugslayer: Consider the shorter `!this.batch`

mobile/ios/ios-emulator-services.ts
```diff
@@ -208,7 +208,12 @@ class IosEmulatorServices implements Mobile.IiOSSimulatorService {
 				this.$logger.trace("Unable to kill simulator: " + e);
 			}
 
-			this.$childProcess.exec(`xcrun simctl launch ${runningSimulatorId} ${appIdentifier}`).wait();
+			setTimeout(() => {
+				// Killall doesn't always finish immediately, and the subsequent
+				// start fails since the app is already running.
+				// Give it some time to die before we attempt restarting.
+				this.$childProcess.exec(`xcrun simctl launch ${runningSimulatorId} ${appIdentifier}`);
```
teobugslayer: You are not .wait()-ing the exec call. Is this intended?

teobugslayer: Can we use $childProcess.spawn instead? .exec runs a sub-shell, which probably is not needed in this case.

hdeshev: > You are not .wait()-ing the exec call. Is this intended?

Yes. It throws an error if you wait for it outside the fiber... thing.

> Can we use $childProcess.spawn instead? .exec runs a sub-shell, which probably is not needed in this case.

I'd have to push all parameters in an array and uglify the code even more. I suggest we stick with the shell.

services/usb-livesync-service-base.ts
hdeshev: Done.

justcodebuilduser:
:heart:

teobugslayer:
:+1:

justcodebuilduser:
:heart:

PR #463, opened 2015-09-17
Fixes the javac version extraction issue

anshulguleria:
Fix NativeScript/nativescript-cli#936. Regex to read javac version was not working for openjdk java
version because it was expecting the output to starts with javac.

output with openjdk is:
```
Picked up JAVA_TOOL_OPTIONS: -javaagent:/usr/share/java/jayatanaag.jar 
javac 1.8.0_45-internal
```


justcodebuilduser:
:heart:

teobugslayer:
Thanks! I'll see that this is included in our next public release.

PR #465, opened 2015-09-25
Remove check for ant version

rosen-vladimirov:
Ant had been used in older versions of NativeScript CLI. We do not use it
anymore, so remove the antVersion check.


justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

teobugslayer:
:+1:

PR #466, opened 2015-09-25
Fix printMarkdown method parameters

Mitko-Kerezov:
Accept ...args rather than a single message
Ping @rosen-vladimirov


justcodebuilduser:
:heart:

rosen-vladimirov:
:heart: :+1:

PR #467, opened 2015-09-29
Plugin variables

Fatme:


justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

plugin-variables-helper.ts
```diff
@@ -0,0 +1,93 @@
...
+						convertedObject[`${propKey}.${innerPropKey}`] = innerPropValue;
+					});
+				}
+
+			});
+
+			return convertedObject;
+		}
+
+		return obj;
```
teobugslayer: If we call this function with a number, it will return the number unchanged. Is this the proper behavior? Shouldn't we assert on non-object arguments?

teobugslayer:
I'd write some unit tests on this code

plugin-variables-helper.ts
```diff
@@ -0,0 +1,93 @@
+///<reference path=".d.ts"/>
+"use strict";
+
+export class PluginVariablesHelper implements IPluginVariablesHelper {
+	constructor(private $options: ICommonOptions) { }
+
+	/**
+	 * Checks if the specified CordovaPluginVariable exists in the --var option specified by user.
```
ligaz: Is this Cordova related?

Fatme:
We have unit tests but they are in icenium repo. I'll move them here :)

justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

plugin-variables-helper.ts
Fatme: If we call this function with a number, it will return the passed number. I think this is the correct behavior.

Fatme:
Actually the unit tests for this code in icenium-cli repo are integrated with the tests for plugins. I think we should separate the code but maybe I'll do it in another PR.

Mitko-Kerezov:
:+1:

PR #471, opened 2015-10-01
Don't cache installed applications

Fatme:
If livesync android --watch command is executed and the application is not deployed on device, {N} deploys the app on every change. This happens because the installed applications are cached and actually the new application never appears in installed applications.

Fixes https://github.com/NativeScript/nativescript-cli/issues/977


teobugslayer:
I think on physically connected Samsung devices the cache is really helpful. We spoke in person and decided to leave the cache but invalidate it when needed.

teobugslayer:
:+1:

justcodebuilduser:
:broken_heart:

justcodebuilduser:
:broken_heart:

PR #481, opened 2015-10-12
Fix tns emulate ios

Fatme:
If `tns emulate ios` command is executed, `undefined is not a function` error is thrown


justcodebuilduser:
:heart:

teobugslayer:
:+1:

PR #484, opened 2015-10-13
Restructure error messaging

Mitko-Kerezov:
References [#300626](http://teampulse.telerik.com/view#item/300626)

Introduces a new messaging service which reads messages from given **.json** files. There is one such file placed in this library and is loaded by default. Its **keys can be overriden** by providing other, client-specific .json files containing the same keys but with a different value.

In order to achieve autocompletion on referencing these error messages a new generation command is introduced - **dev-generate-messages**. It produces a .d.ts interface, as well as a .ts implementation file, which contain all keys from all provided json files as properties. If run as-is it generates files for the client from which the command is launched. If `--default` option is provided generates messages in common lib only.

In order to apply this change I've **moved some of the code-generation logic** related to AppBuilder-CLI's swagger here. The **only change other than physically moving these files** is the introduction of a non-mandatory field - `endingCharacter?: string;` - needed in order to generate Typescript properties more aesthetically

Sample usage of this whole effort can be seen in `devices-services.ts` file in this PR.

### DIFF
--- a/codeGeneration/code-printer.ts
+++ b/codeGeneration/code-printer.ts
@@ -2,9 +2,9 @@
 "use strict";
 
 import {EOL} from "os";
-import codeEntityLib = require("./code-entity");
+import {CodeEntityType} from "./code-entity";
 
-export class SwaggerCodePrinter {
+export class CodePrinter {
 	private static INDENT_CHAR = "\t";
 	private static NEW_LINE_CHAR = EOL;
 	private static START_BLOCK_CHAR = "{";
@@ -15,39 +15,39 @@ export class SwaggerCodePrinter {
 
 		if(block.opener) {
 			content += block.opener;
-			content += SwaggerCodePrinter.START_BLOCK_CHAR;
-			content += SwaggerCodePrinter.NEW_LINE_CHAR;
+			content += CodePrinter.START_BLOCK_CHAR;
+			content += CodePrinter.NEW_LINE_CHAR;
 		}
 
 		_.each(block.codeEntities, (codeEntity: CodeGeneration.ICodeEntity) => {
-			if(codeEntity.codeEntityType === codeEntityLib.CodeEntityType.Line) {
+			if(codeEntity.codeEntityType === CodeEntityType.Line) {
 				content += this.composeLine(<CodeGeneration.ILine>codeEntity, indentSize + 1);
-			} else if(codeEntity.codeEntityType === codeEntityLib.CodeEntityType.Block){
+			} else if(codeEntity.codeEntityType === CodeEntityType.Block){
 				content += this.composeBlock(<CodeGeneration.IBlock>codeEntity, indentSize + 1);
 			}
 		});
 
 		if(block.opener) {
 			content += this.getIndentation(indentSize);
-			content += SwaggerCodePrinter.END_BLOCK_CHAR;
-			content += block.endingCharacter ? block.endingCharacter : '';
+			content += CodePrinter.END_BLOCK_CHAR;
+			content += block.endingCharacter || '';
 		}
 
-		content += SwaggerCodePrinter.NEW_LINE_CHAR;
+		content += CodePrinter.NEW_LINE_CHAR;
 
 		return content;
 	}
 
 	private getIndentation(indentSize: number): string {
-		return Array(indentSize).join(SwaggerCodePrinter.INDENT_CHAR);
+		return Array(indentSize).join(CodePrinter.INDENT_CHAR);
 	}
 
 	private composeLine(line: CodeGeneration.ILine, indentSize: number): string {
 		let content = this.getIndentation(indentSize);
 		content += line.content;
-		content += SwaggerCodePrinter.NEW_LINE_CHAR;
+		content += CodePrinter.NEW_LINE_CHAR;
 
 		return content;
 	}
 }
-$injector.register("swaggerCodePrinter", SwaggerCodePrinter);
+$injector.register("swaggerCodePrinter", CodePrinter);
--- a/codeGeneration/code-printer.ts
+++ b/codeGeneration/code-printer.ts
@@ -1,0 +1,53 @@
+///<reference path="../.d.ts"/>
+"use strict";
+
+import {EOL} from "os";
+import codeEntityLib = require("./code-entity");
+
+export class SwaggerCodePrinter {
+	private static INDENT_CHAR = "\t";
+	private static NEW_LINE_CHAR = EOL;
+	private static START_BLOCK_CHAR = "{";
+	private static END_BLOCK_CHAR = "}";
+
+	public composeBlock(block: CodeGeneration.IBlock, indentSize: number = 0): string {
+		let content = this.getIndentation(indentSize);
+
+		if(block.opener) {
+			content += block.opener;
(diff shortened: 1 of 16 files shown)


justcodebuilduser:
:broken_heart:

justcodebuilduser:
:broken_heart:

justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

codeGeneration/code-printer.ts
```diff
@@ -0,0 +1,53 @@
...
+				content += this.composeLine(<CodeGeneration.ILine>codeEntity, indentSize + 1);
+			} else if(codeEntity.codeEntityType === codeEntityLib.CodeEntityType.Block){
+				content += this.composeBlock(<CodeGeneration.IBlock>codeEntity, indentSize + 1);
+			}
+		});
+
+		if(block.opener) {
+			content += this.getIndentation(indentSize);
+			content += SwaggerCodePrinter.END_BLOCK_CHAR;
+			content += block.endingCharacter ? block.endingCharacter : '';
```
teobugslayer: `block.endingCharacter || ''`

commands/generate-messages.ts
```diff
@@ -0,0 +1,24 @@
...
+				private $staticConfig: Config.IStaticConfig) {
+	}
+
+	allowedParameters: ICommandParameter[] = [];
+
+	execute(args: string[]): IFuture<void> {
+		return (() => {
+			let result = this.$messageContractGenerator.generate().wait(),
+				interfaceFilePath = this.$staticConfig.CLIENT_NAME ? path.join(__dirname, "../../messages.d.ts") : path.join(__dirname, "../messages/messages.d.ts"),
+				implementationFilePath = this.$staticConfig.CLIENT_NAME ? path.join(__dirname, "../../messages.ts") : path.join(__dirname, "../messages/messages.ts");
```
teobugslayer: You can share the check between these two lines. Extract the directory path into a separate variable and then append the file name.

resources/messages/errorMessages.json
```diff
@@ -0,0 +1,7 @@
+{
+	"DEVICES": {
```
teobugslayer: These are not constants in the code. Can we use pascalCase or CamelCase for the keys? Currently they make my eyes bleed.

services/message-contract-generator.ts
```diff
@@ -0,0 +1,83 @@
...
+	}
+
+	public generate(): IFuture<IServiceContractClientCode> {
+		return ((): IServiceContractClientCode => {
+			let interfacesFile= new Block();
+			let implementationsFile = new Block();
+			let definitionsPath = `"${this.$staticConfig.CLIENT_NAME ? "" : "../"}.d.ts"`;
+
+			implementationsFile.writeLine(`///<reference path=${definitionsPath}/>`);
+			implementationsFile.writeLine("\"use strict\";");
```
teobugslayer: You can use single quotes to make the code more readable

services/messages-service.ts
```diff
@@ -0,0 +1,88 @@
...
+
+import * as util from "util";
+import * as path from "path";
+
+export class MessagesService implements IMessagesService {
+	private _pathsToMessageJsonFiles: string[] = null;
+	private _messageJsonFilesContentsCache: any[] = null;
+
+	private get pathToDefaultMessageJson(): string {
+		return path.join(path.dirname(__dirname), "resources", "messages", "errorMessages.json");
```
teobugslayer: A very peculiar construction: `path.dirname(__dirname)`

services/messages-service.ts
```diff
@@ -0,0 +1,88 @@
...
+		}
+
+		return this.getMessageFromJsonRecursive(keys, jsonValue, index + 1);
+	}
+
+	private refreshMessageJsonContentsCache(): void {
+		this._messageJsonFilesContentsCache = [];
+		_.each(this.pathsToMessageJsonFiles, path => {
+			if (!this.$fs.exists(path).wait()) {
+				throw new Error("Message json file " + path + " does not exist.");
```
teobugslayer: The code is correct, but is there a reason not to use IErrors.fail?

services/messages-service.ts
```diff
@@ -0,0 +1,88 @@
...
+			}
+
+			this._messageJsonFilesContentsCache.push(this.$fs.readJson(path).wait());
+		});
+	}
+
+	private getFormatedMessage(message: string, ...args: string[]): string {
+		return ~message.indexOf("%") ? util.format.apply(null, [message, ...args]) : message;
```
teobugslayer: You should be able to use ES6 ... here instead of the apply call.

codeGeneration/code-printer.ts
```diff
@@ -0,0 +1,53 @@
+///<reference path="../.d.ts"/>
+"use strict";
+
+import {EOL} from "os";
+import codeEntityLib = require("./code-entity");
+
+export class SwaggerCodePrinter {
```
Fatme: Maybe we should rename this class - for example to `CodePrinter`

services/messages-service.ts
Fatme: Can we use `$resources.getFilePath()`?

services/messages-service.ts
```diff
@@ -0,0 +1,88 @@
...
+		return this._messageJsonFilesContentsCache;
+	}
+
+	constructor(private $fs: IFileSystem) {
+		this._pathsToMessageJsonFiles = [this.pathToDefaultMessageJson];
+	}
+
+	public get pathsToMessageJsonFiles(): string[] {
+		if (!this._pathsToMessageJsonFiles) {
+			throw new Error("No paths to message json files provided.");
```
Fatme: `$errors.fail` maybe?

services/messages-service.ts
Fatme: Can we use `map` ?

codeGeneration/code-printer.ts
```diff
@@ -0,0 +1,53 @@
+///<reference path="../.d.ts"/>
+"use strict";
+
+import {EOL} from "os";
+import codeEntityLib = require("./code-entity");
```
teobugslayer: Use ES6 import here

services/messages-service.ts
Mitko-Kerezov: Sadly `$resources` is not registered anywhere in the boundaries of the common library so I can't `import` it

services/messages-service.ts
Mitko-Kerezov: @teobugslayer I've chosen `new Error()` out of fear of maximum call stack. Our `$errors` class resolves this `messageService` when getting the message which should be printed on the screen. Hypothetically we could end up in a bottomless recursion

services/messages-service.ts
Mitko-Kerezov: Might cause stack overflow exception - see 3 comments up

justcodebuilduser:
:broken_heart:

justcodebuilduser:
:heart:

Mitko-Kerezov:
Addressed all issues, ping @teobugslayer @Fatme and @rosen-vladimirov for a second review

rosen-vladimirov:
:+1:

PR #487, opened 2015-10-20
Improve prompter behavior in interactive console

Mitko-Kerezov:
If all schemas have default actions just execute them instead of throwing.
If any schema does not however have a default action throw.


justcodebuilduser:
:heart:

teobugslayer:
:+1:

PR #488, opened 2015-10-21
Allow defining subcommands of subcommands

rosen-vladimirov:
In case you have command called 'prop|print' and you want to add subcommand to it,
the injector will not try to resolve it. For example 'prop|print|myprop' will be resolved
as command 'prop|print' with argument 'myprop'.

Fix this by executing the full `_.each` and adjust the real parameters based on the last matching one.


justcodebuilduser:
:heart:

teobugslayer:
:+1: but we may need some more unit tests

rosen-vladimirov:
@teobugslayer , I've added some tests, you can check them if you want :)

rosen-vladimirov:
:+1:  from @Mitko-Kerezov

PR #490, opened 2015-10-22
Move sdk option to common lib

Fatme:


justcodebuilduser:
:heart:

teobugslayer:
Why is this change needed? Does it affect AppBuilder CLI?

Fatme:
Because I need to pass --sdk option to ios-sim.

teobugslayer:
:+1:

PR #492, opened 2015-10-26
Add renameIfExists method to fs

rosen-vladimirov:
Add renameIfExists method to fs. Used to skip ENOENT errors when rename is called directly. Renames specified file to the specified name only in case it exists.
Returns True in case of successful rename. False in case the file does not exist.

### DIFF
--- a/declarations.d.ts
+++ b/declarations.d.ts
@@ -53,6 +53,14 @@ interface IFileSystem {
 	isRelativePath(path: string): boolean /* feels so lonely here, I don't have a Future */;
 	ensureDirectoryExists(directoryPath: string): IFuture<void>;
 	rename(oldPath: string, newPath: string): IFuture<void>;
+	/**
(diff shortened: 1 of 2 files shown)


justcodebuilduser:
:heart:

teobugslayer:
This is non-atomic operation and subject to all race conditions imaginable. Perhaps better looking code:

```
try {
  fs.rename();
  return true
catch(e) {
  if (e.code === ENOENT) {
    return false;
  }
  throw e;
}
```

rosen-vladimirov:
@teobugslayer , you are correct, I've fixed the code and the tests are passing (I've added them in appbuilder-cli for the moment, but I'll move all `fs` tests here in a separate PR).

teobugslayer:
:+1:

justcodebuilduser:
:heart:

PR #493, opened 2015-10-26
Fix livesync file watch exclusion check.

hdeshev:
The check for modified files being excluded from livesync wasn't working.
Fixed by using the same code from the initial sync.


justcodebuilduser:
:heart:

teobugslayer:
Nice catch! :+1:

PR #498, opened 2015-10-28
multiple hook script support

tailsu:


justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

teobugslayer:
:+1:

helpers.ts
```diff
@@ -157,3 +157,43 @@ export function appendZeroesToVersion(version: string, requiredVersionLength: nu
...
+		}
+		return hooksService;
+	}
+
+	return decorateMethod(
+		(self: any, args: any[]) => {
+			getHooksService(self).executeBeforeHooks(commandName).wait();
+		},
+		(self: any, resultPromise: any, args: any[]) => {
+			let result = resultPromise.wait();
```
Fatme: We're expecting `resultPromise` to return `future`. Is it possible to return something different?

services/commands-service.ts
```diff
@@ -56,10 +56,9 @@ export class CommandsService implements ICommandsService {
 						commandName = helpers.stringReplaceAll(commandName, CommandsService.HIERARCHICAL_COMMANDS_DELIMITER, CommandsService.HOOKS_COMMANDS_DELIMITER);
 					}
 
-					this.$hooksService.initialize(commandName);
-					this.$hooksService.executeBeforeHooks().wait();
+					this.$hooksService.executeBeforeHooks(commandName).wait();
 					command.execute(commandArguments).wait();
-					this.$hooksService.executeAfterHooks().wait();
+					this.$hooksService.executeAfterHooks(commandName).wait();
```
Fatme: Can we use `helpers.hook` to decorate build and deploy commands in `icenium-cli` and remove this code?

helpers.ts
tailsu: So far - no. I decided to keep it simple. If need be, it can be extended.

services/commands-service.ts
tailsu: We could, but it would require a lot of code. I'm not sure that's good. Although, it kind of sucks that the two CLI's have different hooking mechanisms.

PR #502, opened 2015-10-30
verify current node version against allowed list

teobugslayer:
update common lib list of verified node versions

See http://teampulse.telerik.com/view#item/304396


enforce-node-version.ts
```diff
@@ -0,0 +1,3 @@
+/**
+ * Created by totev on 30.10.2015 г..
```
rosen-vladimirov: I believe we do not need this file

justcodebuilduser:
:heart:

rosen-vladimirov:
:+1:  after removing the `enforce-node-version.ts` file :)

justcodebuilduser:
:heart:

PR #507, opened 2015-11-03
Fix leaking of adb processes

rosen-vladimirov:
When we print device logs, we execute `adb -s <device identifier> logcat`.
In case the device is not unlocked, we receive `undefined` for deviceIdentifier and the command leaks adb processes.
It looks like executing `adb -s <invalid id> logcat` is not closed even after our process exits.
Fix this by checking the passed value. Also change the return type of `LogcatHelper`'s `start` method to void
as noone is using the returned child_process.

### DIFF
--- a/mobile/android/logcat-helper.ts
+++ b/mobile/android/logcat-helper.ts
@@ -10,28 +10,28 @@ export class LogcatHelper implements Mobile.ILogcatHelper {
 			private $logger: ILogger,
 			private $staticConfig: Config.IStaticConfig) { }
 
-	public start(deviceIdentifier: string): any {
-		let adbPath = this.$staticConfig.getAdbFilePath().wait();
-		this.$childProcess.exec(util.format("%s -s %s logcat -c", adbPath, deviceIdentifier)).wait(); // remove cached logs
-		let adbLogcat = this.$childProcess.spawn(adbPath, ["-s", deviceIdentifier, "logcat"]);
-		let lineStream = byline(adbLogcat.stdout);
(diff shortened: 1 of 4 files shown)


justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

mobile/android/logcat-helper.ts
```diff
@@ -10,28 +10,28 @@ export class LogcatHelper implements Mobile.ILogcatHelper {
...
 
-	public start(deviceIdentifier: string): any {
-		let adbPath = this.$staticConfig.getAdbFilePath().wait();
-		this.$childProcess.exec(util.format("%s -s %s logcat -c", adbPath, deviceIdentifier)).wait(); // remove cached logs
-		let adbLogcat = this.$childProcess.spawn(adbPath, ["-s", deviceIdentifier, "logcat"]);
-		let lineStream = byline(adbLogcat.stdout);
+	public start(deviceIdentifier: string): void {
+		if(deviceIdentifier) {
+			let adbPath = this.$staticConfig.getAdbFilePath().wait();
+			this.$childProcess.exec(util.format("%s -s %s logcat -c", adbPath, deviceIdentifier)).wait(); // remove cached logs
```
teobugslayer: consider spawning instead of execing

rosen-vladimirov: this will be handled in a separate PR (related to all changes I'm making in NativeScript CLI)

Mitko-Kerezov:
:+1:

PR #514, opened 2015-11-08
Fix printing of adb logcat

rosen-vladimirov:
Also add quotes around path to adb, so even when the path contains spaces, the command will be executed without troubles.

Fix checking for javac version to use JAVA_HOME in case it is defined.

### DIFF
--- a/mobile/android/logcat-helper.ts
+++ b/mobile/android/logcat-helper.ts
@@ -1,7 +1,6 @@
 ///<reference path="../../.d.ts"/>
 "use strict";
 import byline = require("byline");
-import * as util from "util";
 
 export class LogcatHelper implements Mobile.ILogcatHelper {
 	constructor(private $childProcess: IChildProcess,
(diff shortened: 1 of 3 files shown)


justcodebuilduser:
:heart:

mobile/android/logcat-helper.ts
```diff
@@ -13,7 +13,7 @@ export class LogcatHelper implements Mobile.ILogcatHelper {
 	public start(deviceIdentifier: string): void {
 		if(deviceIdentifier) {
 			let adbPath = this.$staticConfig.getAdbFilePath().wait();
-			this.$childProcess.exec(util.format("%s -s %s logcat -c", adbPath, deviceIdentifier)).wait(); // remove cached logs
+			this.$childProcess.exec(util.format('"%s" -s "%s" logcat -c', adbPath, deviceIdentifier)).wait(); // remove cached logs
```
ligaz: You can use template strings here.

justcodebuilduser:
:heart:

mobile/android/logcat-helper.ts
```diff
@@ -13,7 +12,8 @@ export class LogcatHelper implements Mobile.ILogcatHelper {
 	public start(deviceIdentifier: string): void {
 		if(deviceIdentifier) {
 			let adbPath = this.$staticConfig.getAdbFilePath().wait();
-			this.$childProcess.exec(util.format("%s -s %s logcat -c", adbPath, deviceIdentifier)).wait(); // remove cached logs
+			// remove cached logs:
+			this.$childProcess.exec(`"${adbPath}" -s "${deviceIdentifier}" logcat -c`).wait();
```
teobugslayer: can we use spawn / execFile instead?

justcodebuilduser:
:heart:

teobugslayer:
:+1:

justcodebuilduser:
:heart:

PR #522, opened 2015-11-16
Show message after successful deploy

rosen-vladimirov:
Get back the message that reports successful deploy on android device. We've deleted it a couple of commits ago.


justcodebuilduser:
:heart:

Mitko-Kerezov:
:+1:

PR #527, opened 2015-11-16
Use emulator executable from ANDROID_HOME

rosen-vladimirov:
In case ANDROID_HOME is defined, use the emulator executable based on it, not from PATH environment variable.


justcodebuilduser:
:heart:

PR #529, opened 2015-11-17
List dev commands when searching help pages

rosen-vladimirov:
When we search for correct help page, we have to list dev commands as well as we want to show the help of dev-test command in NS CLI.


justcodebuilduser:
:heart:

Fatme:
:+1:

PR #544, opened 2015-11-24
Don't duplicate the messages from logcat

Fatme:
Fixes https://github.com/NativeScript/nativescript-cli/issues/1130

### DIFF
--- a/mobile/android/logcat-helper.ts
+++ b/mobile/android/logcat-helper.ts
@@ -3,14 +3,18 @@
 import byline = require("byline");
 
 export class LogcatHelper implements Mobile.ILogcatHelper {
+	private mapDeviceToLoggingStarted: IDictionary<boolean>;
+
 	constructor(private $childProcess: IChildProcess,
 			private $deviceLogProvider: Mobile.IDeviceLogProvider,
 			private $devicePlatformsConstants: Mobile.IDevicePlatformsConstants,
 			private $logger: ILogger,
-			private $staticConfig: Config.IStaticConfig) { }
+			private $staticConfig: Config.IStaticConfig) {
+				this.mapDeviceToLoggingStarted = Object.create(null);
+			}
 
 	public start(deviceIdentifier: string): void {
-		if(deviceIdentifier) {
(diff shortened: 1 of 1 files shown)


justcodebuilduser:
:heart:

teobugslayer:
:+1:

mobile/android/logcat-helper.ts
```diff
@@ -10,7 +12,7 @@ export class LogcatHelper implements Mobile.ILogcatHelper {
 			private $staticConfig: Config.IStaticConfig) { }
 
 	public start(deviceIdentifier: string): void {
-		if(deviceIdentifier) {
+		if (deviceIdentifier && !this.isStarted) {
```
rosen-vladimirov: this check should be per deviceIdentifier or we'll always have logs only for the first connected device (this will break Proton) :)

justcodebuilduser:
:heart:

rosen-vladimirov:
Maybe I'm not correct, but what will happen in the following scenario:
1) We have long living process (Proton) and we attach device with identifier `A`.
2) We call logcatHelper for this device.
3) `A` is added to `mapDeviceToLoggingStarted` dictionary.
4) After some time the device is disconnected. `A` remains in the dictionary. `adb` process for logcat is dead.
5) `A` is connected again.
6) LogcatHelper is called for `A`, but as `A` is already in `mapDeviceToLoggingStarted` dictionary, we do not start new process for `adb logcat`.
7) We do not receive logs for `A`

Maybe a better solution is to add new property to IDevice that indicates if logging is started and it will call LogcatHelper internally. As when new device is attached, I believe we create new instance of IDevice, so the above scenario could work fine there. Of course there could be various other solutions :)

justcodebuilduser:
:heart:

Fatme:
If some device is disconnected, `logcat close` event will be raised. We'll execute this logic - https://github.com/telerik/mobile-cli-lib/pull/544/files#diff-618ae34084a745345392b83e2a7b243eR29 so the next time the device is connected, we will attach again on `logcat data` event and will receive logs from a device.

rosen-vladimirov:
great, :+1:

PR #545, opened 2015-11-25
Fix cocoapods check and add unit tests for doctor service

rosen-vladimirov:
When cocoapods are not installed on Mac machine, doctor is failing.
Fix this by skipping match check in case pods are not installed.
Add basic unit tests for doctor service.

Add unit test for the failing cocoapods check.

Fixes https://github.com/NativeScript/nativescript-cli/issues/1245

### DIFF
--- a/declarations.d.ts
+++ b/declarations.d.ts
@@ -366,10 +366,11 @@ interface ISysInfoData {
 interface ISysInfo {
 	/**
 	 * Returns information for the current system.
+	 * @param {string} pathToPackageJson Path to package.json of the CLI.
 	 * @param {any} androidToolsInfo Defines paths to adb and android executables.
 	 * @return {IFuture<ISysInfoData>} Object containing information for current system.
 	 */
-	getSysInfo(androidToolsInfo?: {pathToAdb: string, pathToAndroid: string}): IFuture<ISysInfoData>;
+	getSysInfo(pathToPackageJson: string, androidToolsInfo?: {pathToAdb: string, pathToAndroid: string}): IFuture<ISysInfoData>;
 }
 
 interface IHostInfo {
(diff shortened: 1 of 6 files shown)


justcodebuilduser:
:broken_heart:

justcodebuilduser:
:broken_heart:

justcodebuilduser:
:broken_heart:

justcodebuilduser:
:broken_heart:

justcodebuilduser:
:heart:

declarations.d.ts
```diff
@@ -559,3 +560,106 @@ interface IPluginVariablesHelper {
 	getPluginVariableFromVarOption(variableName: string, configuration?: string): any;
 	simplifyYargsObject(obj: any, configuration?: string): any;
 }
+
+/**
+ * Describes Registry values are returned from winreg
+ */
+interface IWinRegResult {
```
teobugslayer: this should go into a separate .d.ts, potentially donating it to DefinitelyTyped

sys-info-base.ts
```diff
@@ -115,30 +115,26 @@ export class SysInfoBase implements ISysInfo {
...
+		let output: string;
+		try {
+			output = this.readRegistryValue("ProductName").wait() + " " +
+					this.readRegistryValue("CurrentVersion").wait() + "." +
+					this.readRegistryValue("CurrentBuild").wait();
+		} catch (err) {
+			this.$logger.trace(err);
+		}
+
+		return output;
```
teobugslayer: I do not like that we are returning undefined here. Can we return null instead?

winreg.ts
```diff
@@ -0,0 +1,39 @@
+///<reference path=".d.ts"/>
+"use strict";
+
+import Future = require("fibers/future");
+let Registry = require("winreg");
```
teobugslayer: we really need a .d.ts which will allow us to import winreg here.

teobugslayer:
Tests are awesome!

declarations.d.ts
```diff
@@ -559,3 +560,106 @@ interface IPluginVariablesHelper {
...
+	 */
+	HKCC: IHiveId;
+
+	/**
+	 * HKEY_USERS
+	 */
+	HKU: IHiveId;
+}
+
+
```
rosen-vladimirov: multiple blank lines

declarations.d.ts
rosen-vladimirov: These interfaces are for our usage only, we cannot donate them directly to DefinitelyTyped.

justcodebuilduser:
:heart:

winreg.ts
rosen-vladimirov: I'll try to prepare `winreg.d.ts` and add it here and send PR to DefinitelyTyped. However this will require some time, as I'll have to add unit tests and read they contribution guidelines. Can we live this way for the moment?

justcodebuilduser:
:heart:

teobugslayer:
We can extract the definitions in another PR, :+1: for this one.

PR #553, opened 2015-12-02
Add canStartApplication to applicationManager

rosen-vladimirov:
We have different cases when we can start the application, based on current platform and device platform.
Add `canStartApplication` in applicationManager's which evaluates all circumstances and can be used to check if we are able to start the application on device.


justcodebuilduser:
:heart:

Mitko-Kerezov:
:+1:

PR #555, opened 2015-12-03
Fix getSysInfo when paths to android tools are undefined

rosen-vladimirov:
When paths to android tools are undefined, we call childProcess exec and spawnFromEvent with undefined commands.
This leads to futures left behind as childProcess cannot be created and our future is never resolved.
Do not call exec and spawnFromEvent in case command is empty.


justcodebuilduser:
:heart:

teobugslayer:
:+1:

vchimev:
:+1:

PR #570, opened 2015-12-15
Introduce proton logger

rosen-vladimirov:
Fix `EPERM: operation not permitted, write` error when started from Proton installation and trying to write to stdout.
The process does not have stdout, so all logger calls will cause such failure.
Introduce new logger, special for Proton, that will not do anything at the moment. We'll extend it in the future to send messages to Proton with the messages.

Set version to 0.1.3

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,9 +7,9 @@ Contains common infrastructure for CLIs - mainly AppBuilder and NativeScript.
 Installation
 ===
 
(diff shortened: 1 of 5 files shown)


justcodebuilduser:
:heart:

teobugslayer:
:+1:

PR #571, opened 2015-12-16
Improve device discovery and fix some issues

rosen-vladimirov:
### Improve device discovery for Proton by fixing:
* Crash when iPhone is not trusted.
* Crash when error is raised during device detection. From now on the error is caught. We'll report it to Proton (this will be handled in separate PR).
* Crash when trying to provide sysLogs for unauthorized devices.
* In case more than one Android device has "USB Debugging" disabled, we were showing only one device. Show all devices from now on.

### Add new properties to IDeviceInfo inteface:
* status - string, describing the status of the device, can be `Connected` or `Unauthorized`. Could be expanded in the future.
* errorHelp - in case status is `Unauthorized` we provide some information to the user how to resolve the issue. In case status is `Connected` this should be null.
* color - Color of the device. Not mandatory part of the Interface. Currently provided only for iOS devices.

### Additional changes, implementation details:
Change the way we work with iOS devices - currently if the device is Unauthorized, we were just failing, without creating instance of the iOSDevice.
From now on we create instance and set the values that we cannot get from the device (like Name) to null.
Added **ADNCI_MSG_TRUSTED=4** to iOSDeviceDiscovery. I'm not sure the value is correct, but I've received it when I select "Trust" option on iOS device.
When we receive 4, we remove the old Untrusted device and add a new one with all correct properties.
Keep current behavior in `AndroidDeviceDiscovery` to report device as newly attached when its status is changed.

### DIFF
--- a/appbuilder/proton-bootstrap.ts
+++ b/appbuilder/proton-bootstrap.ts
@@ -2,6 +2,7 @@
 "use strict";
 
 require("../bootstrap");
+$injector.require("messages", "./messages/messages");
 $injector.require("logger", "./appbuilder/proton-logger");
 
 import {OptionsBase} from "../options";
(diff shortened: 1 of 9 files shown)


justcodebuilduser:
:heart:

rosen-vladimirov:
ping @ligaz , @tailsu , @Mitko-Kerezov

justcodebuilduser:
:heart:

Mitko-Kerezov:
:+1: great work!

justcodebuilduser:
:heart:

PR #572, opened 2015-12-18
Fix android and ios device detection

rosen-vladimirov:
Fix android device detection - remove method was not correct and was never removing device.
Fix ios device detection - always set status to Unreachable when there's error.

Change name of Unauthorized status to Unreachable.


rosen-vladimirov:
:+1:  from @Mitko-Kerezov and @nadyaA

justcodebuilduser:
:heart:

PR #577, opened 2015-12-29
Add device log filtering

rosen-vladimirov:
* Add AndroidLogFilter - specific filtering for Android devices
* Add IOSLogFilter - specific filtering for iOS devices
* Add LogFilter class - common filtering logic. Resolves androidLogFilter or iOSLogFilter based on the input.
* Show full output when `<cli> device log` command is used.
* Add unit tests for androidLogFilter
* Add unit tests for iOSLogFilter
* Add unit tests for LogFilter
* Make sure isFuture method returns boolean.
* Expose setLogLevel method in devicesService (used for Proton)

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,9 +7,9 @@ Contains common infrastructure for CLIs - mainly AppBuilder and NativeScript.
 Installation
 ===
 
-Latest version: 0.3.0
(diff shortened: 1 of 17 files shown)


justcodebuilduser:
:heart:

PR #581, opened 2016-01-07
Fix deploy on iOS devices

rosen-vladimirov:
When we deploy on iOS devices, the first sent package sometimes receives error code 21. It turned out this error is not breaking the deployment, so ignore it once and continue deploying.
The error occurs only when we have parallel futures executing for device detection.

Also when error is raised during application starting, we reject the Promise for deploy on device. This is not correct, so ignore such errors and resolve the promise.

One fix to rule them all: http://teampulse.telerik.com/view#item/307776

### DIFF
--- a/mobile/ios/ios-proxy-services.ts
+++ b/mobile/ios/ios-proxy-services.ts
@@ -5,6 +5,7 @@ import * as path from "path";
 import * as iOSCore from "./ios-core";
 import * as helpers from "../../helpers";
 import * as plistlib from "plistlib";
+import Future = require("fibers/future");
 
 export class MobileServices {
 	public static APPLE_FILE_CONNECTION: string = "com.apple.afc";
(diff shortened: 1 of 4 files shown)


justcodebuilduser:
:heart:

bundyo:
:+1:

mobile/ios/ios-proxy-services.ts
```diff
@@ -169,19 +171,46 @@ export class AfcClient implements Mobile.IAfcClient {
...
+					if(err.message.indexOf("Result is: '21'") !== -1) {
+						errorCode21Count++;
+
+						if(errorCode21Count === 1) {
+							this.$logger.warn(err.message);
+						} else {
+							futureThrow(err);
+						}
+					} else {
+						futureThrow(err);
```
teobugslayer: two else clauses can be combined.

mobile/mobile-core/devices-service.ts
```diff
@@ -103,8 +103,8 @@ export class DevicesService implements Mobile.IDevicesService {
 				setInterval(() => {
 					fiberBootstrap.run(() => {
 						try {
-							Future.wait([this.$iOSDeviceDiscovery.checkForDevices(),
-										this.$androidDeviceDiscovery.checkForDevices()]);
+							this.$iOSDeviceDiscovery.checkForDevices().wait();
```
teobugslayer: I think we definitely need a comment why we are not using Future.wait().

justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

Mitko-Kerezov:
:+1:

PR #583, opened 2016-01-15
Fix crash when launching applications on iOS device in debug mode

Fatme:
Fixes https://github.com/NativeScript/nativescript-cli/issues/1362

### DIFF
--- a/definitions/mobile.d.ts
+++ b/definitions/mobile.d.ts
@@ -78,6 +78,12 @@ declare module Mobile {
 		 * Available for iOS only - the value of device's 'DeviceColor' property.
 		 */
 		color?: string;
+
+		/**
+		 *  Optional property describing the architecture of the device
+		 *  Available for iOS only - can be "armv7" or "arm64"
+		 */
+		activeArchitecture?: string;
 	}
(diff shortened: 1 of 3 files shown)


justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

teobugslayer:
:+1:

mobile/ios/ios-device.ts
```diff
@@ -54,6 +54,21 @@ export class IOSDevice implements Mobile.IiOSDevice {
 			this.deviceInfo.version = this.getValue("ProductVersion");
 			this.deviceInfo.color = this.getValue("DeviceColor");
 			this.deviceInfo.isTablet = productType && productType.toLowerCase().indexOf("ipad") !== -1;
+
+			productType = productType.toLowerCase().trim();
+			let majorVersionAsString = _.last(productType.split(",")[0]);
```
Fatme: It is better to use regex here in order to works correctly if productType is for example `iphone10,2`

justcodebuilduser:
:heart:

PR #589, opened 2016-01-21
Add isAppInstalledOnDevices public method

rosen-vladimirov:
Order of commits:
### Fix uncaughtExceptionListener for our unit tests
When staticConfig is not registered (in our tests), when uncaught exception is raised, the listener fails with error that staticConfig cannot be resolved. This makes it hard for debugging. Try-catch the resolve of staticConfig and set the disalbeAnalytics to true when staticConfig cannot be resolved.    This way the tests will not have to register analyticsService.

### Add error property for future objects
Update `fibers.d.ts` with new property that exists in the object - error - required for new unit tests. When the future throws, calling `future.get()` throws. In such cases first check future.error, for example:
```TypeScript
 let futures: IFuture<void>[] = [];
 // fill lots of futures
 try {
	 Future.wait(futures);
 } catch(err) {
 }

 _.each(futures, f => {
	 if(!f.error) {
		 f.get();
	 }
 })
 ```

### Add additional logging for spawnFromEvent

Currently when adb fails, we skip the error and it is hard to understand that it has failed.In fact adb writes to stderr in some cases. In order to debug such cases (until we implement proper parsing), add additional logging for the spawnFromEvent method.

### Add isAppInstalledOnDevices public method
The method return array of boolean Promises. All of them will indicate if app is installed on the specified device identifier. Add unit tests for devicesService and fix some issues.

### DIFF
--- a/mobile/mobile-core/devices-service.ts
+++ b/mobile/mobile-core/devices-service.ts
@@ -60,15 +60,21 @@ export class DevicesService implements Mobile.IDevicesService {
 	}
 
 	public isiOSSimulator(device: Mobile.IDevice): boolean {
-		return this.$mobileHelper.isiOSPlatform(device.deviceInfo.platform) && device.isEmulator;
+		return !!(this.$mobileHelper.isiOSPlatform(device.deviceInfo.platform) && device.isEmulator);
 	}
 
 	/* tslint:disable:no-unused-variable */
 	@exported("devicesService")
-	private setLogLevel(logLevel: string, deviceIdentifier?: string): void {
+	public setLogLevel(logLevel: string, deviceIdentifier?: string): void {
 		this.$deviceLogProvider.setLogLevel(logLevel, deviceIdentifier);
 	}
 	/* tslint:enable:no-unused-variable */
+
+	@exportedPromise("devicesService")
+	public isAppInstalledOnDevices(deviceIdentifiers: string[], appIdentifier: string): IFuture<boolean>[] {
+		this.$logger.trace(`Called isInstalledOnDevices for identifiers ${deviceIdentifiers}. AppIdentifier is ${appIdentifier}.`);
+		return _.map(deviceIdentifiers, deviceIdentifier => this.isApplicationInstalledOnDevice(deviceIdentifier, appIdentifier));
+	}
 
 	public getDeviceInstances(): Mobile.IDevice[] {
 		return _.values(this._devices);
@@ -210,9 +216,7 @@ export class DevicesService implements Mobile.IDevicesService {
 
 			let futures = _.map(sortedDevices, (device: Mobile.IDevice) => {
 				if (!canExecute || canExecute(device)) {
-					let future = action(device);
-					Future.settle(future);
-					return future;
+					return action(device);
 				} else {
 					return Future.fromResult();
 				}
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

**Comment by justcodebuilduser** [comment]
:heart:

File: mobile/mobile-core/devices-service.ts [hunk 1]
```diff
@@ -378,6 +401,13 @@ export class DevicesService implements Mobile.IDevicesService {
 		}
 
 		return this.executeOnAllConnectedDevices(action, canExecute);
+}
+
+	private isApplicationInstalledOnDevice(deviceIdentifier: string, appIdentifier: string): IFuture<boolean> {
+		return ((): boolean => {
```
Fatme: You don't need this feature :smile_cat:

rosen-vladimirov: why? In fact I prefer to have it as the method is used for Proton only. The idea is that this future will be waited in the decorator and if it fails, the Promise will be rejected.
In case I remove it and `this.getDeviceByIdentifier` fails, the public method will fail before returning future for this device and it will hang :)

mobile/mobile-core/devices-service.ts
```diff
@@ -296,7 +314,7 @@ export class DevicesService implements Mobile.IDevicesService {
 			}
 
 			if (!this.$hostInfo.isDarwin && this._platform && this.$mobileHelper.isiOSPlatform(this._platform) && this.$options.emulator) {
-				this.$errors.failWithoutHelp("You are not allowed to use iOS simulator on Windows.");
+				this.$errors.failWithoutHelp("You are allowed to use iOS simulator only on MacOS.");
```
Fatme: Maybe OSX?

rosen-vladimirov: OS X or OSX or Mac OSX
According to wikipedia, it should be `OS X`

test/unit-tests/mobile/devices-service.ts
```diff
@@ -0,0 +1,918 @@
...
+		it("returns true when iOS device is passed", () => {
+			assert.deepEqual(devicesService.isiOSDevice(<any>iOSDevice), true);
+		});
+
+		it("returns true when iOS device is passed, assert case insensitivity", () => {
+			assert.deepEqual(devicesService.isiOSDevice(<any>{ deviceInfo: { platform: "iOs"}}), true);
+			assert.deepEqual(devicesService.isiOSDevice(<any>{ deviceInfo: { platform: "IOS"}}), true);
+		});
+
+		it("returns false when iOS device is passed", () => {
```
Fatme: android device :smiley_cat:

test/unit-tests/mobile/devices-service.ts
```diff
@@ -0,0 +1,918 @@
...
+			assert.deepEqual(devicesService.getDeviceByDeviceOption(), androidDevice);
+		});
+
+		it("returns deviceIdentifier when devicesService is initialized with deviceId and platform", () => {
+			androidDeviceDiscovery.emit("deviceFound", androidDevice);
+			devicesService.initialize({ deviceId: androidDevice.deviceInfo.identifier, platform: "android" }).wait();
+			assert.deepEqual(devicesService.getDeviceByDeviceOption(), androidDevice);
+		});
+
+		it("returns deviceIdentifier when devicesService is initialized with deviceId (passed as number) and platforms", () => {
```
Fatme: platform maybe?

justcodebuilduser:
:heart:

mobile/mobile-core/devices-service.ts
Fatme: maybe we should add a comment ;)

Fatme:
:+1: great work

PR #590, opened 2016-01-22
Facilitate disable hooks mechanism

Mitko-Kerezov:
Introduce a way to disable hooks globally through a .json property only.
Ping @rosen-vladimirov


rosen-vladimirov:
:+1:  after green build

justcodebuilduser:
:heart:

PR #591, opened 2016-01-22
Speed up livesync on android device

Fatme:


justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

mobile/android/android-device-file-system.ts
```diff
@@ -39,17 +44,57 @@ export class AndroidDeviceFileSystem implements Mobile.IDeviceFileSystem {
...
 		}).future<void>()();
 	}
 
 	public transferDirectory(deviceAppData: Mobile.IDeviceAppData, localToDevicePaths: Mobile.ILocalToDevicePathData[], projectFilesPath: string): IFuture<void> {
 		return (() => {
-			this.adb.executeCommand(["push", projectFilesPath, deviceAppData.deviceProjectRootPath]).wait();
+			let devicePaths: string[] = [],
+				currentShasums: string[] = [];
+
+			localToDevicePaths.map(localToDevicePathData => {
```
rosen-vladimirov: this should be `.forEach`

rosen-vladimirov:
The following workflow is not working as expected:
* `tns run android`
* `tns livesync android`
* Break the process
* `tns livesync android` without changing anything

We move the "commands" file in `/data/local/tmp/<app-id>/fullsync` and runtime thinks that this is a full application, so it deletes the app files and moves the files from fullsync to real app.
Soo the app does not start on device and all of its files are missing

rosen-vladimirov:
Fixed the above mentioned scenario :)

justcodebuilduser:
:heart:

definitions/mobile.d.ts
```diff
@@ -493,4 +493,32 @@ declare module Mobile {
...
+	interface IAndroidDeviceHashService {
+		/**
+		 * Returns the hash file path on device
+		 */
+		hashFileDevicePath: string;
+		/**
+		 * If hash file exists on device, read the hashes from the file and returns them as array
+		 * If hash file doesn't exist on device, returns null
+		 */
+		getShasumsFromDevice(): IFuture<string[]>;
```
teobugslayer: It is more functunal-style to return an empty array instead of null, but I do not insist.

helpers.ts
```diff
@@ -276,6 +276,20 @@ export function connectEventually(factory: () => net.Socket, handler: (_socket:
 	tryConnect();
 }
 
+export function convertToDictionary(arr: string[]): IStringDictionary {
+	let dictionary: IStringDictionary = Object.create(null);
+	_.each(arr, item => {
+		let [key, value] = item.split(" ");
```
teobugslayer: The file name may contain spaces and this code will fail in such case. We actually defend on other places in this PR against this by using double quotes, like https://github.com/telerik/mobile-cli-lib/pull/591/files#diff-1596e63db9f2f8338a52444248dee614R68 .

mobile/android/android-device-file-system.ts
```diff
@@ -39,16 +44,65 @@ export class AndroidDeviceFileSystem implements Mobile.IDeviceFileSystem {
...
+			let filesToChmodOnDevice: string[] = devicePaths;
+			if (this.$options.force) {
+				this.adb.executeShellCommand(["rm", "-rf", deviceHashService.hashFileDevicePath]).wait();
+				this.adb.executeCommand(["push", projectFilesPath, deviceAppData.deviceProjectRootPath]).wait();
+			} else {
+				// Create or update file hashes on device
+				let oldShasums = deviceHashService.getShasumsFromDevice().wait();
+				if (oldShasums) {
+					let changedShasums = _(currentShasums)
+						.filter(hash => !_.find(oldShasums, h => hash.trim().toLowerCase() === h.trim().toLowerCase()))
```
teobugslayer: As hashes are under our control, can we omit calling `toLowerCase`? Also perhaps we can omit calling `trim`.

teobugslayer: There is no chaining of functions in here. I suggest to use the shorter call

`_.filter(currentShasums, hash => ...);`

mobile/android/android-device-hash-service.ts
```diff
@@ -0,0 +1,126 @@
...
+		if (!this._hashFileDevicePath) {
+			this._hashFileDevicePath = this.$mobileHelper.buildDevicePath(AndroidDeviceHashService.DEVICE_ROOT_PATH, this.appIdentifier, AndroidDeviceHashService.HASH_FILE_NAME);
+		}
+
+		return this._hashFileDevicePath;
+	}
+
+	public doesShasumFileExistsOnDevice(): IFuture<boolean> {
+		return ((): boolean => {
+			let lsResult = this.adb.executeShellCommand(["ls", this.hashFileDevicePath]).wait();
```
teobugslayer: Do we support hidden files? If so, we need "-a" parameter to ls.

rosen-vladimirov: no, no, the idea is to list one file only.
example output of calling ls for existing file is:
```
~ $ ls testInq.js 
testInq.js
```
Output of trying ls of not existing file is:
```
~ $ ls notExisting
ls: cannot access notExisting: No such file or directory
```
The idea is that if the file exists, the output (after trim) will be the same as the file path.

rosen-vladimirov: Also this code is executed on the Android device itself

helpers.ts
rosen-vladimirov: I'm not using this function anymore :)

justcodebuilduser:
:heart:

rosen-vladimirov:
ping @teobugslayer - I've changed hash file to be JSON (check the last commit)

helpers.ts
teobugslayer: Then delete the code?

helpers.ts
```diff
@@ -276,6 +276,20 @@ export function connectEventually(factory: () => net.Socket, handler: (_socket:
...
+		dictionary[key] = value;
+	});
+
+	return dictionary;
+}
+
+export function convertToString(dictionary: IStringDictionary): string {
```
teobugslayer: Also delete this one

dtopuzov:
:+1:

mobile/android/android-device-hash-service.ts
teobugslayer: Oh I see, this code is only for the file containing the hashes.

teobugslayer:
:+1: after deleting the unused code. I do not insist on the other small stylistic comments.

justcodebuilduser:
:heart:

PR #592, opened 2016-01-26
Show device status when listing available devices

rosen-vladimirov:
Add new column - Status.
The values in it will be "Connected" when the device is reachable and "Unreachable" in case it is locked, not trusted, etc.


justcodebuilduser:
:heart:

Mitko-Kerezov:
:+1:

teobugslayer:
:+1:

PR #593, opened 2016-01-28
Use new Analytics SDK and track more information

rosen-vladimirov:
* Use new Analytics SDK - it has some improvements and it is also minified, so we'll reduce the size of the package.
* Each command is considered a new user session, so persist a counter in user settings and increment it when new session is started. The counter has integer type on Analytics side, and we are incrementing it per user, so we should never overflow it.
* When starting the analytics monitor, call `setStartCount` with the new session number. This is required by the new Analytics SDK.
* Add new method that will track feature no matter of the current user settings `trackFeatureCore`.
* Track information for user's answer when asking them if they want to enable feature usage tracking. In analytics, the feature will be shown as `AcceptFeatureUsageTracking` and it will be true when user accepts the prompt and false otherwise.
* Track when `TrackExceptions` or `TrackFeatureUsage` value is change (by using `<cli name> usage-reporting disable/enable`.
* Fix exceptions tracking by:
  * Fix the call to `analyticsService.trackExceptions` method (incorrect parameters, missing `.wait()`) in `installUncaughtExceptionListener`.
  * Track exceptions in `errors.beginCommand` before process exiting.
  * As tracking the exception might take a while, make sure to wait until exception is send to server before exiting the process. In order to achieve this, continiously check the value of isSending property of the monitor. If it's false, the monitor had send the required information.
* Move unit tests for AnalyticsService from AppBuilder CLI to common.

Added new Tests:
* setStatus
  * tracks that user had disabled feature tracking
  * tracks that user had enabled feature tracking
  * tracks that user had disabled exceptions tracking
  * tracks that user had enabled exceptions tracking

* checkConsent
  * sends information that user had accepted feature tracking
  * sends information that user had rejected feature tracking

Also remove the isEverythingDisabled check from analytics start method.
In case featureUsageTracking is disabled and exceptions reporting is disabled,
we might still want to track some information, so make sure eqatecMonitor can start.
Such case is when ExceptionsTracking is disabled, we ask the user if he wants to enable FeatureTracking and he refuses.
All tracking is disabled, but we want to track down that the user refused to enable feature tracking.


justcodebuilduser:
:broken_heart:

justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

errors.ts
```diff
@@ -99,6 +83,26 @@ export function installUncaughtExceptionListener(actionOnException?: () => void)
 	});
 }
 
+function tryTrackException(error: Error, injector: IInjector): void {
+	let disableAnalytics: boolean;
+	try {
```
teobugslayer: I think setting disableAnalytics to a default value belongs to the static config and not here.

rosen-vladimirov: I do not want to set the global value, the idea is to have it set to `true` just here for skipping AnalyticsService. This is usefult only in our unit tests and we should never set it to true in a real use-case.

services/analytics-service.ts
```diff
@@ -34,6 +40,11 @@ export class AnalyticsService implements IAnalyticsService {
 
 					let trackFeatureUsage = this.$prompter.confirm(message, () => true).wait();
 					this.setStatus(this.$staticConfig.TRACK_FEATURE_USAGE_SETTING_NAME, trackFeatureUsage).wait();
+					if(trackFeatureUsage) {
+						this.trackFeatureCore(`${this.acceptTrackFeatureSetting}.true`).wait();
+					} else {
+						this.trackFeatureCore(`${this.acceptTrackFeatureSetting}.false`).wait();
+					}
```
Fatme: You can simplify `if else` with the following
`this.trackFeatureCore(`${this.acceptTrackFeatureSetting}.${!!trackFeatureUsage}`).wait();`

justcodebuilduser:
:heart:

errors.ts
teobugslayer: Doesn't this mean that we have to write our unit tests to mock this value to true and remove this code from here?

rosen-vladimirov: it means that every single unit test must mock staticConfig and set disableAnalytics to true. As in the tests we are using testInjector, we should register the staticConfig in each of them. That's what I'm trying to prevent, but if you insist, I'll handle this in a separate PR :)

teobugslayer:
:+1:

PR #594, opened 2016-02-02
Fix exception tracking message

rosen-vladimirov:
Remove the message of the exception as it is already printed in most of the cases.


teobugslayer:
:+1:

justcodebuilduser:
:heart:

vchimev:
What about Mac users? @rosen-vladimirov

rosen-vladimirov:
Ctrl + C is SIGINT signal and as far as I know, it's the same on Mac. @vchimev

vchimev:
:+1:

PR #602, opened 2016-02-15
Fix error for missing Fiber when tracking uncaught exceptions

rosen-vladimirov:
When tracking uncaught exceptions, we do not have fiber, but our code works with futures. So it fails with error "Can't wait without fiber".
Fix this by creating fiber.


justcodebuilduser:
:heart:

teobugslayer:
:+1:

PR #607, opened 2016-02-26
Fix childProcess arguments printing

rosen-vladimirov:
In childProcess we print the arguments passed to a method. In case it does not have arguments, we our method for printing fails.
Fix the method by checking arguments before printing them.


justcodebuilduser:
:heart:

teobugslayer:
:+1:

PR #608, opened 2016-02-29
Fix sync of platform specific and config specific files

rosen-vladimirov:
Fix LiveSync of platform specific and configuration specific files by extracting logic for parsing filename and use it in NS CLI.
Add ProjectFilesProviderBase class that holds the logic, so it can still be used from ProjectFilesManager.

Add --release option to common options so when `tns livesync <platform> --release --watch` is used, the files will be processed correctly.
Add tests for the new ProjectFilesProviderBase class.

### DIFF
--- a/declarations.d.ts
+++ b/declarations.d.ts
@@ -485,6 +485,7 @@ interface ICommonOptions {
 	template: string;
 	var: Object;
 	default: Boolean;
+	release: boolean;
 }
 
 interface IYargArgv extends IDictionary<any> {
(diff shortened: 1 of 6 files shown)


justcodebuilduser:
:heart:

ErjanGavalji:
:+1:

ligaz:
:+1:

declarations.d.ts
```diff
@@ -766,6 +767,12 @@ interface IProgressIndicator {
 	showProgressIndicator(future: IFuture<any>, timeout: number, options?: { surpressTrailingNewLine?: boolean }): IFuture<void>;
 }
 
+interface IProjectFileInfo {
```
Mitko-Kerezov: Some JSDocs would be nice here too

services/project-files-provider-base.ts
```diff
@@ -0,0 +1,47 @@
...
+		let parsed = this.parseFile(filePath, this.$mobileHelper.platformNames, platform || "");
+		if (!parsed) {
+			parsed = this.parseFile(filePath, ["debug", "release"], this.$options.release ? "release" : "debug");
+		}
+
+		return parsed || {
+			filePath: filePath,
+			onDeviceFileName: path.basename(filePath),
+			shouldIncludeFile: true,
+			value: platform
```
Mitko-Kerezov: Consider removing this `value` property - it seems that it it's not used anywhere

justcodebuilduser:
:heart:

PR #609, opened 2016-03-01
Add option for (dis)allowing an empty string in prompter

Mitko-Kerezov:
Implemented to facilitate calls requiring nonempty strings.
Ping @rosen-vladimirov

### DIFF
--- a/declarations.d.ts
+++ b/declarations.d.ts
@@ -220,10 +220,18 @@ interface IAnalyticsService {
 	track(featureName: string, featureValue: string): IFuture<void>;
 }
 
+interface IAllowEmpty {
+	allowEmpty?: boolean
+}
+
+interface IPrompterOptions extends IAllowEmpty {
+	defaultAction?: () => string
+}
+
 interface IPrompter extends IDisposable {
 	get(schemas: IPromptSchema[]): IFuture<any>;
-	getPassword(prompt: string, options?: {allowEmpty?: boolean}): IFuture<string>;
-	getString(prompt: string, defaultAction?: () => string): IFuture<string>;
+	getPassword(prompt: string, options?: IAllowEmpty): IFuture<string>;
+	getString(prompt: string, options?: IPrompterOptions): IFuture<string>;
 	promptForChoice(promptMessage: string, choices: any[]): IFuture<string>;
 	confirm(prompt: string, defaultAction?: () => boolean): IFuture<boolean>;
 }
--- a/prompter.ts
+++ b/prompter.ts
@@ -66,7 +66,7 @@ export class Prompter implements IPrompter {
 		return future;
 	}
 
-	public getPassword(prompt: string, options?: {allowEmpty?: boolean}): IFuture<string> {
+	public getPassword(prompt: string, options?: IAllowEmpty): IFuture<string> {
 		return (() => {
 			let schema: IPromptSchema = {
 				message: prompt,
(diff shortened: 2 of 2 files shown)


justcodebuilduser:
:heart:

prompter.ts
```diff
@@ -66,7 +66,7 @@ export class Prompter implements IPrompter {
 		return future;
 	}
 
-	public getPassword(prompt: string, options?: {allowEmpty?: boolean}): IFuture<string> {
+	public getPassword(prompt: string, options?: IAllowEmpty): IFuture<string> {
```
rosen-vladimirov: Can't we have default action here? Why not using `IPrompterOptions`?

prompter.ts
```diff
@@ -83,16 +83,20 @@ export class Prompter implements IPrompter {
...
-	public getString(prompt: string, defaultAction?: () => string): IFuture<string> {
+	public getString(prompt: string, options?: IPrompterOptions): IFuture<string> {
 		return (() => {
 			let schema: IPromptSchema = {
 				message: prompt,
 				type: "input",
-				name: "inputString"
+				name: "inputString",
+				validate: (value: any) => {
+					let allowEmpty = options === undefined || options.allowEmpty === undefined ? true : options.allowEmpty;
```
rosen-vladimirov:
```TypeScript
let allowEmpty = !options || options.allowEmpty === undefined || options.allowEmpty;
```

prompter.ts
```diff
@@ -83,16 +83,20 @@ export class Prompter implements IPrompter {
...
+					return (!allowEmpty && !value) ? `${prompt} must be non-empty` : true;
```
rosen-vladimirov: Even better:
```TypeScript
let doesNotAllowEmpty = options && options.allowEmpty;
return (doesNotAllowEmpty && !value) ? `${prompt} must be non-empty` : true;
```

But why the return of validate method is string in one case and boolean in the other?

declarations.d.ts
```diff
@@ -220,10 +220,18 @@ interface IAnalyticsService {
 	track(featureName: string, featureValue: string): IFuture<void>;
 }
 
+interface IAllowEmpty {
+	allowEmpty?: boolean
```
rosen-vladimirov: why is this in a separate interface, can't you add it as an optional property in IPrompterOptions?

prompter.ts
Mitko-Kerezov: I'd rather not have a default action here because I can't imagine a case where one'd want a default action to take if a password is not provided.
If, on the other hand, we use `IPrompterOptions` here and just don't respect the `defaultAction` that would be misleading to the coder.

declarations.d.ts
Mitko-Kerezov: image axd

If we combine those two interfaces into one it would be misleading when using them inside methods which never reference one or more of these interface members

Mitko-Kerezov:
Ping @rosen-vladimirov for a double-chek after fixing the comment

justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

prompter.ts
```diff
@@ -83,16 +83,20 @@ export class Prompter implements IPrompter {
...
+				validate: (value: any) => {
+					let doesNotAllowEmpty = options && _.has(options, "allowEmpty") && !options.allowEmpty;
+					return (doesNotAllowEmpty && !value) ? `${prompt} must be non-empty` : true;
+				}
 			};
 
-			if(defaultAction) {
-				schema.default = defaultAction;
+			if(options && options.defaultAction) {
+				schema.default = options.defaultAction;
```
rosen-vladimirov: maybe we can remove the `if`:
```TypeScript
schema.default = options && options.defaultAction;
```

rosen-vladimirov:
:+1:  after considering my comment for default action

justcodebuilduser:
:heart:

PR #610, opened 2016-03-01
Move common options tests to common

rosen-vladimirov:
Move common options tests to common lib (from icenium-cli), where they belong.
Also modify the $options visibility to protected, so it can be called from classes that extend ProjectFilesProviderBase.
Add -r as shorthand for --release.

### DIFF
--- a/test/unit-tests/common-options.ts
+++ b/test/unit-tests/common-options.ts
@@ -1,0 +1,316 @@
+///<reference path="../.d.ts"/>
+"use strict";
+
+import { Errors } from "../../errors";
+import { Yok } from "../../yok";
+import { assert } from "chai";
+import { OptionType, OptionsBase } from "../../options";
+
+let isExecutionStopped = false;
+
+let knownOpts = {
+	"path1": { type: OptionType.String },
+	"help": { type: OptionType.Boolean } ,
+	"verbose": { type: OptionType.Boolean, alias: "v" },
+	"profileDir": { type: OptionType.String },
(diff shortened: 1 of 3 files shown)


justcodebuilduser:
:broken_heart:

justcodebuilduser:
:heart:

test/unit-tests/common-options.ts
```diff
@@ -0,0 +1,307 @@
+///<reference path="../.d.ts"/>
+"use strict";
+
+let optionsPath = require("../../options");
```
Mitko-Kerezov: You can use ES6 imports here

test/unit-tests/common-options.ts
```diff
@@ -0,0 +1,307 @@
...
+import {Errors} from "../../errors";
+import yok = require("../../yok");
+import {assert} from "chai";
+let optionType = optionsPath.OptionType;
```
Mitko-Kerezov: You can use ES6 imports here also

test/unit-tests/common-options.ts
```diff
@@ -0,0 +1,307 @@
+///<reference path="../.d.ts"/>
+"use strict";
+
+let optionsPath = require("../../options");
+import {Errors} from "../../errors";
+import yok = require("../../yok");
```
Mitko-Kerezov: ES6 imports

Mitko-Kerezov:
:+1: after fixing comments

justcodebuilduser:
:heart:

PR #625, opened 2016-03-31
Add list dependencies

TsvetanMilanov:
Add list plugins registered in package.json file when executing tns plugin in project.
Add plugin find command to list the nativescript plugins available in npm. The command has --all and --count flags to change the output format.


justcodebuilduser:
:heart:

rosen-vladimirov:
:+1:

PR #638, opened 2016-04-14
Enable logging for Proton

rosen-vladimirov:
We've disabled logging for Proton as the electron's fork was not giving our process stdout stream where to write.
However after some changes on Proton side, all output from common-lib is written to file, so let's fill it with some data.


justcodebuilduser:
:heart:

Mitko-Kerezov:
Log all the things! :+1:

PR #624, opened 2016-03-30
 Add warn for new versions

TsvetanMilanov:
When executing tns doctor outside of project if the nativescript cli needs update it will print information for the available update.
If tns doctor is executed in project it will check the installed runtimes versions and the nativescript core modules version and print information if an update is needed.
Add tns info command to display versions information for NativeScript CLI, core modules and runtimes.

PR #642, opened 2016-04-18
Add warn for new versions

TsvetanMilanov:
When executing tns doctor outside of project if the nativescript cli needs update it will print information for the available update.
If tns doctor is executed in project it will check the installed runtimes versions and the nativescript core modules version and print information if an update is needed.
Add tns info command to display versions information for NativeScript CLI, core modules and runtimes.


TsvetanMilanov:
Approved in https://github.com/telerik/mobile-cli-lib/pull/624

justcodebuilduser:
:heart:

PR #646, opened 2016-04-18
Compare versions properly

Mitko-Kerezov:
When using the filter `x => x` the number 0 gets filtered out incorrectly
Ping @rosen-vladimirov


justcodebuilduser:
:heart:

rosen-vladimirov:
:+1:

PR #647, opened 2016-04-18
Fix analytics tracking

rosen-vladimirov:
Do not track setting of feature-usage-tracking and error-reporting when we are asking for confirmation of Feature Usage Tracking.


Mitko-Kerezov:
:+1:

justcodebuilduser:
:heart:

PR #648, opened 2016-04-18
Gather information about xcodeproj ruby gem's location

Mitko-Kerezov:
This gem is a dependency of CocoaPods but users may not want to install or use CocoaPods. However this gem is crucial for performing local builds for the iOS platform.

Ping @rosen-vladimirov for a quick review


justcodebuilduser:
:heart:

rosen-vladimirov:
:+1:

PR #652, opened 2016-04-21
Fix new Android emulator fail to start application

TsvetanMilanov:
After updating the Android default emulator the CLI starts it and cannot deploy the application on it. This is caused by the change in the OK message after the net.connect(port) client connects to the emulator.

Fixes https://github.com/NativeScript/nativescript-cli/issues/1689

### DIFF
--- a/mobile/android/android-emulator-services.ts
+++ b/mobile/android/android-emulator-services.ts
@@ -24,16 +24,16 @@ class AndroidEmulatorServices implements Mobile.IAndroidEmulatorServices {
 	private static RUNNING_ANDROID_EMULATOR_REGEX = /^(emulator-\d+)\s+device$/;
 
 	private static MISSING_SDK_MESSAGE = "The Android SDK is not configured properly. " +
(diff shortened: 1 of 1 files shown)


justcodebuilduser:
:heart:

mobile/android/android-emulator-services.ts
```diff
@@ -223,45 +223,64 @@ class AndroidEmulatorServices implements Mobile.IAndroidEmulatorServices {
...
+
+		let firstIndexOfOk = _.indexOf(lines, "OK");
+
+		if (firstIndexOfOk < 0) {
+			return null;
+		}
+
+		let secondIndexOfOk = _.indexOf(lines, "OK", firstIndexOfOk + 1);
+
+		if (secondIndexOfOk < 0) {
```
rosen-vladimirov: you do not need this check. In case `firstIndexOfOk` is at least 0, adding 1 to it will never result in negative value

rosen-vladimirov:
Great work
:+1:

PR #654, opened 2016-04-22
Add successfully installed message on post install

rosen-vladimirov:
Add successfully installed message on post install. Show it only when the property in static config is set.


Mitko-Kerezov:
:+1:

justcodebuilduser:
:heart:

PR #657, opened 2016-04-26
Fix iOS Device detection 

rosen-vladimirov:
Use sysInfoBase as it's the only one registered in common lib.


justcodebuilduser:
:broken_heart:

Mitko-Kerezov:
:+1: after passing tests

justcodebuilduser:
:heart:

PR #660, opened 2016-04-28
Fix isLiveSyncSupported check for iOS devices on Mac

rosen-vladimirov:
On Mac `receiveMessage` method returns only the last message instead of all. It contains only the Success/Fail status, so we decide that there are no applications that can be LiveSynced. This breaks Proton.


rosen-vladimirov:
👍

justcodebuilduser:
:heart:

PR #661, opened 2016-04-29
Fix livesync for iOS on Mac

rosen-vladimirov:
The event loop is really messed with our setInterval that should detect devices. As it's running on 500ms, when livesync operation is in progress, it constantly fails.
As a workaround, stop the device detection while livesyncing and start it again after that.
Try moving device detection to a separate PR in the bright future.


justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

PR #662, opened 2016-05-10
Fix detection when iOS Sim is closed and reopened

rosen-vladimirov:
In case you do:
 - start simulator
 - close simulator
 - start simulator

We are not reporting it as started the second time, as we are not clearing our cache. Clear the cache and add unit tests.
Use deep clone of the simulator object returned by ios-sim-portable in order to make the code testable.


justcodebuilduser:
:heart:

Mitko-Kerezov:
:+1:

PR #664, opened 2016-05-11
Allow using .zip files for deploy to iOS simulator

rosen-vladimirov:
Allow using .zip files for deploy to iOS simulator - if the passed path is .zip, extract it to temp dir and pass the temp dir to the simulator.


justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

PR #666, opened 2016-05-12
Fix detection of untrusted iOS device

rosen-vladimirov:
When iOS device is not trusted, it will not have productType and our logic for detecting active architecture will fail.
As it is in the constructor of iOSDevice, it breaks the whole logic. Fix it by setting active architecture to empty string when iOS device is not trusted.


justcodebuilduser:
:heart:

TsvetanMilanov:
:+1:

PR #669, opened 2016-05-13
Fix starting many adb processes

rosen-vladimirov:
When new device is attached, common lib is trying to get all of its applications and to send intents for each of them in order to check is livesync supported for them.
This leads to tens or even hundreds of started adb processes. At some point, when you have more than one device, adb cannot handle the whole flow and it suicides itself.
After that you have to kill manually the only existing adb process or it will never detect any devices.
In order to prevent this behavior, change the way of detection is LiveSync supported. Expose new public method that will be used to check if LiveSync is supported for specific devices (and one appIdentifier).
Remove the reporting of isLiveSyncSupported in the "applicationInstalled" event.

### DIFF
--- a/appbuilder/device-emitter.ts
+++ b/appbuilder/device-emitter.ts
@@ -66,14 +66,14 @@ class DeviceEmitter extends EventEmitter {
 	}
 
 	private attachApplicationChangedHandlers(device: Mobile.IDevice): void {
-		device.applicationManager.on("applicationInstalled", (applicationName: string, applicationSettings: {isLiveSyncEnabled: boolean}) => {
-			this.emit("applicationInstalled", device.deviceInfo.identifier, applicationName, applicationSettings);
-			this.checkCompanionAppChanged(device, applicationName, "companionAppInstalled");
+		device.applicationManager.on("applicationInstalled", (appIdentifier: string) => {
+			this.emit("applicationInstalled", device.deviceInfo.identifier, appIdentifier);
+			this.checkCompanionAppChanged(device, appIdentifier, "companionAppInstalled");
 		});
 
-		device.applicationManager.on("applicationUninstalled", (applicationName: string) => {
-			this.emit("applicationUninstalled", device.deviceInfo.identifier, applicationName);
-			this.checkCompanionAppChanged(device, applicationName, "companionAppUninstalled");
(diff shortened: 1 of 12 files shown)


justcodebuilduser:
:heart:

mobile/application-manager-base.ts
```diff
@@ -44,13 +42,7 @@ export abstract class ApplicationManagerBase extends EventEmitter implements Mob
...
 					let removedAppIdentifiers = _.difference(previouslyInstalledAppIdentifiers, currentlyInstalledAppIdentifiers);
 
-					if(newAppIdentifiers.length || removedAppIdentifiers.length) {
-						this.applicationsLiveSyncStatus = _.reject(this.applicationsLiveSyncStatus,
-															(app: Mobile.IApplicationLiveSyncStatus) => _.contains(removedAppIdentifiers, app.applicationIdentifier));
-						this.getApplicationsLiveSyncSupportedStatus(newAppIdentifiers).wait();
-					}
-
-					_.each(newAppIdentifiers, application => this.emit("applicationInstalled", application, {isLiveSyncSupported: this.isLiveSyncSupported(application).wait()} ));
+					_.each(newAppIdentifiers, application => this.emit("applicationInstalled", application));
```
Mitko-Kerezov: I'd rather you renamed the captured `application` argument to something like `applicationIdentifier` or `appIdentifier`

TsvetanMilanov:
:+1:

appbuilder/device-emitter.ts
```diff
@@ -66,8 +66,8 @@ class DeviceEmitter extends EventEmitter {
 	}
 
 	private attachApplicationChangedHandlers(device: Mobile.IDevice): void {
-		device.applicationManager.on("applicationInstalled", (applicationName: string, applicationSettings: {isLiveSyncEnabled: boolean}) => {
-			this.emit("applicationInstalled", device.deviceInfo.identifier, applicationName, applicationSettings);
+		device.applicationManager.on("applicationInstalled", (applicationName: string) => {
```
Mitko-Kerezov: I'd rather you renamed `applicationName` to `applicationIdentifier` so it's not as misleading

Mitko-Kerezov:
:+1:

justcodebuilduser:
:heart:

PR #670, opened 2016-05-13
Fix LiveSync to companion app

rosen-vladimirov:
Currently LiveSync to companion app via Proton fails with error that application is not installed. In fact it fails because we've not changed the appIdentifier to be the companion app's identifier.


TsvetanMilanov:
:+1:

justcodebuilduser:
:heart:

PR #671, opened 2016-05-13
Disable iOS device detection interval on OS X

rosen-vladimirov:
On Mac OS X, there's no need to call iOSDeviceDetection process in setInterval, as when runLoopRun is called once (first time before the setInterval), after that it starts calling our callback whenever a device is changed (attached/detached).
So ignore the code on Mac.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,9 +7,9 @@ Contains common infrastructure for CLIs - mainly AppBuilder and NativeScript.
 Installation
 ===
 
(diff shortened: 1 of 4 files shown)


TsvetanMilanov:
:+1:

justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

PR #672, opened 2016-05-14
Fix Android device detection when adb is not in the PATH

rosen-vladimirov:
When adb is not in the PATH, common lib is trying to spawn it's private adb, which leads to recursion errors.
Make sure to spawn our own adb in this cases.
Also prevent shelljs cp errors by skipping new copy of files in case they are already there.
Prevent shelljs.chmod errors on Windows - we do not need to chmod on Windows.

### Update adb
Use latest official adb version from Android Platform Tools 23.1
Android Debug Bridge version 1.0.32
Revision 09a0d98bebce-android


justcodebuilduser:
:heart:

Mitko-Kerezov:
:+1:

PR #674, opened 2016-05-14
Fix add android platform dependencies

TsvetanMilanov:
When checking if platform dependency is installed in the current project before adding the platform need to check if the current version is valid or check in the node_modules directory for the installed plugin version.


TsvetanMilanov:
Close and retarget to release.

justcodebuilduser:
:heart:

PR #677, opened 2016-05-16
Ignore error code 21 for iOS Devices

rosen-vladimirov:
It looks like this error does not prevent us from working correctly. Just log it as warning.


justcodebuilduser:
:heart:

TsvetanMilanov:
:+1:

PR #678, opened 2016-05-16
Do not set iOS Status to UNREACHABLE in all cases when smth fails

rosen-vladimirov:
It looks like that even when some operation fails, we are still able to execute iOS Device related operations.
So do not set the status to UNREACHABLE in case some of the operations that are not connect (or pairing) fails.
Set the status to CONNECTED on every successful operation.


TsvetanMilanov:
:+1:

justcodebuilduser:
:heart:

PR #680, opened 2016-05-17
Use installationProxy to get installed applications on iOS

rosen-vladimirov:
Currently we are using AMDeviceLookupApplications to check the installed applications on iOS.
As we need information is LiveSync supported for all user's applications, we have to make another call through installationProxy, that returns both CFBundleIdentifier and IsLiveSyncSupported property.
So we do not need the first call at all as the second one gives us all the required information.
Also cache the installation proxy instead of creating new one each time.

Use the cached data for isApplicationInstalled method. In case the app is not there, applicationInstalled will be raised when it is installed.
Cache the Installation Plist service, so we'll have only one instance per each iOS device.


justcodebuilduser:
:heart:

justcodebuilduser:
:broken_heart:

justcodebuilduser:
:heart:

PR #682, opened 2016-05-20
Remove caching of iOS services

rosen-vladimirov:
### Remove caching of iOS services
Revert caching of iOS services as this caused several issues with multiple data parsing, keeping sockets opened, etc.
Make sure to close the socket after each operation.

### Get currently installed applications when checking is app installed
In case you call isApplicationInstalled before checkForApplicationUpdates, the cache will not be populated and the result will always be false.
In case there's no data in the cache, call checkForApplicationUpdates.

### DIFF
--- a/mobile/ios/device/ios-proxy-services.ts
+++ b/mobile/ios/device/ios-proxy-services.ts
@@ -224,7 +224,6 @@ export class AfcClient implements Mobile.IAfcClient {
 
 export class InstallationProxyClient {
 	private plistService: Mobile.IiOSDeviceSocket = null;
-	private plistInstallationService: iOSCore.PlistService;
 
 	constructor(private device: Mobile.IiOSDevice,
 		private $logger: ILogger,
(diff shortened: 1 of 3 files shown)


justcodebuilduser:
:heart:

mobile/ios/device/ios-proxy-services.ts
```diff
@@ -266,7 +266,11 @@ export class InstallationProxyClient {
 	}
 
 	public closeSocket() {
-		return this.plistService.close();
+		if (this.plistService) {
```
Mitko-Kerezov: shouldn't this method close both sockets, if they exist, instead of prioritizing

justcodebuilduser:
:heart:

mobile/ios/device/ios-proxy-services.ts
rosen-vladimirov: Fixed :)

Mitko-Kerezov:
:+1:

PR #683, opened 2016-05-20
Fix livesync when app is not installed

rosen-vladimirov:
Fix livesync of application when it is not installed on device. Currently it is running two builds for Android.


TsvetanMilanov:
:+1:

justcodebuilduser:
:heart:

PR #686, opened 2016-05-23
iOS Device operations fixes

rosen-vladimirov:
### 	Retry iOS device file related operations
Most of the iOS File related operations may fail on Mac OS X, but after retrying them, they succeed.
Retry the operations several times and fail in case they do not succeed.
Also fix error: "Futures left behind" during transfer of files when `ensureDevicePathExist` throws (the future had not been resolved).

### Set iOS Device status to Unreachable only in constructor
Set iOS Device status to Unreachable only in constructor, not when some operation fails.
Most of the operations may fail a single time only, so the device is not Unreachable.
When device is trusted, we will remove current instance of iOS Device and we'll create new one, where the status will be Connected.

### DIFF
--- a/mobile/ios/device/ios-proxy-services.ts
+++ b/mobile/ios/device/ios-proxy-services.ts
@@ -17,6 +17,20 @@ export class MobileServices {
 	public static DEBUG_SERVER: string = "com.apple.debugserver";
 }
 
+const NUMBER_OF_RETRIES = 5;
+function executeActionUntilSuccess(action: Function) {
+	let result: number;
+	for (let currentTry = 0; currentTry < NUMBER_OF_RETRIES && result !== 0; currentTry++) {
+		try {
+			result = action();
+		} catch (err) {
+			this.$logger.trace(`Error #${currentTry} while trying to execute action. Error is: `, err);
+		}
+	}
+
+	return result;
+}
+
 export class AfcFile implements Mobile.IAfcFile {
 	private open: boolean = false;
 	private afcFile: number;
@@ -25,7 +39,8 @@ export class AfcFile implements Mobile.IAfcFile {
 		mode: string,
 		private afcConnection: NodeBuffer,
 		private $mobileDevice: Mobile.IMobileDevice,
-		private $errors: IErrors) {
+		private $errors: IErrors,
+		private $logger: ILogger) {
 		let modeValue = 0;
 		if (mode.indexOf("r") > -1) {
 			modeValue = 0x1;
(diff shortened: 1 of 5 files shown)


justcodebuilduser:
:heart:

mobile/ios/device/ios-proxy-services.ts
```diff
@@ -17,6 +17,20 @@ export class MobileServices {
 	public static DEBUG_SERVER: string = "com.apple.debugserver";
 }
 
+const NUMBER_OF_RETRIES = 5;
+function executeActionUntilSuccess(action: Function) {
```
TsvetanMilanov: Maybe create base Afc class which has protected executeActionUntilSuccess(action: Function): number

mobile/ios/device/ios-proxy-services.ts
```diff
@@ -17,6 +17,20 @@ export class MobileServices {
 	public static DEBUG_SERVER: string = "com.apple.debugserver";
 }
 
+const NUMBER_OF_RETRIES = 5;
```
TsvetanMilanov: Number of retries based on what? Is it fast to retry 5 times?

mobile/ios/device/ios-proxy-services.ts
rosen-vladimirov: great idea

justcodebuilduser:
:heart:

mobile/ios/device/ios-proxy-services.ts
```diff
@@ -109,15 +141,16 @@ export class AfcClient implements Mobile.IAfcClient {
 	}
 
 	public mkdir(path: string) {
-		let result = this.$mobileDevice.afcDirectoryCreate(this.afcConnection, path);
+		let result = this.tryExecuteAfcAction(() => this.$mobileDevice.afcDirectoryCreate(this.afcConnection, path));
+
 		if (result !== 0) {
-			this.$errors.fail("Unable to make directory: %s. Result is %s", path, result);
+			this.$logger.warn(`Unable to make directory: ${path}. Result is ${result}.`);
```
rosen-vladimirov: this should be `$errors.fail`

justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

mobile/ios/device/ios-proxy-services.ts
```diff
@@ -85,16 +115,18 @@ export class AfcFile implements Mobile.IAfcFile {
...
 	private afcConnection: NodeBuffer = null;
 
 	constructor(private service: number,
 		private $mobileDevice: Mobile.IMobileDevice,
 		private $coreFoundation: Mobile.ICoreFoundation,
 		private $fs: IFileSystem,
 		private $errors: IErrors,
-		private $logger: ILogger,
+		protected $logger: ILogger,
 		private $injector: IInjector) {
```
TsvetanMilanov: Can you move this above the `$logger`?

TsvetanMilanov:
:+1:

justcodebuilduser:
:heart:

PR #688, opened 2016-05-25
Kill adb logcat process when parent process dies

rosen-vladimirov:
`adb logcat` remains live on Mac OS X when common-lib process dies. Kill it in such cases, so we'll not have orphans when common-lib is used in Proton.


TsvetanMilanov:
:+1:

justcodebuilduser:
:heart:

PR #690, opened 2016-05-26
Pass CLI Name to verifyNodeVersion

rosen-vladimirov:
Update message for Node.js 0.10.x deprecation and add additional arguments to verification method - CLI's name and version in which support for Node.js 0.10.x will be removed.


justcodebuilduser:
:heart:

TsvetanMilanov:
:+1:

PR #692, opened 2016-05-30
Exclude explicitly changes in App_Resources folder and warn the user.

enchev:
Related to https://github.com/NativeScript/nativescript-cli/issues/1773

### DIFF
--- a/services/livesync-service-base.ts
+++ b/services/livesync-service-base.ts
@@ -7,6 +7,7 @@ import * as shell from "shelljs";
 import * as path from "path";
 import * as temp from "temp";
 import * as minimatch from "minimatch";
+import * as constants from "../../constants";
 let gaze = require("gaze");
 
 class LiveSyncServiceBase implements ILiveSyncServiceBase {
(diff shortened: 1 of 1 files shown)


justcodebuilduser:
:broken_heart:

justcodebuilduser:
:broken_heart:

services/livesync-service-base.ts
```diff
@@ -7,6 +7,7 @@ import * as shell from "shelljs";
 import * as path from "path";
 import * as temp from "temp";
 import * as minimatch from "minimatch";
+import * as constants from "../../constants";
```
rosen-vladimirov: this file does not exist in common lib, so you cannot use it

rosen-vladimirov: you can move the constant that you need in mobile/constants.ts file of common lib :)

enchev: Why our constants are not common? Looks a lot common for me! :)

rosen-vladimirov: because constants like TNS_IOS_RUNTIME_NAME, TNS_ANDROID_RUNTIME_NAME, etc. are not valuable for the common lib. They are usable only for NativeScript CLI

https://github.com/NativeScript/nativescript-cli/blob/master/lib/constants.ts#L12

PR #695, opened 2016-05-31
Start application on isLivesSyncSupported call from Proton

rosen-vladimirov:
LiveSync plugin requires application to be started on device at least once in order to report that LiveSync for current app is enabled.
The reason is that when the application is started for first time, LiveSync plugin bootstraps the app, by replacing some files. In case we send our newly changed files before that and start the application, LiveSync plugin will replace our files with the original ones from .apk.
So for Proton, start the application when `isLiveSyncSupported` called. The workflow is:
* Call `Build and Deploy` from Proton
* Install the app
* common-lib raises `appInstalled` event
* Proton catches `appInstalled` and calls `isLiveSyncSupported` for the specific devices and application.
* common-lib starts the application and after that calls isLiveSyncSupported.

The other case is when Proton is started and project is opened. In this case Proton automatically calls `isLiveSyncSupported` for current app, so we'll immediately start it on devices (in case it is installed).

### BREAKING CHANGE
Add new parameter to `isLiveSyncSupported` method  - framework. Valid values are cordova and nativescript.

### DIFF
--- a/appbuilder/project/project-base.ts
+++ b/appbuilder/project/project-base.ts
@@ -19,7 +19,7 @@ export class Project implements Project.IProjectBase {
 	public get projectData(): Project.IData {
 		if(this.projectDir) {
 			let projectFile = path.join(this.projectDir, this.$projectConstants.PROJECT_FILE);
-			let jsonContent = this.$fs.readJson(projectFile).wait();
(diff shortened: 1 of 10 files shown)


justcodebuilduser:
:broken_heart:

justcodebuilduser:
:heart:

Mitko-Kerezov:
👍

PR #696, opened 2016-06-01
Remove isLiveSyncSupported from public API

rosen-vladimirov:
### BREAKING CHANGE
Remove isLiveSyncSupported method as isAppInstalledOnDevices returns the same information, so it should be used.

Also from now on, when `isAppInstalledOnDevices` is called, it will start the application, if possible.

### 	Add isCompanionAppInstalledOnDevices method to public API

Add new method to check if companion app is installed on devices.
`isCompanionAppInstalledOnDevices(deviceIdentifiers: string[], framework: string): Promise<IAppInstalledInfo>[]`

The difference from `isAppInstalledOnDevices` is that this method will not start the application on device.

### Set version to 0.12.0
Set version to 0.12.0

### DIFF
--- a/README.md
+++ b/README.md
@@ -422,6 +422,35 @@ Sample result will be:
 }]
 ```
 
+* `isCompanionAppInstalledOnDevices(deviceIdentifiers: string[], framework: string): Promise<IAppInstalledInfo>[]` - checks if the companion application is installed on each of the specified devices and is LiveSync supported for this application.
+The returned type for each device is `IAppInstalledInfo` (check above for full description of the interface).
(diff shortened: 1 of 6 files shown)


rosen-vladimirov:
As requested by @tsvetie

justcodebuilduser:
:heart:

justcodebuilduser:
:broken_heart:

justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

TsvetanMilanov:
:+1:

PR #697, opened 2016-06-01
Add public API for deleting files during liveSync operation

rosen-vladimirov:
Add public method in liveSyncService that should be used to delete files when LiveSync operation is in progress.
This way, when user deletes file in Proton, the application will call `liveSyncService.deleteFiles` method for it and it will be removed from device.

### DIFF
--- a/appbuilder/services/livesync/android-livesync-service.ts
+++ b/appbuilder/services/livesync/android-livesync-service.ts
@@ -2,7 +2,8 @@
 "use strict";
 
 import { AndroidLiveSyncService } from "../../../mobile/android/android-livesync-service";
-import Future = require("fibers/future");
+import * as path from "path";
+import * as helpers from "../../../helpers";
 
 export class AppBuilderAndroidLiveSyncService extends AndroidLiveSyncService implements IPlatformLiveSyncService {
 	constructor(private _device: Mobile.IAndroidDevice,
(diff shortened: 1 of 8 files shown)


justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

appbuilder/services/livesync/android-livesync-service.ts
```diff
@@ -25,8 +26,17 @@ export class AppBuilderAndroidLiveSyncService extends AndroidLiveSyncService imp
 		}).future<void>()();
 	}
 
-	public removeFiles(): IFuture<void> {
-		return Future.fromResult();
+	public removeFiles(appIdentifier: string, localToDevicePaths: Mobile.ILocalToDevicePathData[], deviceProjectRootPath?: string): IFuture<void> {
+		return (() => {
+			let commands =_.map(localToDevicePaths, ldp => {
+				let relativePath = path.relative(deviceProjectRootPath, ldp.getDevicePath()),
```
TsvetanMilanov: `deviceProjectRootPath` is optional and `path.relative` will throw an exception if it is `undefined`.

TsvetanMilanov:
:+1:

justcodebuilduser:
:heart:

PR #700, opened 2016-06-03
Fix device run command

TsvetanMilanov:
When executing `appbuilder device run appid` outside of project the CLI cannot determine which startPackageActivity name to use - for NativeScript or for Cordova application. The fix is to check where the command is executed and if it is not in project we need to run the adb command for both NativeScript and Cordova projects.

### DIFF
--- a/mobile/android/android-application-manager.ts
+++ b/mobile/android/android-application-manager.ts
@@ -2,7 +2,7 @@
 "use strict";
 import {EOL} from "os";
 import {ApplicationManagerBase} from "../application-manager-base";
-import { LiveSyncConstants } from "../../mobile/constants";
+import { LiveSyncConstants, StartPackageActivityNames } from "../../mobile/constants";
 
 export class AndroidApplicationManager extends ApplicationManagerBase {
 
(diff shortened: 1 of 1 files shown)


justcodebuilduser:
:heart:

mobile/android/android-application-manager.ts
```diff
@@ -34,15 +35,27 @@ export class AndroidApplicationManager extends ApplicationManagerBase {
...
+		return this.adb.executeShellCommand(["pm", "uninstall", `${appIdentifier}`], { treatErrorsAsWarnings: true });
 	}
 
 	public startApplication(appIdentifier: string): IFuture<void> {
 		return (() => {
-			this.adb.executeShellCommand(["am", "start",
-				"-a", "android.intent.action.MAIN",
-				"-n", `${appIdentifier}/${this.$staticConfig.START_PACKAGE_ACTIVITY_NAME}`,
-				"-c", "android.intent.category.LAUNCHER"]).wait();
+			if (this.$project.projectData) {
```
rosen-vladimirov: in NativeScript we do not have `$project`

rosen-vladimirov:
let's apply the fix in the command only (`device|run`) - I imagine something like:
```TypeScript
if (<there's a project>) { 
   <call the original method>
} else { 
   <call the start method with framework Cordova>
   < call the start method with framework NativeScript>
}
```

as the code is changed a lot in master and it will be easier to implement it there, let's move this fix for master branch

TsvetanMilanov:
Close and retarget to master.

PR #701, opened 2016-06-03
Add post action on exportedPromise decorator

rosen-vladimirov:
Add postAction on exportedPromise decorator. This action will be called when all Promises are fulfilled.
Stop deviceDetection interval when starting deployOnDevices as it causes issues when deploying on iPhones from Mac.
In the post action start the device detection interval.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,9 +7,9 @@ Contains common infrastructure for CLIs - mainly AppBuilder and NativeScript.
 Installation
 ===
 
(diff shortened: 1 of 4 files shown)


rosen-vladimirov:
👍  after green build :)

justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

PR #703, opened 2016-06-04
Stop device detection and checking for application updates

rosen-vladimirov:
When you have deploy operation or livesync operation, we must stop device detection and checking for application changes as it breaks the deployment on iOS Devices.
The problem is that in case ANY async callback is called during uploading .ipa file to iOS Device for installation, the installation fails with different errors (PackageExtractionFailed, PackageVerificationFailed, etc.).
Add new future inside the setInterval and in case it's not resolved, the next time when setInterval is called, it will see that there's a future pending and will not schedule new execution.
Also when stopDeviceDetectionInterval is called and the future from setInterval is still pending, wait for it before starting the real operation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,9 +7,9 @@ Contains common infrastructure for CLIs - mainly AppBuilder and NativeScript.
 Installation
 ===
 
-Latest version: 0.13.1
+Latest version: 0.13.2
 
-Release date: 2016, June 3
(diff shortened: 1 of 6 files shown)


justcodebuilduser:
:broken_heart:

justcodebuilduser:
:heart:

PR #705, opened 2016-06-07
No more "use strict" and reference ".d.ts", tsc with tsconfig.json should suffice

PanayotCankov:


justcodebuilduser:
:heart:

rosen-vladimirov:
@PanayotCankov can you describe why have you changed `messages.d.ts` to `messages.interface.d.ts`, so we'll not lose the knowledge about this change.

PanayotCankov:
@rosen-vladimirov 

The `tsconfig.json` without `files` field ignores `<name>.d.ts` files if there is a `<name>.ts` file near it. 

The `messages.d.ts` has an interface that is used in `messages.ts`. Before it used to work as both of them were referenced in the `.d.ts` using `/// <reference path="..." />` comments.

Now as the `.d.ts` is removed and we do not want to explicitly list all files in the `tsconfig.json`, I renamed the `messages.d.ts` to `messages.interface.d.ts` to avoid the collision with the `messages.ts` so they both get included in the typescript project by default.

[And the filtering is performed here in tsc](https://github.com/Microsoft/TypeScript/blob/master/src/compiler/commandLineParser.ts#L741).
There is a good reason about it. It is not a bug.

justcodebuilduser:
:heart:

PR #707, opened 2016-06-08
Fix grep command arguments in android process service

TsvetanMilanov:
Remove the -a and --text from all grep commands in the android-process-service because they are not supported on every system.


justcodebuilduser:
:heart:

TsvetanMilanov:
Closed because this bug is fixed in https://github.com/telerik/mobile-cli-lib/pull/708

PR #716, opened 2016-06-17
Do not display all files when livesync is executed

TsvetanMilanov:
When the livesync command is executed we do not need to log info for each file that is transfered. The information should be displayed when the --watch flag is set and some file is changed.

### DIFF
--- a/services/livesync-service-base.ts
+++ b/services/livesync-service-base.ts
@@ -10,6 +10,7 @@ import * as util from "util";
 let gaze = require("gaze");
 
 class LiveSyncServiceBase implements ILiveSyncServiceBase {
+	private showFullLiveSyncInformation: boolean = false;
 	private fileHashes: IDictionary<string>;
 
 	constructor(protected $devicesService: Mobile.IDevicesService,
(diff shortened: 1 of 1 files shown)


services/livesync-service-base.ts
```diff
@@ -11,6 +11,7 @@ let gaze = require("gaze");
 
 class LiveSyncServiceBase implements ILiveSyncServiceBase {
 	private fileHashes: IDictionary<string>;
+	private showFullLiveSyncInformation: boolean;
```
rosen-vladimirov: just set it to false here:
```TypeScript
private showFullLiveSyncInformation: boolean = false;
```
And skip the change in the constructor.

justcodebuilduser:
:broken_heart:

justcodebuilduser:
:heart:

services/livesync-service-base.ts
```diff
@@ -245,7 +247,7 @@ class LiveSyncServiceBase implements ILiveSyncServiceBase {
 
 	private transferFiles(deviceAppData: Mobile.IDeviceAppData, localToDevicePaths: Mobile.ILocalToDevicePathData[], projectFilesPath: string, isFullSync: boolean): IFuture<void> {
 		return (() => {
-			this.logFilesSyncInformation(localToDevicePaths, "Transferring %s.");
+			this.logFilesSyncInformation(localToDevicePaths, "Transferring %s.", this.$logger.trace);
```
rosen-vladimirov: Please add 
```TypeScript
this.$logger.info("Transferring project files...");
```
before this call, so the users will know what's going on.

justcodebuilduser:
:heart:

rosen-vladimirov:
:+1:  after green build

justcodebuilduser:
:heart:

PR #720, opened 2016-06-20
Throw correct error during LiveSync if file does not exist

rosen-vladimirov:
Check if file exists before executing LiveSync operation and throw correct error in case any of the file that should be synced is missing.


justcodebuilduser:
:heart:

TsvetanMilanov:
:+1:

Mitko-Kerezov:
Seems legit 👍

PR #722, opened 2016-06-20
Fix error when converting to boolean

rosen-vladimirov:
If passed value is object or number, toBoolean helper method fails as they do not have .toLowerCase() method.


justcodebuilduser:
:heart:

TsvetanMilanov:
:+1:

justcodebuilduser:
:heart:

PR #724, opened 2016-06-21
Fix LiveSync in AppBuilder and Proton

rosen-vladimirov:
Fix LiveSync in AppBuilder and Proton by calling correct method in livesyncProvider.


justcodebuilduser:
:heart:

TsvetanMilanov:
:+1:

PR #726, opened 2016-06-23
Delete unzip.d.ts

rosen-vladimirov:
Delete unzip.d.ts as unzip module had been used in AppBuilder CLI only, but we've decided to drop it as it's not updated for the last two years.
When using unzip module with Node.js 6.x.x a warning is printed:
```
(node:13828) fs: re-evaluating native module sources is not supported. If you are using the graceful-fs module, please update it to a more recent version.
```


justcodebuilduser:
:heart:

TsvetanMilanov:
:+1:

PR #727, opened 2016-06-24
Make mobile-cli-lib usable with node

rosen-vladimirov:
Update dependencies of mobile-cli-lib, so when it is used as npm package, it can be used directly with Node.js. Until now, the C++ dependencies had been rebuilt for Electron, so they could not run with Node.js.
When used with Electron on Mac, the iOS related operations were leaking the callback for detected devices, so it's been called without need to add it in setInterval. In Node.js there's no such leak, so in order to get currently available devices, we have to execute the function in setInterval.


justcodebuilduser:
:heart:

TsvetanMilanov:
:+1:

PR #735, opened 2016-06-29
Set version to 0.16.0

rosen-vladimirov:
Set version to 0.16.0
Ignore  .vscode dir and .npmignore file - we do not need them in the npm package.


TsvetanMilanov:
(Y)

justcodebuilduser:
:heart:

PR #737, opened 2016-06-30
Remove title property from debuggable apps info

rosen-vladimirov:
When information about debuggable apps is returned, the title property does not give any value. It's not real app title, as when the page is changed in WebView, it will be changed as well. So it is removed now.
Also rename the interface and extract new one, that can be used whenever device identifier and application identifier should be used.


justcodebuilduser:
:heart:

Mitko-Kerezov:
👍

PR #742, opened 2016-07-06
Remove logic about storing configuration-specific keys

Mitko-Kerezov:
Instead of caching the old keys, do a deep diff between configuration-specific data and current project data.


justcodebuilduser:
:heart:

TsvetanMilanov:
:+1:

PR #749, opened 2016-07-11
Fix device run command

rosen-vladimirov:
Fix `tns device run` command by removing `$projectConstants` dependency, which is not available in NativeScirpt CLI. It has not been used in the command, only declared in the constructor.


TsvetanMilanov:
:+1:

Mitko-Kerezov:
👍

justcodebuilduser:
:heart:

PR #750, opened 2016-07-11
Fix maximum call stack size exceeded when reading project data 

rosen-vladimirov:
Fix maximum call stack size exceeded when reading project data. Clone the cached object, not the one that calls the same method.


TsvetanMilanov:
:+1:

justcodebuilduser:
:heart:

PR #752, opened 2016-07-12
Fix debug operations for Proton

rosen-vladimirov:
Fix mapAbstractToTcpPort, which was incorrectly reading device files instead of getting them from local machine.
`mapAbstractToTcpPort` must return the same port in case there's already forwarded port for this application - call `adb -s <device id> forward --list` and parse result in order to return the already created one.
Fix returning multiple `chrome` apps when getting debuggable applications. In fact they are all the same, so we can just return the unique values.
Fix getting PIDs of processes, which was returning incorrect results when we have apps with similar names (com.telerik.Debugger and com.telerikDebuggerTests).

In order to get free port on local machine, introduce `$net` dependency, which is wrapper for Node.js's net module. Use OS's logic that prevents EADDRINUSE error - tell the OS that we need free port (pass 0) and it will return the first free one.

### DIFF
--- a/mobile/mobile-core/android-process-service.ts
+++ b/mobile/mobile-core/android-process-service.ts
@@ -8,7 +8,8 @@ export class AndroidProcessService implements Mobile.IAndroidProcessService {
 	constructor(private $errors: IErrors,
 		private $staticConfig: Config.IStaticConfig,
 		private $injector: IInjector,
-		private $httpClient: Server.IHttpClient) {
+		private $httpClient: Server.IHttpClient,
+		private $net: INet) {
 		this._devicesAdbs = {};
 	}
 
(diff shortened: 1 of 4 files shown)


mobile/mobile-core/android-process-service.ts
```diff
@@ -8,7 +8,8 @@ export class AndroidProcessService implements Mobile.IAndroidProcessService {
 	constructor(private $errors: IErrors,
 		private $staticConfig: Config.IStaticConfig,
 		private $injector: IInjector,
-		private $httpClient: Server.IHttpClient) {
+		private $httpClient: Server.IHttpClient,
```
TsvetanMilanov: We don't need this $httpClient.

justcodebuilduser:
:heart:

mobile/mobile-core/android-process-service.ts
rosen-vladimirov: we'll need it soon, so I'll keep it :)

TsvetanMilanov:
:+1:

PR #755, opened 2016-07-14
Implement new events for debuggable WebViews detection

rosen-vladimirov:
Implement `debuggableViewFound`, `debuggableViewLost` and `debuggableViewChanged` events that will be raised when there's a change in application's debuggable views.

### DIFF
--- a/mobile/android/android-application-manager.ts
+++ b/mobile/android/android-application-manager.ts
@@ -11,6 +11,7 @@ export class AndroidApplicationManager extends ApplicationManagerBase {
 		private $options: ICommonOptions,
 		private $logcatHelper: Mobile.ILogcatHelper,
 		private $androidProcessService: Mobile.IAndroidProcessService,
+		private $httpClient: Server.IHttpClient,
 		$logger: ILogger) {
 		super($logger);
 	}
@@ -81,6 +82,30 @@ export class AndroidApplicationManager extends ApplicationManagerBase {
 
 	public getDebuggableApps(): IFuture<Mobile.IDeviceApplicationInformation[]> {
 		return this.$androidProcessService.getDebuggableApps(this.identifier);
+	}
+
+	public getDebuggableAppViews(appIdentifiers: string[]): IFuture<IDictionary<Mobile.IDebugWebViewInfo[]>> {
+		return ((): IDictionary<Mobile.IDebugWebViewInfo[]> => {
+			let mappedAppIdentifierPorts = this.$androidProcessService.getMappedAbstractToTcpPorts(this.identifier, appIdentifiers).wait(),
+				applicationViews: IDictionary<Mobile.IDebugWebViewInfo[]> = {};
+
+			_.each(mappedAppIdentifierPorts, (port: number, appIdentifier: string) => {
+				applicationViews[appIdentifier] = [];
+				let localAddress = `http://127.0.0.1:${port}/json`;
+
+				try {
+					if (port) {
+						let apps = this.$httpClient.httpRequest(localAddress).wait().body;
+						applicationViews[appIdentifier] = JSON.parse(apps);
+					}
+				} catch (err) {
+					this.$logger.trace(`Error while checking ${localAddress}. Error is: ${err.message}`);
+				}
+			});
+
+			return applicationViews;
+
+		}).future<IDictionary<Mobile.IDebugWebViewInfo[]>>()();
 	}
 
 	private getStartPackageActivity(framework?: string): string {
--- a/mobile/application-manager-base.ts
+++ b/mobile/application-manager-base.ts
@@ -1,8 +1,10 @@
 import { EventEmitter } from "events";
+import {TARGET_FRAMEWORK_IDENTIFIERS} from "../constants";
 
 export abstract class ApplicationManagerBase extends EventEmitter implements Mobile.IDeviceApplicationManager {
 	private lastInstalledAppIdentifiers: string[];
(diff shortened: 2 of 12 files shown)


justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

mobile/application-manager-base.ts
```diff
@@ -94,7 +97,43 @@ export abstract class ApplicationManagerBase extends EventEmitter implements Mob
...
+
+			let currentlyAvailableAppViews = this.getDebuggableAppViews(cordovaDebuggableAppIdentifiers).wait();
+
+			_.each(currentlyAvailableAppViews, (currentlyAvailableViews, appIdentifier) => {
+				let previouslyAvailableViews = this.lastAvailableDebuggableAppViews[appIdentifier];
+
+				let newAvailableViews = _.differenceBy(currentlyAvailableViews, previouslyAvailableViews, "id");
+				let notAvailableViews = _.differenceBy(previouslyAvailableViews, currentlyAvailableViews, "id");
+
+				_.each(notAvailableViews, debugWebViewInfo => this.emit("debuggableViewLost", appIdentifier, debugWebViewInfo));
```
Mitko-Kerezov: I'd prefer if this lambda expression did not return the result of `this.emit` because we might get unexpected results in the event of `this.emit` returning `false` or `undefined`

Mitko-Kerezov: This goes for the line below too

mobile/mobile-core/android-process-service.ts
```diff
@@ -51,46 +51,76 @@ export class AndroidProcessService implements Mobile.IAndroidProcessService {
...
 		return ((): Mobile.IDeviceApplicationInformation[] => {
 			let adb = this.getAdb(deviceIdentifier);
 			let androidWebViewPortInformation = (<string>this.getAbstractPortsInformation(adb).wait()).split(EOL);
 
 			// TODO: Add tests and make sure only unique names are returned.
 			return _(androidWebViewPortInformation)
-				.map((line: string) => this.getApplicationInfoFromWebViewPortInformation(adb, deviceIdentifier, line).wait())
-				.filter(appIdentifier => !!appIdentifier)
-				.uniqBy("appIdentifier")
+				.map((line: string) => {
```
Mitko-Kerezov: You probably don't need the `( )` brackets and you definitely don't need the `{ }` ones

mobile/mobile-core/android-process-service.ts
```diff
@@ -174,21 +214,31 @@ export class AndroidProcessService implements Mobile.IAndroidProcessService {
...
 
-	private parseMultilineResult(input: string, regex: RegExp): number {
-		let selectedNumber: number;
+	private getApplicationIdentifierFromPid(adb: Mobile.IDeviceAndroidDebugBridge, pid: string, psData?: string): IFuture<string> {
+		return ((): string => {
+			psData = psData || adb.executeShellCommand(["ps"]).wait();
+			// Process information will look like this (without the columns names):
+			// USER     PID   PPID  VSIZE   RSS   WCHAN    PC         NAME
+			// u0_a63   25512 1334  1519560 96040 ffffffff f76a8f75 S com.telerik.appbuildertabstest
+			return this.getFirstMatchingGroupFromMultilineResult<string>(psData, new RegExp(`.*?\s+${pid}\s+.*\s+(.*?)$`));
```
Mitko-Kerezov: I may be mistaken, but it appears you do not need the beginning of the RegExp
`.*?\s+${pid}\s+.*\s+(.*?)$` could become `${pid}\s+.*\s+(.*?)$` seeing as you're only interested in the end of the string

rosen-vladimirov: incorrect, consider the following line:
```
u0_a63   125512 1334  1519560 96040 ffffffff f76a8f75 S com.telerik.appbuildertabstest
```
If I use your RegExp and check for 25512, it will match...

mobile/android/android-application-manager.ts
```diff
@@ -83,6 +84,25 @@ export class AndroidApplicationManager extends ApplicationManagerBase {
...
 
+	public getDebuggableAppViews(appIdentifiers: string[]): IFuture<IDictionary<Mobile.IDebugWebViewInfo[]>> {
+		return ((): IDictionary<Mobile.IDebugWebViewInfo[]> => {
+			let mappedAppIdentifierPorts = this.$androidProcessService.getMappedAbstractToTcpPorts(this.identifier, appIdentifiers).wait(),
+				applicationViews: IDictionary<Mobile.IDebugWebViewInfo[]> = {};
+
+			_.each(mappedAppIdentifierPorts, (port: number, appIdentifier: string) => {
+				if (port) {
+					let apps = this.$httpClient.httpRequest(`http://127.0.0.1:${port}/json`).wait().body;
+					applicationViews[appIdentifier] = JSON.parse(apps);
```
Mitko-Kerezov: We could do with some error prevention/handling hare - if `apps` is false-like `JSON.parse` will throw an error

mobile/mobile-core/android-process-service.ts
Mitko-Kerezov: You're right but you can still do without a non-greedy match for any symbol in the beginning and just go with `\s+${pid}\s+.*\s+(.*?)$`

Mitko-Kerezov:
👍 after addressing comments

justcodebuilduser:
:heart:

PR #756, opened 2016-07-14
Split livesync-service-base so that we could have different implementations for AppBuilder and NativeScript

KristinaKoeva:


justcodebuilduser:
:heart:

rosen-vladimirov:
:+1:

PR #759, opened 2016-07-19
Allow debugging of WebViews inside NativeScript applications

rosen-vladimirov:
When there are webviews inside NativeScript application, mobile-cli-lib does not return them in getDebuggableApps as they have the same appIdentifier.
Also when trying to map abstract port for NativeScript application with WebViews, mobile-cli-lib finds the WebView first and maps port for it. In order to fix this, add `framework` parameter that will be passed from Proton.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,9 +7,9 @@ Contains common infrastructure for CLIs - mainly AppBuilder and NativeScript.
 Installation
 ===
 
-Latest version: 0.18.0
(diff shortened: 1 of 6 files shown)


justcodebuilduser:
:heart:

Mitko-Kerezov:
👍 after green build

justcodebuilduser:
:heart:

PR #762, opened 2016-07-20
Copy only files which are relevant to LiveSync

Mitko-Kerezov:
The array `localToDevicePaths` already contains all the filtered files which are relevant for performing LiveSync. Instead of copying the whole project folder, copy only those files.


justcodebuilduser:
:heart:

rosen-vladimirov:
:+1:

PR #765, opened 2016-07-20
Fix start package activity name

TsvetanMilanov:
We do not need to default to the Cordova start package activity name when we don't have project data because we can try to start NS application and the Cordova name will not work.


justcodebuilduser:
:heart:

PR #769, opened 2016-07-26
NativeScript LiveSyncServiceBase refactoring

KristinaKoeva:
Introduce platformSpecificLiveSyncServices map that is used in {N} live sync service
Rename the IPlatformLiveSyncService interface because it is more device oriented that platform


justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

PR #770, opened 2016-07-29
Clear forwarded ports for Android on process exit

TsvetanMilanov:
We need to clear the local ports which we use to debug Android applications after we receive the exit event of the process.

### DIFF
--- a/mobile/mobile-core/android-process-service.ts
+++ b/mobile/mobile-core/android-process-service.ts
@@ -1,15 +1,21 @@
 import {EOL} from "os";
+import * as shelljs from "shelljs";
 import {DeviceAndroidDebugBridge} from "../android/device-android-debug-bridge";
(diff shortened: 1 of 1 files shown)


justcodebuilduser:
:heart:

mobile/mobile-core/android-process-service.ts
```diff
@@ -25,6 +31,21 @@ export class AndroidProcessService implements Mobile.IAndroidProcessService {
 
 	public mapAbstractToTcpPort(deviceIdentifier: string, appIdentifier: string, framework: string): IFuture<string> {
 		return (() => {
+			if (this._shouldAddProcessExitEventListeners) {
```
rosen-vladimirov: can you extract the whole `if` to a new private method, for example `tryAttachToProcessExitSignals`

rosen-vladimirov:
:+1:

justcodebuilduser:
:heart:

PR #772, opened 2016-08-01
Remove the check whether the application is installed

KristinaKoeva:
When this method is executed right after the application is installed, the isApplicationInstalled returns false so we have to explicitly call checkForApplicationUpdates to update the application list. Since the tryStartApplication would fail if the app is not installed, and the exception would be caught, remove the isApplicationInstalled check


rosen-vladimirov:
:+1:  after green build

justcodebuilduser:
:heart:

PR #773, opened 2016-08-03
Add connectEventually with a timeout because otherwise we could wait indefinitely.

KristinaKoeva:


justcodebuilduser:
:heart:

rosen-vladimirov:
:+1:

PR #774, opened 2016-08-03
Fix process service unit test

TsvetanMilanov:
When we test the execution of all attached listeners we need to emit SIGTERM because if we emit exit or SIGINT the execution of the tests after this one will fail.


rosen-vladimirov:
:+1:  after green

justcodebuilduser:
:heart:

PR #775, opened 2016-08-09
Add support for installing npm modules and @types

rosen-vladimirov:
Add support for installing npm dependencies. Provide two methods for public API:
 - install - installs everything from package.json. In case a specific dependency must be installed, it's described as second argument with its name, version and should `@types` be installed.
 - uninstall - uninstalls the specified dependency from both node_modules and package.json. The call will also remove `@types/<dependency>` entry from devDependencies.

Still to do (in separate PRs)
* Add android17.d.ts and ios.d.ts to the project and reference them in .abreferences.d.ts
* Add unit tests
* Update README for new public API (requires bump of version in package.json)

### DIFF
--- a/appbuilder/declarations.d.ts
+++ b/appbuilder/declarations.d.ts
@@ -305,7 +305,7 @@ interface INpmService {
 
 	/**
 	 * Installs everything from package.json or specified dependency.
-	 * In case there's information which dependency to install, the method will check it and install only this dependency and possibly it's @types.
+	 * In case there's information which dependency to install, the method will check it and install only this dependency and possibly its @types.
 	 * @param {string} projectDir Directory of the project, where package.json is located.
 	 * @param @optional {INpmDependency} dependency Description of the dependency that has to be installed.
 	 * @return {IFuture<INpmInstallResult>} Returns object that will have error in case something fails.
--- a/appbuilder/declarations.d.ts
+++ b/appbuilder/declarations.d.ts
@@ -10,23 +10,24 @@ interface IDeployHelper {
 
 declare module Project {
 	interface IConstants {
-		PROJECT_FILE: string;
-		PROJECT_IGNORE_FILE: string;
+		ADDITIONAL_FILES_DIRECTORY: string;
+		ADDITIONAL_FILE_DISPOSITION: string;
+		APPBUILDER_PROJECT_PLATFORMS_NAMES: IDictionary<string>;
+		APPIDENTIFIER_PROPERTY_NAME: string;
+		CORDOVA_PLUGIN_VARIABLES_PROPERTY_NAME: string;
+		CORE_PLUGINS_PROPERTY_NAME: string;
 		DEBUG_CONFIGURATION_NAME: string;
 		DEBUG_PROJECT_FILE_NAME: string;
+		EXPERIMENTAL_TAG: string;
+		IMAGE_DEFINITIONS_FILE_NAME: string;
+		IONIC_PROJECT_PLATFORMS_NAMES: IDictionary<string>;
+		NATIVESCRIPT_APP_DIR_NAME: string;
+		PACKAGE_JSON_NAME: string;
+		PROJECT_FILE: string;
+		PROJECT_IGNORE_FILE: string;
+		REFERENCES_FILE_NAME: string;
 		RELEASE_CONFIGURATION_NAME: string;
 		RELEASE_PROJECT_FILE_NAME: string;
-		CORE_PLUGINS_PROPERTY_NAME: string;
-		CORDOVA_PLUGIN_VARIABLES_PROPERTY_NAME: string;
-		APPIDENTIFIER_PROPERTY_NAME: string;
-		EXPERIMENTAL_TAG: string;
-		NATIVESCRIPT_APP_DIR_NAME: string;
-		IMAGE_DEFINITIONS_FILE_NAME: string;
-		PACKAGE_JSON_NAME: string;
-		ADDITIONAL_FILE_DISPOSITION: string;
-		ADDITIONAL_FILES_DIRECTORY: string;
-		APPBUILDER_PROJECT_PLATFORMS_NAMES: IDictionary<string>;
-		IONIC_PROJECT_PLATFORMS_NAMES: IDictionary<string>;
 	}
 
 	interface ICapabilities {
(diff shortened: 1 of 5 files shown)


justcodebuilduser:
:heart:

appbuilder/services/npm-service.ts
```diff
@@ -0,0 +1,233 @@
...
+	}
+
+	public getPackageJsonFromNpmRegistry(packageName: string, version?: string): IFuture<any> {
+		return (() => {
+			let packageJsonContent: any;
+			version = version || "latest";
+			try {
+				let url = this.buildNpmRegistryUrl(packageName, version);
+				// This call will return error with message '{}' in case there's no such package.
+				let result = this.$httpClient.httpRequest(url).wait().body;
```
TsvetanMilanov: This call will not return error with this message. It will throw an exception with this message and response code 404 if the plugin is not found.
If you don't check the response code of the exception this method will never return {}.

appbuilder/services/npm-service.ts
```diff
@@ -0,0 +1,233 @@
...
+				this.$logger.trace(err.message);
+			}
+
+			return packageJsonContent;
+		}).future<any>()();
+	}
+
+	private hasTypesForDependency(packageName: string): IFuture<boolean> {
+		return (() => {
+			return !!this.getPackageJsonFromNpmRegistry(`${NpmService.TYPES_DIRECTORY}${packageName}`).wait();
```
TsvetanMilanov: In the current implementation of `getPackageJsonFromNpmRegistry` is planned to have case in which the method will return `{ }` (it is planned because you expect the request to return {} if the plugin is not found but actually the method will never return `{ }` if you don't address my comment for `this.$httpClient.httpRequest`). If  `getPackageJsonFromNpmRegistry` return `{ }` the result from `hasTypesForDependency` will be true (`!!{ }` is equal to `true`).

appbuilder/services/npm-service.ts
rosen-vladimirov: Our httpRequest method will throw error when the status code is 404 and will set the message to the content of the response. The content of the response is empty object. So yes, the method will never return empty object, but it will throw an error with message ={} and we'll go to the `catch`

appbuilder/services/npm-service.ts
```diff
@@ -0,0 +1,233 @@
...
+				_(packageJsonContent.devDependencies)
+					.keys()
+					.each(devDependency => {
+						if (this.isFromTypesRepo(devDependency)) {
+							let nodeModulesDirectory = path.join(projectDir, constants.NODE_MODULES_DIR_NAME);
+							let definitionFiles = this.$fs.enumerateFilesInDirectorySync(path.join(nodeModulesDirectory, devDependency),
+								(file, stat) => _.endsWith(file, constants.FileExtensions.TYPESCRIPT_DEFINITION_FILE) || stat.isDirectory(), { enumerateDirectories: false });
+
+							let defs = _.map(definitionFiles, def => {
+								return this.getReferenceLine(fromWindowsRelativePathToUnix(path.relative(nodeModulesDirectory, def)));
```
rosen-vladimirov: Ohh, you ate right. I've changed this in the last minute and maybe I didn't test it very well. Feel free to fix it in case it really produces incorrect .abreferences.d.ts

justcodebuilduser:
:heart:

TsvetanMilanov:
:+1:

appbuilder/services/npm-service.ts
```diff
@@ -0,0 +1,233 @@
...
+	@exportedPromise("npmService")
+	public install(projectDir: string, dependencyToInstall?: INpmDependency): IFuture<INpmInstallResult> {
+		return (() => {
+			let npmInstallResult: INpmInstallResult = {};
+
+			if (dependencyToInstall) {
+				npmInstallResult.result = {
+					isInstalled: false,
+					isTypesInstalled: false
+				};;
```
Mitko-Kerezov: Appears to have one `;` too many

appbuilder/services/npm-service.ts
```diff
@@ -0,0 +1,233 @@
...
+
+	private generateReferencesFile(projectDir: string): IFuture<void> {
+		return (() => {
+			let packageJsonContent = this.getPackageJsonContent(projectDir).wait();
+
+			let pathToReferenceFile = this.getPathToReferencesFile(projectDir),
+				lines: string[] = [];
+
+			if (packageJsonContent && packageJsonContent.dependencies && packageJsonContent.dependencies[constants.TNS_CORE_MODULES]) {
+				let relativePathToTnsCoreModulesDts = `./${constants.NODE_MODULES_DIR_NAME}/${constants.TNS_CORE_MODULES}/${NpmService.TNS_CORE_MODULES_DEFINITION_FILE_NAME}`;
```
Mitko-Kerezov: I know that we're talking about `nodejs` here, but isn't it safer to use `path.join` instead of using `UNIX`-style directory separators everywhere

appbuilder/services/npm-service.ts
```diff
@@ -0,0 +1,233 @@
...
+
+				if (this.$fs.exists(path.join(projectDir, relativePathToTnsCoreModulesDts)).wait()) {
+					lines.push(this.getReferenceLine(relativePathToTnsCoreModulesDts));
+				}
+			}
+
+			if (packageJsonContent && packageJsonContent.devDependencies) {
```
Mitko-Kerezov: This check seems unnecessary - `packageJsonContent` will either be an object or the method `this.getPackageJsonContent` will throw. As for checking for `packageJsonContent.devDependencies` I do believe `lodash` will handle the case of it being `undefined`

appbuilder/services/npm-service.ts
```diff
@@ -0,0 +1,233 @@
...
+				_(packageJsonContent.devDependencies)
+					.keys()
+					.each(devDependency => {
+						if (this.isFromTypesRepo(devDependency)) {
+							let nodeModulesDirectory = path.join(projectDir, constants.NODE_MODULES_DIR_NAME);
+							let definitionFiles = this.$fs.enumerateFilesInDirectorySync(path.join(nodeModulesDirectory, devDependency),
+								(file, stat) => _.endsWith(file, constants.FileExtensions.TYPESCRIPT_DEFINITION_FILE) || stat.isDirectory(), { enumerateDirectories: false });
+
+							let defs = _.map(definitionFiles, def => {
```
Mitko-Kerezov: you can safely remove the curly brackets along with the `return` statement in this fat arrow function

appbuilder/services/npm-service.ts
```diff
@@ -0,0 +1,233 @@
...
+
+			if (this.$hostInfo.isWindows) {
+				this._npmExecutableName += ".cmd";
+			}
+		}
+
+		return this._npmExecutableName;
+	}
+
+	private getNpmArguments(command: string, npmArguments?: string[]): string[] {
```
Mitko-Kerezov: I do believe you can set `npmArguments` a default value like so
```TypeScript
private getNpmArguments(command: string, npmArguments: string[] = []): string[] {
```

and avoid the check on the line below

appbuilder/services/npm-service.ts
```diff
@@ -0,0 +1,233 @@
...
+		let args: string[] = npmArguments || [];
+		return args.concat([command]);
```
Mitko-Kerezov: Shouldn't this `concat` be the other way around. I could be mistaken, but I believe this way we spawn `$ npm lodash install` for example

Mitko-Kerezov: Nevermind - those are only dashed arguments and there's no difference between `$ npm install lodash --save` and `$ npm --save install lodash` so ignore me ❎

appbuilder/declarations.d.ts
```diff
@@ -238,3 +239,86 @@ interface ICompanionAppsService {
...
+	 * The method will remove them from package.json and from node_modules dir.
+	 * @param {string} projectDir Directory of the project, where package.json is located.
+	 * @param {string} dependency The name of the dependency that has to be removed.
+	 * @return {IFuture<void>}
+	 */
+	uninstall(projectDir: string, dependency: string): IFuture<void>;
+
+	/**
+	 * Installs everything from package.json or specified dependency.
+	 * In case there's information which dependency to install, the method will check it and install only this dependency and possibly it's @types.
```
Mitko-Kerezov: Typo - **its** instead of it's

appbuilder/services/npm-service.ts
rosen-vladimirov: Nope, the path in references.d.ts must be Unix style. So I prefer to write it Unix style instead of using `path.join` and convert from Windows style to Unix

justcodebuilduser:
:heart:

Mitko-Kerezov:
LGTM :+1:

PR #779, opened 2016-08-16
Provide a way to find scoped dependencies

TsvetanMilanov:
We need to have a way to find scoped dependencies (e.g. @angular/core).
Currently the flow is:
Call https://www.npmjs.com/search?q={query}
If that fails:
If the query starts with @, we know npm search will fail, so try to get the package details directly from http://registry.npmjs.org. It will only work if the name is exact match, but it's better than nothing.
If the query does not start with @, fall back to npm search.


justcodebuilduser:
:broken_heart:

justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

appbuilder/services/npm-service.ts
```diff
@@ -19,6 +19,10 @@ export class NpmService implements INpmService {
 		private $logger: ILogger,
 		private $projectConstants: Project.IConstants) { }
 
+	private get scopedDependencyRegExp(): RegExp {
```
Mitko-Kerezov: this could be `STATIC`

appbuilder/services/npm-service.ts
```diff
@@ -75,6 +79,54 @@ export class NpmService implements INpmService {
...
 	}
 
+	public search(projectDir: string, keywords: string[], args?: string[]): IFuture<IBasicPluginInformation[]> {
+		return ((): IBasicPluginInformation[] => {
+			let result: IBasicPluginInformation[] = [];
+			let spawnResult = this.executeNpmCommand(projectDir, args || [], keywords.join(" ")).wait();
+			if (spawnResult.stderr) {
+				// npm will write "npm WARN Building the local index for the first time, please be patient" to the stderr and if it is the only message on the stderr we should ignore it.
+				let splitError = spawnResult.stderr.split("\n");
+				if (splitError.length > 2 || splitError[0].indexOf("Building the local index for the first time") === -1) {
```
Mitko-Kerezov: The comment above states that if `Building the local index...` is the **only**error message we should ignore it, but then why is the condition `splitError.length > 2` instead of `splitError.length > 1`? If I get two errors and the first is `Building the local index...` then I won't actually get an error.

appbuilder/services/npm-service.ts
```diff
@@ -75,6 +79,54 @@ export class NpmService implements INpmService {
 		}).future<void>()();
 	}
 
+	public search(projectDir: string, keywords: string[], args?: string[]): IFuture<IBasicPluginInformation[]> {
+		return ((): IBasicPluginInformation[] => {
+			let result: IBasicPluginInformation[] = [];
+			let spawnResult = this.executeNpmCommand(projectDir, args || [], keywords.join(" ")).wait();
```
Mitko-Kerezov: It appears that the argument `"search"` is missing here. If only keywords are passed you'd get an error here since you'd spawn `$ npm <keywords>` instead of `$ npm search <keywords>`

appbuilder/services/npm-service.ts
```diff
@@ -75,6 +79,54 @@ export class NpmService implements INpmService {
...
+			// Sample output:
+			// NAME                    DESCRIPTION             AUTHOR        DATE       VERSION  KEYWORDS
+			// cordova-plugin-console  Cordova Console Plugin  =csantanapr…  2016-04-20 1.0.3    cordova console ecosystem:cordova cordova-ios
+			let pluginsRows: string[] = spawnResult.stdout.split("\n");
+
+			// Remove the table headers row.
+			pluginsRows.shift();
+
+			let npmNameGroup = "(\\S+)";
+			let npmDateGroup = "(\\d+\\-\\d+\\-\\d+)\\s";
```
Mitko-Kerezov: I believe escaping the `-` symbol is unnecessary

appbuilder/services/npm-service.ts
```diff
@@ -75,6 +79,54 @@ export class NpmService implements INpmService {
...
+			let npmFreeTextGroup = "([^=]+)";
+			let npmAuthorsGroup = "((?:=\\S+\\s?)+)\\s+";
+
+			// Should look like this /(\S+)\s+([^=]+)((?:=\S+\s?)+)\s+(\d+\-\d+\-\d+)\s(\S+)(\s+([^=]+))?/
+			let pluginRowRegExp = new RegExp(`${npmNameGroup}\\s+${npmFreeTextGroup}${npmAuthorsGroup}${npmDateGroup}${npmNameGroup}(\\s+${npmFreeTextGroup})?`);
```
Mitko-Kerezov: This whole RegExp can be static also along with all of its components so that they're not created every time the function is executed

appbuilder/services/npm-service.ts
```diff
@@ -93,6 +145,21 @@ export class NpmService implements INpmService {
...
+		let matches = dependency.match(this.scopedDependencyRegExp);
+
+		return !!(matches && matches[0]);
+	}
+
+	public getScopedDependencyInformation(dependency: string): IScopedDependencyInformation {
+		let matches = dependency.match(this.scopedDependencyRegExp);
+
+		return {
+			name: matches[1],
```
Mitko-Kerezov: You should probably check if this is in fact a scoped dependency by calling `this.isScopedDependency` beforehand. Otherwise you might end up with `matches` being undefined and a null reference here

services/plugins/npm-plugins-source.ts
```diff
@@ -0,0 +1,48 @@
...
+		return (() => {
+			super.initialize(projectDir, keywords).wait();
+
+			let searchFuture = this.$npmService.search(this._projectDir, keywords);
+
+			this.$logger.printInfoMessageOnSameLine("Searching for plugins in npm, please wait.");
+
+			this.$progressIndicator.showProgressIndicator(searchFuture, 2000).wait();
+
+			this._plugins = searchFuture.isResolved() ? searchFuture.get() : searchFuture.wait();
```
Mitko-Kerezov: `showProgressIndicator`'s implementation `wait`s for the future passed as a parameter. Is there a reason you check if it's resolved here instead of directly getting the result?

services/plugins/npm-plugins-source.ts
```diff
@@ -0,0 +1,48 @@
...
+		}).future<void>()();
+	}
+
+	public getPlugins(page: number, count: number): IFuture<IBasicPluginInformation[]> {
+		return ((): IBasicPluginInformation[] => {
+			let result: IBasicPluginInformation[] = [];
+			let skip = page * count;
+
+			result = _.slice(this._plugins, skip, skip + count);
```
Mitko-Kerezov: you can just `return` here without the need of the additional `result` variable

services/plugins/npm-registry-plugins-source.ts
```diff
@@ -0,0 +1,52 @@
...
+			return Future.fromResult(null);
+		}
+	}
+
+	public getAllPlugins(): IFuture<IBasicPluginInformation[]> {
+		return Future.fromResult(this._plugins);
+	}
+
+	private prepareScopedPluginName(plugin: string): string {
+		let pluginName = plugin.replace("/", "%2F");
```
Mitko-Kerezov: you can `return` here

services/plugins/npm-plugins-source.ts
```diff
@@ -0,0 +1,48 @@
...
+
+			return result;
+		}).future<IBasicPluginInformation[]>()();
+	}
+
+	public getAllPlugins(): IFuture<IBasicPluginInformation[]> {
+		return Future.fromResult(this._plugins);
+	}
+
+	public hasPlugins(): boolean {
```
Mitko-Kerezov: There's no need to override this method - you can just delete it and use the parent's

services/plugins/npmjs-plugins-source.ts
```diff
@@ -0,0 +1,113 @@
...
+				if (currentPluginsFound && currentPluginsFound.length) {
+					result = result.concat(currentPluginsFound);
+				}
+			} while (currentPluginsFound && currentPluginsFound.length);
+
+			return result;
+		}).future<IBasicPluginInformation[]>()();
+	}
+
+	public hasPlugins(): boolean {
```
Mitko-Kerezov: There's no need to override this method too - you can just delete it and use the parent's

services/plugins/npm-registry-plugins-source.ts
```diff
@@ -0,0 +1,52 @@
...
+	public getPlugins(page: number, count: number): IFuture<IBasicPluginInformation[]> {
+		if (!this._hasReturnedPlugin) {
+			this._hasReturnedPlugin = true;
+			return Future.fromResult(this._plugins);
+		} else {
+			return Future.fromResult(null);
+		}
+	}
+
+	public getAllPlugins(): IFuture<IBasicPluginInformation[]> {
```
Mitko-Kerezov: You can extract this implementation in the base class and delete it in all inheritors apart from `NpmjsPluginsSource`

services/plugins/npm-registry-plugins-source.ts
```diff
@@ -0,0 +1,52 @@
...
+			this._plugins = [this.getPluginFromNpmRegistry(keywords[0]).wait()];
+		}).future<void>()();
+	}
+
+	public getPlugins(page: number, count: number): IFuture<IBasicPluginInformation[]> {
+		if (!this._hasReturnedPlugin) {
+			this._hasReturnedPlugin = true;
+			return Future.fromResult(this._plugins);
+		} else {
+			return Future.fromResult(null);
```
Mitko-Kerezov: ❔ Why would you only want to return the result once ❔

services/plugins/npmjs-plugins-source.ts
```diff
@@ -0,0 +1,113 @@
...
+		return !!(this._plugins && this._plugins.length);
+	}
+
+	private prepareScopedPluginName(keywords: string[]): string {
+		let pluginName = keywords[0];
+
+		pluginName = pluginName.replace("@", "%40");
```
Mitko-Kerezov: You could do a full `encodeURIComponent` here instead of replacing `@` and `/` only - there may be more special characters along the way too

services/plugins/npmjs-plugins-source.ts
```diff
@@ -0,0 +1,113 @@
...
+		pluginName = pluginName.replace("/", "%2F");
+
+		return pluginName;
+	}
+
+	private getPluginsFromNpmjs(keywords: string[], page: number): IFuture<IBasicPluginInformation[]> {
+		return ((): IBasicPluginInformation[] => {
+			let pluginName = this.prepareScopedPluginName(keywords);
```
Mitko-Kerezov: The logic here does not allow for someone to query for plugins using multiple keywords. If for example I'd want to search for `$ appbuilder plugin find native script` I'd only get results for the first word instead of all the other words

services/plugins/npmjs-plugins-source.ts
```diff
@@ -0,0 +1,113 @@
...
+			let url = `${NpmjsPluginsSource.NPMJS_ADDRESS}/search?q=${pluginName}&page=${page}`;
+
+			try {
+				let responseBody: string = this.$httpClient.httpRequest(url).wait().body;
+				let $ = cheerio.load(responseBody);
```
Mitko-Kerezov: the `cheerio` module might be a bit overkill for what we're trying to achieve here. You can take a look at [parser5](https://github.com/inikulin/parse5) or [htmlparser2](https://github.com/fb55/htmlparser2) and see if we can't tailor that for our needs

services/plugins/npmjs-plugins-source.ts
```diff
@@ -0,0 +1,113 @@
...
+
+			this._keywords = keywords;
+
+			this._plugins = this.getPluginsFromNpmjs(keywords, 0).wait();
+		}).future<void>()();
+	}
+
+	public getPlugins(page: number, count: number): IFuture<IBasicPluginInformation[]> {
+		return ((): IBasicPluginInformation[] => {
+			if (!this._hasReturnedTheInitialPlugins && page === 0) {
```
Mitko-Kerezov: I don't quite understand this logic - `this._hasReturnedTheInitialPlugins`'s value is never changed so it'll always be `undefined`. Moreover, you return `this._plugins` here and this object may contain multiple pages worth of information because of the line `this._plugins = this._plugins.concat(result);` below

justcodebuilduser:
:heart:

appbuilder/services/npm-service.ts
TsvetanMilanov: I think it is better to check if the dependency is scoped before this method is called because if we check here and return `null` if the dependency is not scoped the caller of this method should check if the result is `null`.

justcodebuilduser:
:broken_heart:

justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

services/plugins/npm-plugins-service.ts
```diff
@@ -0,0 +1,49 @@
...
+
+	public search(projectDir: string, keywords: string[], modifySearchQuery: (keywords: string[]) => string[]): IFuture<IPluginsSource> {
+		return ((): IPluginsSource => {
+			let npmjsPluginsSource = this.preparePluginsSource(NpmjsPluginsSource, projectDir, keywords).wait();
+
+			if (npmjsPluginsSource.hasPlugins()) {
+				return npmjsPluginsSource;
+			}
+
+			let npmRegistryPluginsSource = this.preparePluginsSource(NpmRegistryPluginsSource, projectDir, keywords).wait();
```
rosen-vladimirov: you can extract the lines from 16 to 20 to a private method and call them here and in `optimizedSearch` method as well

services/plugins/print-plugins-service.ts
```diff
@@ -0,0 +1,84 @@
...
+	private createPluginsTable(data: string[][]): any {
+		let headers: string[] = ["Plugin", "Version", "Author", "Description"];
+
+		let table: any = createTable(headers, data);
+
+		return table;
+	}
+
+	private createTableCells(plugins: IBasicPluginInformation[]): string[][] {
+		return plugins.map((plugin: IBasicPluginInformation) => {
```
rosen-vladimirov: use lodash here, just in case someone passes undefined argument ;)

services/plugins/print-plugins-service.ts
```diff
@@ -0,0 +1,84 @@
...
+				this.displayTableWithPlugins(getAllPluginsFuture.get());
+				return;
+			}
+
+			let pluginsSearchFuture = pluginsSource.getPlugins(this._page++, count);
+
+			this.$progressIndicator.showProgressIndicator(pluginsSearchFuture, PrintPluginsService.PROGRESS_INDICATOR_TIMEOUT).wait();
+
+			let pluginsToDisplay: IBasicPluginInformation[] = pluginsSearchFuture.get();
+			let shouldDisplayMorePlugins: boolean = true;
```
rosen-vladimirov: there's no need to define the type when you set it to boolean value ;)

rosen-vladimirov:
:+1:  after green build

justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

PR #780, opened 2016-08-17
Add documentation for npm service

TsvetanMilanov:
Add description and sample usage for the public methods in our npm service.

New methods:
`install(projectDir: string, dependencyToInstall?: INpmDependency): Promise<INpmInstallResult>` - Installs everything from package.json or specified dependency. In case there's information which dependency to install, the method will check it and install only this dependency and possibly it's @types.
`uninstall(projectDir: string, dependency: string): Promise` - Uninstalls the dependency and the @types/ devDependency. The method will remove them from package.json and from node_modules dir.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,9 +7,9 @@ Contains common infrastructure for CLIs - mainly AppBuilder and NativeScript.
 Installation
 ===
 
-Latest version: 0.19.0
+Latest version: 0.20.0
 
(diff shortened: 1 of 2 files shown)


justcodebuilduser:
:heart:

Mitko-Kerezov:
👍

PR #783, opened 2016-08-22
Delete typings dir if it is empty

TsvetanMilanov:
After we migrate TS projects we should delete the typings directory if it is empty after we remove the tns-core-modules directory.


justcodebuilduser:
:heart:

rosen-vladimirov:
:+1:

PR #784, opened 2016-08-22
Fix error when getting android id

TsvetanMilanov:
When we have only AndroidManifest.xml or config.xml in the project and the app id in it is not changed we do not need to throw exception for app id mismatch.

### DIFF
--- a/appbuilder/project/project-base.ts
+++ b/appbuilder/project/project-base.ts
@@ -88,6 +88,8 @@ export abstract class ProjectBase implements Project.IProjectBase {
 	public getAppIdentifierForPlatform(platform?: string): IFuture<string> {
 		return ((): string => {
 			if (!this._platformSpecificAppIdentifier) {
+				this._platformSpecificAppIdentifier = this.projectData.AppIdentifier;
+
 				if (platform &&
 					platform.toLowerCase() === this.$projectConstants.ANDROID_PLATFORM_NAME.toLowerCase() &&
 					this.projectData.Framework === TARGET_FRAMEWORK_IDENTIFIERS.Cordova) {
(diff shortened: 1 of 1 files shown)


justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

appbuilder/project/project-base.ts
```diff
@@ -101,17 +103,15 @@ export abstract class ProjectBase implements Project.IProjectBase {
...
 					if (appIdentifierInAndroidManifest && appIdentifierInConfigXml && appIdentifierInAndroidManifest === appIdentifierInConfigXml) {
-						if (appIdentifierInAndroidManifest === ProjectBase.APP_IDENTIFIER_PLACEHOLDER) {
-							this._platformSpecificAppIdentifier = this.projectData.AppIdentifier;
-						} else {
+						if (appIdentifierInAndroidManifest !== ProjectBase.APP_IDENTIFIER_PLACEHOLDER) {
 							this._platformSpecificAppIdentifier = appIdentifierInAndroidManifest;
 						}
 					} else if (appIdentifierInAndroidManifest || appIdentifierInConfigXml) {
-						this.$errors.failWithoutHelp(`Your package in ${ProjectBase.ANDROID_MANIFEST_NAME} and id in ${ProjectBase.CONFIG_XML_NAME} do not match. They must be the same to be able to build your application.`);
+						let appId = appIdentifierInAndroidManifest || appIdentifierInConfigXml;
```
rosen-vladimirov: move this above the
```TypeScript
if (appIdentifierInAndroidManifest && appIdentifierInConfigXml && appIdentifierInAndroidManifest === appIdentifierInConfigXml)
```
and change the else if here to:
```TypeScript
else if (appId && (appId !== ProjectBase.APP_IDENTIFIER_PLACEHOLDER))
```
so the else if's body can remain as it was - containing only
```TypeScript
this.$errors.failWithoutHelp(`Your package in ${ProjectBase.ANDROID_MANIFEST_NAME} and id in ${ProjectBase.CONFIG_XML_NAME} do not match. They must be the same to be able to build your application.`);
```

justcodebuilduser:
:heart:

rosen-vladimirov:
:+1:

PR #786, opened 2016-08-25
Support noEmitOnError parameter

rosen-vladimirov:
Previously there were some cases when TypeScript was unable to generate .js files. After adding the `noEmitOnError` parameter to tsconfig, TypeScript will not generate .js files ONLY when it is set to true.
Currently our TypeScriptService does not handle this case as it thinks that the .js files are not generated in some cases. Respect the parameter and always show the output of the compiler to the user.


justcodebuilduser:
:heart:

Mitko-Kerezov:
👍 Seems legit

PR #787, opened 2016-08-25
Regenerate .abreferences.d.ts file on plugin remove

rosen-vladimirov:
When uninstalling plugin, the .abreferences.d.ts file must be regenerated in order to remove the entries from @types, which are removed from the file system.


justcodebuilduser:
:heart:

TsvetanMilanov:
:+1:

PR #791, opened 2016-08-30
Improve logging in iOS

tzraikov:
- Excluded "NativeScript loaded bundle" messages from console output. Those messages occur only when running on device. Probably we could define an option (e.g. --log debug)
- Excluded file:row:column info from console log messages for device in iOS. This syncs the output with iOS simulator and Android.
- Changed the `Mounting` message from info to debug. This way it can be hidden by default, it will be visible when running the cli with option --log debug

### DIFF
--- a/mobile/ios/device/ios-core.ts
+++ b/mobile/ios/device/ios-core.ts
@@ -1112,6 +1112,23 @@ class GDBStandardOutputAdapter extends stream.Transform {
 				}
 			}
 
+			if (result.indexOf("NativeScript loaded bundle") >= 0) {
+				result = "";
+			}
+
+			// CONSOLE LOG messages comme in the following form:
+ 			// <date> <domain> <app>[pid] CONSOLE LOG file:///location:row:column: <actual message goes here>
+ 			// This code removes the first part and leaves only the message as specified by the call to console.log function.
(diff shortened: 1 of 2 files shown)


justcodebuilduser:
:broken_heart:

justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

rosen-vladimirov:
I'm closing this PR as with latest changes, filtering in the output should be in the ios-log-filter. Check this PR for more information. In case a new way of filtering is needed, that is not applicable to Cordova projects you can:
1) Create new class for iOS Log Filter and register it in `nativescript-cli`.
2) Move the [require from common bootstrap](https://github.com/telerik/mobile-cli-lib/blob/master/bootstrap.ts#L120) to [appbuilder-bootstrap](https://github.com/telerik/mobile-cli-lib/blob/master/appbuilder/appbuilder-bootstrap.ts).
3) Add new require in nativescript CLI's [bootstrap](https://github.com/NativeScript/nativescript-cli/blob/master/lib/bootstrap.ts) and register iOSLogFilter in it.

This way the filtering logic will be used only in NativeScript CLI.

PR #793, opened 2016-08-30
Separate .d.ts files from .ts files.

TsvetanMilanov:
When we get the TS files data of the project we don't need the .d.ts files in the typeScriptFiles property.
Also when we add the TS files to tscommand.txt we need to add quotation marks to their names because some of them may contain spaces.

### DIFF
--- a/services/typescript-service.ts
+++ b/services/typescript-service.ts
@@ -3,6 +3,7 @@ import * as os from "os";
 import temp = require("temp");
 import {exportedPromise} from "../decorators";
 import {NODE_MODULES_DIR_NAME, FileExtensions} from "../constants";
+import {quoteString} from "../helpers";
 temp.track();
 
 interface ITypeScriptCompilerMessages {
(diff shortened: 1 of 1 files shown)


justcodebuilduser:
:heart:

services/typescript-service.ts
```diff
@@ -310,6 +314,16 @@ export class TypeScriptService implements ITypeScriptService {
 			return tempDir;
 		}).future<string>()();
 	}
+
+	private addQuotationMarksToFileNames(files: string[]): string[] {
```
rosen-vladimirov: I would name this `quoteFileName`, but it's just an opinion

services/typescript-service.ts
```diff
@@ -310,6 +314,16 @@ export class TypeScriptService implements ITypeScriptService {
...
+		return _.map(files, (fileName: string) => {
+			if (fileName[0] === "\"" && fileName[fileName.length - 1] === "\"") {
+				return fileName;
+			}
+
+			return `"${fileName}"`;
```
rosen-vladimirov: the code is incorrect - on Linux and Mac OS X you can have `"` in the filename (try executing `$ touch 'test"1'`).
Maybe instead of adding quotes, we can replace the space in the filename with `\ `. Will this work?

justcodebuilduser:
:heart:

services/typescript-service.ts
```diff
@@ -310,6 +315,10 @@ export class TypeScriptService implements ITypeScriptService {
 			return tempDir;
 		}).future<string>()();
 	}
+
+	private quoteFileName(files: string[]): string[] {
```
rosen-vladimirov: `quoteFileNames`

rosen-vladimirov:
:+1:  after fixing comment

justcodebuilduser:
:broken_heart:

justcodebuilduser:
:broken_heart:

justcodebuilduser:
:heart:

PR #797, opened 2016-09-07
Use monkey to start Android applications

rosen-vladimirov:
Use the monkey tool to start Android applications on device. This way we do not have to care about the startPackageActivity name and we can start the application in case a custom activity is used.
Remove unused code and variables.


mobile/android/android-application-manager.ts
```diff
@@ -41,18 +41,10 @@ export class AndroidApplicationManager extends ApplicationManagerBase {
...
-			let startActivityNames = startActivityName ? [startActivityName] : defaultActivityNames;
-
-			_.each(startActivityNames, (activityName: string) => {
-				this.adb.executeShellCommand(["am", "start",
-					"-a", "android.intent.action.MAIN",
-					"-n", `${appIdentifier}/${activityName}`,
-					"-c", "android.intent.category.LAUNCHER"]).wait();
-			});
+			this.adb.executeShellCommand(["monkey",
+				"-p", `${appIdentifier}`,
```
Mitko-Kerezov: I don't think there's a need for the template string here - `appIdentifier` is already of type string

rosen-vladimirov: :smile_cat: thanks for noticing this

justcodebuilduser:
:heart:

Mitko-Kerezov:
👍

justcodebuilduser:
:heart:

PR #802, opened 2016-09-10
Fix get first page plugins trom npmjs twice

TsvetanMilanov:
When we get the plugins from http://npmjs.org we need to start from page 1 not from page 0 because page 1 and page 0 are the same.


justcodebuilduser:
:heart:

PR #803, opened 2016-09-12
Use rimraf to remove tns-core-modules

TsvetanMilanov:
When we migrate TS project and we remove the tns-core-modules we use the shelljs to remove the directory but there is some bug when using it - it fails to delete some of the subdirecotries. Thet's why we will use rimraf instead of shelljs.

Related to: https://github.com/Icenium/icenium-cli/pull/1553


justcodebuilduser:
:heart:

PR #805, opened 2016-09-15
Keep the current indentation of json files

TsvetanMilanov:
Since our other clients use 2 spaces for indentation in json files we should use 2 spaces too. Since some users are using only the CLI we should not change their current indentation.

### DIFF
--- a/file-system.ts
+++ b/file-system.ts
@@ -9,6 +9,9 @@ import * as shelljs from "shelljs";
 
 @injector.register("fs")
 export class FileSystem implements IFileSystem {
+	private static DEFAULT_INDENTATION_CHARACTER = "\t";
+	private static JSON_OBJECT_REGEXP = new RegExp(`{\\r*\\n*(\\W*)"`, "m");
+
 	constructor(private $injector: IInjector) { }
 
 	//TODO: try 'archiver' module for zipping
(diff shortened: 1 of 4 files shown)


TsvetanMilanov:
run ci

justcodebuilduser:
:broken_heart:

justcodebuilduser:
:broken_heart:

TsvetanMilanov:
run ci

justcodebuilduser:
:broken_heart:

file-system.ts
```diff
@@ -523,4 +530,22 @@ export class FileSystem implements IFileSystem {
...
+	private getIndentationCharacter(filePath: string): IFuture<string> {
+		return ((): string => {
+			if (!this.exists(filePath).wait()) {
+				return FileSystem.DEFAULT_INDENTATION_CHARACTER;
+			}
+
+			let fileContent = this.readText(filePath).wait().trim();
+			let matches = fileContent.match(FileSystem.JSON_OBJECT_REGEXP);
+
+			let indentation = matches[1];
```
Mitko-Kerezov: You'd get a null reference here in case the Regex doesn't match

Mitko-Kerezov:
👍 hopefully after green build

justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

test/unit-tests/file-system.ts
```diff
@@ -18,6 +18,57 @@ function isOsCaseSensitive(testInjector: IInjector): boolean {
...
+			testCondition: "when the indentation is space",
+			expectedIndentation: " "
+		}, {
+			exists: true,
+			text: `{\n  "a" : 5 }`,
+			testCondition: "when the indentation is two spaces",
+			expectedIndentation: "  "
+		}, {
+			exists: false,
+			text: `{\n "a" : 5 }`,
```
Mitko-Kerezov: Why'd you have any `text` when the file doesn't exist 👀 
I'd prefer this property have an empty value for additional clarity

TsvetanMilanov: I need the `text` property here because i need to verify that in this test case this text will be written using `\t` instead of ` `.

PR #806, opened 2016-09-17
Validate hook arguments before execute

TsvetanMilanov:
Since we can have hooks with arguments which are invalid for the AppBuilder CLI (for example the before-livesync hook of nativescript-angular and watch hook of nativescript-dev-typescript) we need to check if the arguments of the hook are valid before we execute it. We can do the check only for hooks which can be executed in our process.

Fixes http://teampulse.telerik.com/view#item/321923

### DIFF
--- a/services/hooks-service.ts
+++ b/services/hooks-service.ts
@@ -1,5 +1,6 @@
 import * as path from "path";
 import * as util from "util";
+import {annotate} from "../helpers";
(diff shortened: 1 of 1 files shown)


justcodebuilduser:
:heart:

rosen-vladimirov:
:+1:

justcodebuilduser:
:heart:

PR #809, opened 2016-09-21
Respect npm settings - proxy and registry

rosen-vladimirov:
Use npm proxy for connections to registry.npmjs.org. This way in case the user has set a proxy, we'll use it.

CLI makes calls to registry.npmjs.org. But the users can configure different registry (`npm config set registry <...>`), so CLI should respect it instead of calls to `registry.npmjs.org`.

### DIFF
--- a/appbuilder/services/npm-service.ts
+++ b/appbuilder/services/npm-service.ts
@@ -171,7 +171,7 @@ export class NpmService implements INpmService {
 			let packageJsonContent: any;
 			version = version || "latest";
 			try {
-				let url = this.buildNpmRegistryUrl(packageName, version);
+				let url = this.buildNpmRegistryUrl(packageName, version).wait();
 				// This call will return error with message '{}' in case there's no such package.
 				let result = this.$httpClient.httpRequest(url, this.getNpmProxySettings().wait()).wait().body;
 				packageJsonContent = JSON.parse(result);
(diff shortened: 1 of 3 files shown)


justcodebuilduser:
:heart:

appbuilder/services/npm-service.ts
```diff
@@ -203,8 +206,32 @@ export class NpmService implements INpmService {
 		}).future<boolean>()();
 	}
 
-	private buildNpmRegistryUrl(packageName: string, version: string): string {
-		return `${NpmService.NPM_REGISTRY_URL}/${packageName.replace("/", "%2F")}?version=${encodeURIComponent(version)}`;
+	private buildNpmRegistryUrl(packageName: string, version: string): IFuture<string> {
+		return (() => {
+			let registryUrl = this.getNpmRegistryUrl().wait();
+			return `${registryUrl}/${packageName.replace("/", "%2F")}?version=${encodeURIComponent(version)}`;
```
Mitko-Kerezov: There's one [more place](https://github.com/telerik/mobile-cli-lib/blob/master/services/plugins/npm-registry-plugins-source.ts#L31) we use this dreaded `.replace("/", "%2F")`. I'd prefer we extract it to a common place in case some other symbol needs to be replaced this way.

This could be done in a separate PR though

appbuilder/services/npm-service.ts
```diff
@@ -320,5 +347,31 @@ export class NpmService implements INpmService {
 	private executeNpmCommandCore(projectDir: string, npmArguments: string[]): IFuture<ISpawnResult> {
 		return this.$childProcess.spawnFromEvent(this.npmExecutableName, npmArguments, "close", { cwd: projectDir });
 	}
+
+	private getNpmProxySettings(): IFuture<IProxySettings> {
+		return ((): IProxySettings => {
+			if (!this._hasCheckedNpmProxy) {
+				try {
+					let npmProxy = (this.$childProcess.exec("npm config get proxy").wait() || "").toString();
```
Mitko-Kerezov: Same remark about the `.trim()`

appbuilder/services/npm-service.ts
rosen-vladimirov: Yep, a separate PR it will be.

appbuilder/services/npm-service.ts
```diff
@@ -203,8 +206,32 @@ export class NpmService implements INpmService {
...
+	}
+
+	private _npmRegistryUrl: string;
+	private getNpmRegistryUrl(): IFuture<string> {
+		return ((): string => {
+			if (!this._npmRegistryUrl) {
+				let currentNpmRegistry: string;
+
+				try {
+					currentNpmRegistry = (this.$childProcess.exec("npm config get registry").wait() || "").toString();
```
rosen-vladimirov: handled

appbuilder/services/npm-service.ts
rosen-vladimirov: handled

justcodebuilduser:
:heart:

PR #810, opened 2016-09-21
Fix registry url

rosen-vladimirov:
Fix registry URL as it contains two slashes which causes code 500 of the http request.


justcodebuilduser:
:heart:

PR #811, opened 2016-09-26
Fix check for ios simulator

TsvetanMilanov:
When we check if there is running simulator we use ps cax but if the simulator is killed for a short period of time after that there is a process listed when using ps cax and we decide that there is a running simulator and then we try to use it but ios-sim-portable returns null. We should check if the simulator returned from ios-sim-portable is not null or undefined in order to continue to work with it.

### DIFF
--- a/test/unit-tests/mobile/ios-simulator-discovery.ts
+++ b/test/unit-tests/mobile/ios-simulator-discovery.ts
@@ -4,14 +4,11 @@ import Future = require("fibers/future");
 import { assert } from "chai";
 import { DevicePlatformsConstants } from "../../../mobile/device-platforms-constants";
 
-let currentlyRunningSimulator: any,
-	isCurrentlyRunning: boolean;
+let currentlyRunningSimulator: any;
 
 function createTestInjector(): IInjector {
 	let injector = new Yok();
-	injector.register("childProcess", {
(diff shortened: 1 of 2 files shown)


justcodebuilduser:
:broken_heart:

justcodebuilduser:
:heart:

test/unit-tests/mobile/ios-simulator-discovery.ts
```diff
@@ -4,13 +4,12 @@ import Future = require("fibers/future");
...
 
-let currentlyRunningSimulator: any,
-	isCurrentlyRunning: boolean;
+let currentlyRunningSimulator: any;
 
 function createTestInjector(): IInjector {
 	let injector = new Yok();
 	injector.register("childProcess", {
-		exec: (command: string) => Future.fromResult(isCurrentlyRunning ? 'launchd_sim' : '')
+		exec: (command: string) => Future.fromResult("")
```
rosen-vladimirov: I believe you no longer need this mock of `childProcess`

justcodebuilduser:
:broken_heart:

TsvetanMilanov:
run ci

justcodebuilduser:
:heart:

PR #812, opened 2016-09-27
Do not use proxy when npm config returns null

rosen-vladimirov:
In case there's no npm proxy, `npm config get proxy` returns string "null" and we think this is the proxy.
This leads to incorrect calls to registry.npmjs.org.


justcodebuilduser:
:broken_heart:

justcodebuilduser:
:broken_heart:

rosen-vladimirov:
run ci

justcodebuilduser:
:heart:

PR #813, opened 2016-09-28
Add timeout to http request and use it for calls to npm registry

rosen-vladimirov:
Add timeout option to http request which defines the time that we'll wait for response.
In case there's no response, the call will throw error.
Set timeout of 6000 for calls to registry.npmjs.org.

### DIFF
--- a/http-client.ts
+++ b/http-client.ts
@@ -11,7 +11,7 @@ export class HttpClient implements Server.IHttpClient {
 
 	constructor(private $logger: ILogger,
 		private $staticConfig: Config.IStaticConfig,
-		private $config: Config.IConfig) {}
+		private $config: Config.IConfig) { }
 
 	httpRequest(options: any, proxySettings?: IProxySettings): IFuture<Server.IResponse> {
 		return (() => {
@@ -48,7 +48,7 @@ export class HttpClient implements Server.IHttpClient {
 			options.headers = options.headers || {};
 			let headers = options.headers;
 
-			if(proxySettings || this.$config.USE_PROXY) {
+			if (proxySettings || this.$config.USE_PROXY) {
 				options.path = requestProto + "://" + options.host + options.path;
 				headers.Host = options.host;
 				options.host = (proxySettings && proxySettings.hostname) || this.$config.PROXY_HOSTNAME;
@@ -82,7 +82,16 @@ export class HttpClient implements Server.IHttpClient {
 				headers["Accept-Encoding"] = "gzip,deflate";
 			}
 
-			let result = new Future<Server.IResponse>();
+			let result = new Future<Server.IResponse>(),
+				timerId: number;
+
+			if (options.timeout) {
+				timerId = setTimeout(() => {
+					this.setResponseResult(result, timerId, { err: new Error(`Request to ${unmodifiedOptions.url} timed out.`) });
+				}, options.timeout);
+
+				delete options.timeout;
+			}
 
 			this.$logger.trace("httpRequest: %s", util.inspect(options));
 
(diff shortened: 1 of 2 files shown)


http-client.ts
```diff
@@ -175,6 +171,30 @@ export class HttpClient implements Server.IHttpClient {
 		}).future<Server.IResponse>()();
 	}
 
+	private setResponseResult(result: IFuture<Server.IResponse>, timerId: number, resultData: { response?: Server.IRequestResponseData, body?: string, err?: Error }): void {
+		if (timerId) {
+			clearTimeout(timerId);
+			timerId = null;
```
rosen-vladimirov: no need of this assignment

http-client.ts
```diff
@@ -82,7 +82,16 @@ export class HttpClient implements Server.IHttpClient {
 				headers["Accept-Encoding"] = "gzip,deflate";
 			}
 
-			let result = new Future<Server.IResponse>();
+			let result = new Future<Server.IResponse>(),
+				timerId: number;
+
+			if (options.timeout) {
+				timerId = setTimeout(() => {
+					this.setResponseResult(result, timerId, { err: new Error(`Request to ${unmodifiedOptions.url} timed out.`) });
```
Mitko-Kerezov: I believe you can shorten this call a bit like this:
`this.setResponseResult(result, timerId, { response });`

http-client.ts
```diff
@@ -107,12 +116,7 @@ export class HttpClient implements Server.IHttpClient {
 				if (pipeTo) {
 					pipeTo.on("finish", () => {
 						this.$logger.trace("httpRequest: Piping done. code = %d", response.statusCode.toString());
-						if(!result.isResolved()) {
-							result.return({
-								response: response,
-								headers: response.headers
```
Mitko-Kerezov: You can shorten this call too and you might want to consider renaming `responseBody` to simply `body` in order to shorten it even further

justcodebuilduser:
:heart:

http-client.ts
```diff
@@ -128,28 +132,20 @@ export class HttpClient implements Server.IHttpClient {
```
Mitko-Kerezov: Consider renaming `theError` to `err` - it's not a better name, but it'll allow for cleaner code IMO

http-client.ts
```diff
@@ -128,28 +132,20 @@ export class HttpClient implements Server.IHttpClient {
 						let responseBody = data.join("");
 
 						if (successful || isRedirect) {
-							if(!result.isResolved()) {
-								result.return({
-									body: responseBody,
```
Mitko-Kerezov: `(error: Error)` could be `(err: Error)` for cleaner code below

http-client.ts
```diff
@@ -11,7 +11,7 @@ export class HttpClient implements Server.IHttpClient {
 
 	constructor(private $logger: ILogger,
 		private $staticConfig: Config.IStaticConfig,
-		private $config: Config.IConfig) {}
+		private $config: Config.IConfig) { }
```
Mitko-Kerezov: shouldn't this type be `NodeJS.Timer`

rosen-vladimirov: nope, TypeScript defines the return result of setTimeout to number

http-client.ts
rosen-vladimirov: In fact it's better to keep it, the timerId is in fact NodeJS.Timer, but the TypeScript thinks it's number (as the result of setTimeout is set to number due to TypeScript definitions)

justcodebuilduser:
:heart:

rosen-vladimirov:
@Mitko-Kerezov all requested changed are applied

ndinev:
:+1:

Issue #723, opened 2016-06-21
Error messages not printed


enchev:
_From @jasssonpet on June 21, 2016 9:5_

_From @PanayotCankov on June 20, 2016 9:19_

In a blank application, the following `app.js`:
```
var application = require("application");
console.log("Here and now!");
throw new Error("Some fancy error");
```
Prints the following output correctly:
```
CONSOLE LOG file:///app/app.js:2:12: Here and now!
***** Fatal JavaScript exception - application has been terminated. *****
Native stack trace:
1   0x9bf2f -[TNSRuntime executeModule:]
2   0x61449 main
3   0x259f8873 <redacted>
JavaScript stack trace:
1   @file:///app/app.js:3:16
JavaScript error:
file:///app/app.js:3:16: JS ERROR Error: Some fancy error
```

While if the `console.log` call is omitted:
```
var application = require("application");
// console.log("Here and now!");
throw new Error("Some fancy error");
```
The error message is not printed:
```
NativeScript loaded bundle file:///System/Library/Frameworks/MapKit.framework
NativeScript loaded bundle file:///System/Library/Frameworks/SceneKit.framework
NativeScript loaded bundle file:///System/Library/Frameworks/QuickLook.framework
NativeScript loaded bundle file:///System/Library/Frameworks/CoreSpotlight.framework
***** Fatal JavaScript exception - application has been terminated. *****
Native stack trace:
1   0x66f2f -[TNSRuntime executeModule:]
2   0x2c449 main
3   0x259f8873 <redacted>
JavaScript stack trace:
1   @file:///app/app.js:2:16
JavaScript error:
```

_Copied from original issue: NativeScript/ios-runtime#583_

_Copied from original issue: NativeScript/nativescript-cli#1867_

enchev:
_From @jasssonpet on June 21, 2016 9:5_

The JavaScript error of the runtime crash log is actually being printed to the standard error stream as can be seen from `idevicesyslog` for example:

```
Jun 21 10:35:29 NativeScripts-iPhone-6 TestApp3[885] <Notice>: 1   0x1000e3434 -[TNSRuntime executeModule:]
Jun 21 10:35:29 NativeScripts-iPhone-6 TestApp3[885] <Notice>: 2   0x1000a03e4 main
Jun 21 10:35:29 NativeScripts-iPhone-6 TestApp3[885] <Notice>: 3   0x1811a68b8 <redacted>
[...]

PR #814, opened 2016-09-28
Get full ios logs on mac

TsvetanMilanov:
When we use net.Socket we need to use .toString() to get the whole content of the buffer because it may contain many \0 and ref.readCString will return the string which is before the first \0 not the whole content.

Fixes https://github.com/telerik/mobile-cli-lib/issues/723


rosen-vladimirov:
👍

justcodebuilduser:
:heart:

PR #816, opened 2016-10-03
Show helpful information when Sync over WIFI error is detected

rosen-vladimirov:
iTunes has an option to sync changes over wifi for each iOS device. However, for some devices, when trying to start the installation_proxy service and `Sync over Wi-Fi` option is enabled, the connection fails.
Catch such cases and show helpful message describing the problem and how to resolve it.

Should help for issues like: https://github.com/NativeScript/nativescript-cli/issues/1398


justcodebuilduser:
:heart:

PR #819, opened 2016-10-06
Fix validation of -- options

rosen-vladimirov:
The validation of -- options passed on the command line depends on yargs module. Update it to latest versions as the previously used one didn't set correct values of options marked as strings.
Also in our code we pass object of valid options to yargs. Options which have `-` in the name are treated in a special manner in yargs. However instead of passing them with `-`, we pass their secondary representation (remove the `-` and capitalize the next letter).
This way, when the user passes option with `-`, yargs treat it with it's default behavior and does not respect our options for it (as we have not defined that we have option with `-`).
Example:
```JavaScript
var yargs = require("yargs");
var opts = {
  "profile-dir": { type: "string" }
};
console.log(yargs(process.argv).options(opts).argv);

var opts1 = {
  "profileDir": { type: "string" }
};
console.log(yargs(process.argv).options(opts1).argv);
```

When called with:
```
node testYargs.js --profile-dir
```
The result is:
```
{
  'profile-dir': '',
  profileDir: '',
  '$0': 'testYargs.js'
}

{
  'profile-dir': true,
  profileDir: true,
  '$0': 'testYargs.js'
}
```

As you can see, passing "profileDir" and defining it as string does not work. But when we use "profile-dir" in our opts, everything works as expected and yargs sets the value to empty string.

In order to resolve the problem, make sure we pass options with dashes to yargs.

### Fix isNullOrWhitespace method
When boolean value is passed to isNullOrWhitespace helper method, it fails as true.replace is not available function.

### Add unit tests
Add unit tests for:
 - options issue - assert correct behavior when dashed option is passed on the terminal
 - helpers isNullOrWhitespace method
 - helpers isBoolean method

### DIFF
--- a/helpers.ts
+++ b/helpers.ts
@@ -114,7 +114,7 @@ export function isInteractive(): boolean {
 }
 
 export function toBoolean(str: any) {
-	return str && str.toString().toLowerCase() === "true";
+	return !!(str && str.toString().toLowerCase() === "true");
 }
 
 export function block(operation: () => void): void {
@@ -135,12 +135,12 @@ export function fromWindowsRelativePathToUnix(windowsRelativePath: string): stri
 	return windowsRelativePath.replace(/\\/g, "/");
 }
 
-export function isNullOrWhitespace(input: string): boolean {
-	if (!input) {
+export function isNullOrWhitespace(input: any): boolean {
+	if (!input && input !== false) {
 		return true;
 	}
 
-	return input.replace(/\s/gi, "").length < 1;
+	return _.isString(input) && input.replace(/\s/gi, "").length < 1;
 }
 
 export function getCurrentEpochTime(): number {
@@ -186,8 +186,7 @@ export function trimSymbol(str: string, symbol: string) {
 	return str;
 }
 
-// TODO: Use generic for predicatе predicate: (element: T|T[]) when TypeScript support this.
-export function getFuturesResults<T>(futures: IFuture<T | T[]>[], predicate: (element: any) => boolean): T[] {
+export function getFuturesResults<T>(futures: IFuture<T | T[]>[], predicate: (element: T|T[]) => boolean): T[] {
 	Future.wait(futures);
 	return _(futures)
 		.map(f => f.get())
--- a/helpers.ts
+++ b/helpers.ts
@@ -113,8 +113,8 @@ export function isInteractive(): boolean {
 	return process.stdout.isTTY && process.stdin.isTTY;
 }
 
-export function toBoolean(str: any) {
-	return !!(str && str.toString().toLowerCase() === "true");
+export function toBoolean(str: any): boolean {
+	return !!(str && str.toString && str.toString().toLowerCase() === "true");
 }
 
 export function block(operation: () => void): void {
@@ -186,7 +186,8 @@ export function trimSymbol(str: string, symbol: string) {
 	return str;
 }
 
-export function getFuturesResults<T>(futures: IFuture<T | T[]>[], predicate: (element: T|T[]) => boolean): T[] {
+// TODO: Use generic for predicatе predicate: (element: T|T[]) when TypeScript support this.
+export function getFuturesResults<T>(futures: IFuture<T | T[]>[], predicate: (element: any) => boolean): T[] {
 	Future.wait(futures);
 	return _(futures)
 		.map(f => f.get())
--- a/options.ts
+++ b/options.ts
@@ -1,4 +1,3 @@
-import * as util from "util";
 import * as helpers from "./helpers";
 import * as yargs from "yargs";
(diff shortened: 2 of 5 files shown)


justcodebuilduser:
:heart:

helpers.ts
```diff
@@ -114,7 +114,7 @@ export function isInteractive(): boolean {
 }
 
 export function toBoolean(str: any) {
```
rosen-vladimirov: set return type

test/unit-tests/helpers.ts
```diff
@@ -134,19 +140,125 @@ describe("helpers", () => {
...
+				expectedResult: false
+			},
+			{
+				input: true,
+				expectedResult: false
+			},
+			{
+				input: false,
+				expectedResult: false
+			}
```
Mitko-Kerezov: You could add a couple more cases like 
```JavaScript
{
    input: '\n',
    expectedResult: false
},
{
    input: '\r\n',
    expectedResult: false
},
{
    input: '\t',
    expectedResult: false
}
```

In order to ensure that the regex won't be changed for example from `/\s/` to simply `/ /`

helpers.ts
```diff
@@ -186,8 +186,7 @@ export function trimSymbol(str: string, symbol: string) {
 	return str;
 }
 
-// TODO: Use generic for predicatе predicate: (element: T|T[]) when TypeScript support this.
-export function getFuturesResults<T>(futures: IFuture<T | T[]>[], predicate: (element: any) => boolean): T[] {
+export function getFuturesResults<T>(futures: IFuture<T | T[]>[], predicate: (element: T|T[]) => boolean): T[] {
```
Mitko-Kerezov: 🎉 🎈 🎊

options.ts
```diff
@@ -13,12 +13,12 @@ export class OptionType {
 export class OptionsBase {
 	private optionsWhiteList = ["ui", "recursive", "reporter", "require", "timeout", "_", "$0"]; // These options shouldn't be validated
 	public argv: IYargArgv;
-	private static GLOBAL_OPTIONS: IDictionary<IDashedOption> = {
+	private GLOBAL_OPTIONS: IDictionary<IDashedOption> = {
```
Mitko-Kerezov: Instead of setting this property to dynamic in favor of static, you could assign
`OptionsBase.GLOBAL_OPTIONS.profileDir.default = this.defaultProfileDir`
in the constructor.

This is only a minor suggestion of course as the benefit would be next to none so definitely not a merge-stopper

options.ts
```diff
@@ -106,7 +106,7 @@ export class OptionsBase {
 				return;
 			}
 
-			let optionName = this.getCorrectOptionName(originalOptionName);
+			let optionName = this.getNonDashedOptionName(originalOptionName);
```
Mitko-Kerezov: AFAIK this leads to a minor change in the behavior. Is this change correct?

options.ts
```diff
@@ -154,36 +154,49 @@ export class OptionsBase {
...
 			let secondaryOptionName = util.format("%s%s%s", matchUpperCaseLetters[1], matchUpperCaseLetters[3].toUpperCase(), matchUpperCaseLetters[4] || '');
-			return this.getSecondaryOptionName(secondaryOptionName);
+			return this.getNonDashedOptionName(secondaryOptionName);
+		}
+
+		return optionName;
+	}
+
+	private getDashedOptionName(optionName: string): string {
+		let matchUpperCaseLetters = optionName.match(/(.+?)([A-Z])(.*)/);
```
Mitko-Kerezov: You can assign the regex to a `static` variable so that it's not created each time we enter this method. This is valid for `getNonDashedOptionName` too

options.ts
```diff
@@ -154,36 +154,49 @@ export class OptionsBase {
 	// ex, "$ <cli name> emulate android --profile-dir" will add profile-dir to yargs.argv as profile-dir and profileDir
 	// IMPORTANT: In your code, it is better to use the value without dashes (profileDir in the example).
 	// This way your code will work in case "$ <cli name> emulate android --profile-dir" or "$ <cli name> emulate android --profileDir" is used by user.
-	private getSecondaryOptionName(optionName: string): string {
+	private getNonDashedOptionName(optionName: string): string {
 		let matchUpperCaseLetters = optionName.match(/(.+?)([-])([a-zA-Z])(.*)/);
```
Mitko-Kerezov: Off-topic:
We don't need to capture `[-]` in a group.

options.ts
rosen-vladimirov: reverting this change, it's by mistake

justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

PR #820, opened 2016-10-10
LiveSync to /Documents

ventsislav-georgiev:
Related to: [#322738](http://teampulse.telerik.com/view#item/322738)

### DIFF
--- a/definitions/mobile.d.ts
+++ b/definitions/mobile.d.ts
@@ -526,7 +526,7 @@ declare module Mobile {
 	}
 
 	interface IHouseArrestClient {
-		getAfcClientForAppContainer(applicationIdentifier: string): Mobile.IAfcClient;
+		getAfcClientForAppDocuments(applicationIdentifier: string): Mobile.IAfcClient;
 		closeSocket(): void;
 	}
 
--- a/definitions/mobile.d.ts
+++ b/definitions/mobile.d.ts
@@ -526,6 +526,7 @@ declare module Mobile {
 	}
 
 	interface IHouseArrestClient {
+		getAfcClientForAppContainer(applicationIdentifier: string): Mobile.IAfcClient;
 		getAfcClientForAppDocuments(applicationIdentifier: string): Mobile.IAfcClient;
 		closeSocket(): void;
 	}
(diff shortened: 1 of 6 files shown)


justcodebuilduser:
:heart:

e2l3n:
LGTM :+1:

definitions/mobile.d.ts
```diff
@@ -526,7 +526,7 @@ declare module Mobile {
 	}
 
 	interface IHouseArrestClient {
-		getAfcClientForAppContainer(applicationIdentifier: string): Mobile.IAfcClient;
+		getAfcClientForAppDocuments(applicationIdentifier: string): Mobile.IAfcClient;
```
rosen-vladimirov: Can we introduce new method instead of renaming the current one? We'll need both of them

mobile/ios/device/ios-device-file-system.ts
```diff
@@ -72,19 +72,19 @@ export class IOSDeviceFileSystem implements Mobile.IDeviceFileSystem {
 
 	public deleteFile(deviceFilePath: string, appIdentifier: string): void {
 		let houseArrestClient: Mobile.IHouseArrestClient = this.$injector.resolve(iOSProxyServices.HouseArrestClient, {device: this.device});
-		let afcClientForContainer = houseArrestClient.getAfcClientForAppContainer(appIdentifier);
-		afcClientForContainer.deleteFile(deviceFilePath);
+		let afcClientForAppDocuments = houseArrestClient.getAfcClientForAppDocuments(appIdentifier);
+		afcClientForAppDocuments.deleteFile(deviceFilePath);
```
rosen-vladimirov: This code will not work in NativeScript CLI. The files there are always inside `/Library/Application Support/` directory, so this code will not be able to delete them. 
Maybe we should check the passed `deviceFilePath` and decide if we should use `getAfcClientForAppDocuments` or `getAfcClientForAppContainer`

mobile/ios/device/ios-device-file-system.ts
```diff
@@ -72,19 +72,19 @@ export class IOSDeviceFileSystem implements Mobile.IDeviceFileSystem {
...
 		houseArrestClient.closeSocket();
 	}
 
 	public transferFiles(deviceAppData: Mobile.IDeviceAppData, localToDevicePaths: Mobile.ILocalToDevicePathData[]): IFuture<void> {
 		return (() => {
 			let houseArrestClient: Mobile.IHouseArrestClient = this.$injector.resolve(iOSProxyServices.HouseArrestClient, { device: this.device });
-			let afcClientForAppContainer = houseArrestClient.getAfcClientForAppContainer(deviceAppData.appIdentifier);
+			let afcClientForAppDocuments = houseArrestClient.getAfcClientForAppDocuments(deviceAppData.appIdentifier);
```
rosen-vladimirov: Same as above - in NativeScript CLI this code will not work. Maybe we can use localToDevicePaths in order to understand if we should use `getAfcClientForAppContainer` or `getAfcClientForAppDocuments`

justcodebuilduser:
:heart:

PR #821, opened 2016-10-10
if there is no app folder just return an empty array of files

Plamen5kov:
currently if there is no given folder the line:
```typescript
let contents = this.readDirectory(directoryPath).wait();
```
blows up because there is no directory. Currently if directory present, the function will return an empty array.

ping: @rosen-vladimirov


justcodebuilduser:
:heart:

PR #823, opened 2016-10-12
Fixed: Ctrl+C should work in cases when there are no listeners attached

tzraikov:
Ctrl+C should work in cases when there are no listeners for exit signals attached


rosen-vladimirov:
👍  after green build

justcodebuilduser:
:heart:

PR #825, opened 2016-10-13
Fixed: Incorrect behavior when uncaught exception of type NS Exceptio…

tzraikov:
Fixed: Incorrect behavior when uncaught exception of type NS Exception is thrown and the phone is connected with cable

Fixes: https://github.com/NativeScript/nativescript-cli/issues/2053


justcodebuilduser:
:heart:

PR #829, opened 2016-10-19
Display a message when starting the devices-service to indicate what is happening

tzraikov:


justcodebuilduser:
:heart:

rosen-vladimirov:
Closing this as the initialize method is used in some async cases, where we should not modify the stdout (in case it's not error). Instead, it's better to set the logging before calling `$devicesService.initialize()`

PR #835, opened 2016-11-04
Fix CLI is not disconnected on iOS exception

KristinaKoeva:
The code used to read only the first digit, but according to the documentation the signal code is a two-digit hexadecimal number
With the current version of the runtime the signal that is raised is SIGABRT == 0x06

Fixes https://github.com/NativeScript/nativescript-cli/issues/2053

### DIFF
--- a/mobile/ios/device/ios-core.ts
+++ b/mobile/ios/device/ios-core.ts
@@ -1137,11 +1137,18 @@ class GDBSignalWatcher extends stream.Writable {
 
 	public _write(packet: any, encoding: string, callback: Function) {
 		try {
-			for (let i = 0; i < packet.length - 2; i++) {
-				if (packet[i] === getCharacterCodePoint("$") && (packet[i + 1] === getCharacterCodePoint("T") || packet[i + 1] === getCharacterCodePoint("S"))) {
-					// SIGKILL || SIGABRT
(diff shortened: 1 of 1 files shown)


justcodebuilduser:
:heart:

mobile/ios/device/ios-core.ts
```diff
@@ -1137,11 +1137,15 @@ class GDBSignalWatcher extends stream.Writable {
 
 	public _write(packet: any, encoding: string, callback: Function) {
 		try {
-			for (let i = 0; i < packet.length - 2; i++) {
+			// The reply packages take the following form (the space in the reply templates is included for clarity)
+			// ‘S AA’  or ‘T AA n1:r1;n2:r2;...’ meaning that the program received signal number AA (a two-digit hexadecimal number)
+			for (let i = 0; i < packet.length - 3; i++) {
 				if (packet[i] === getCharacterCodePoint("$") && (packet[i + 1] === getCharacterCodePoint("T") || packet[i + 1] === getCharacterCodePoint("S"))) {
```
rosen-vladimirov: Can you move the values of `$`, `T` and `S` code points to a constants outside of the `for` as currently they are evaluated on each iteration

justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

PR #837, opened 2016-11-10
Fix cannot find module ios-log-filter

TsvetanMilanov:
Change the path to ios-log-filter to point to the correct folder.


Mitko-Kerezov:
👍

justcodebuilduser:
:heart:

PR #843, opened 2016-11-27
Transpile to ES6

rosen-vladimirov:
Add support for ES6 transpilation. Change dependency injection in order to match the `class` and `constructor` keywords.
Update istanbul to latest version as its previous versions do not support ES6.

Remove bluebird.d.ts as it conflicts with Promises from TypeScript's ES6 d.ts.
Remove bluebird as dependency - we will use native Promises from now on.

Add support for Node.js 4 and 5 - both versions have limited support for ES6 features, so in order to support them we have to:
 - pass --harmony flag to Node.js
 - do not use some features like default values of method arguments.
 - do not use spread operator - it's implementation in Node.js 4 is limited only to arrays. Even this implementation has issues. So use `.apply` instead of using spread operator. Remove all other usages of spread operator (for objects).

Add tests scripts for istanbul and mocha, as they both will need the `--harmony` flag as well, so we cannot execute them directly with Node.js 4 and 5. Instead we'll execute a custom script, that will start a new node process.


rosen-vladimirov:
run ci

justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

test-scripts/istanbul.js
```diff
@@ -0,0 +1,18 @@
+"use strict";
+
+const childProcess = require("child_process");
+const path = require("path");
+const pathToIstanbul = path.join(__dirname, "..", "node_modules", "istanbul", "lib", "cli.js");
+const pathToMocha = path.join(__dirname, "..", "node_modules", "mocha", "bin", "_mocha");
+
+const istanbulArgs = [ pathToIstanbul, "cover", pathToMocha ];
```
rosen-vladimirov: I feel more comfortable this way 😃

justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

PR #845, opened 2016-11-29
Fix wrong appid when LiveSyncing to Companion app

Mitko-Kerezov:
Instead of using the application's own appid we should use the companion app's appid when performing LiveSync for the companion


Mitko-Kerezov:

justcodebuilduser:
:heart:

PR #846, opened 2016-12-01
Use TypeScript 2.0

TsvetanMilanov:
Update the TS related dependencies to use TS 2.0 and fix the errors from the update.

TS 2.0 does not respect our .abreferences.d.ts file because it is hidden and we have to rename it to abreferences.d.ts.


justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

AssiNET:
👍

PR #850, opened 2016-12-05
Fix iOS 10 device logs

TsvetanMilanov:
On iOS 10 devices the device logs contain \n\0 after each line.
When we use readCString for buffers which contain many \0
the method will return the content of the buffer only before the first \0 character.
We need to replace the \0 with "" in order read the whole content.

Approved in https://github.com/telerik/mobile-cli-lib/pull/848.

Fixes: http://teampulse.telerik.com/view#item/325955 and http://teampulse.telerik.com/view#item/326467


justcodebuilduser:
:heart:

justcodebuilduser:
:broken_heart:

justcodebuilduser:
:heart:

PR #853, opened 2016-12-13
Use sync api of fs

rosen-vladimirov:
Several methods were declared as `IFuture<T>` as in their body they are calling async methods from `fs` (for example `$fs.exists(...).wait()`). As now most of these methods are sync, convert all possible methods to sync (instead of IFuture).


justcodebuilduser:
:heart:

PR #856, opened 2016-12-15
Remove npm as dependency

rosen-vladimirov:
Npm as dependency had not been used. The scenario for using it in Proton is no longer on the roadmap, so remove it as it causes too slow installation and in most of the cases - fail on Windows with path too long exception.

Set version to 0.22.0


justcodebuilduser:
:heart:

PR #860, opened 2016-12-19
Fix iOS operations

rosen-vladimirov:
Fix iOS operations when xml protocol is used. Due to changes in the Node API we are using, all iOS related operations which use XML protocol, started failing.
Use correct way to parse the result.


Mitko-Kerezov:
Looks good 👍

justcodebuilduser:
:heart:

PR #861, opened 2016-12-19
added error handling on shelljs operations where necessary

Plamen5kov:
Adds error handling where necessary, because shelljs is now throwing errors by default.

related to: https://github.com/NativeScript/nativescript-cli/pull/2364

### DIFF
--- a/mobile/mobile-core/android-process-service.ts
+++ b/mobile/mobile-core/android-process-service.ts
@@ -11,7 +11,8 @@ export class AndroidProcessService implements Mobile.IAndroidProcessService {
 		private $staticConfig: Config.IStaticConfig,
 		private $injector: IInjector,
 		private $net: INet,
-		private $processService: IProcessService) {
+		private $processService: IProcessService,
+		private $logger: ILogger) {
 		this._devicesAdbs = {};
 		this._forwardedLocalPorts = [];
 	}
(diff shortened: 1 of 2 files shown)


justcodebuilduser:
:heart:

mobile/mobile-core/android-process-service.ts
```diff
@@ -285,7 +286,12 @@ export class AndroidProcessService implements Mobile.IAndroidProcessService {
 				// We need to use shelljs here instead of $adb because listener functions of exit, SIGINT and SIGTERM must only perform synchronous operations.
 				// The Node.js process will exit immediately after calling the 'exit' event listeners causing any additional work still queued in the event loop to be abandoned.
 				// See the official documentation for more information and examples - https://nodejs.org/dist/latest-v6.x/docs/api/process.html#process_event_exit.
-				shelljs.exec(`adb forward --remove tcp:${port}`);
+
+				try {
+					shelljs.exec(`adb forward --remove tcp:${port}`);
```
TsvetanMilanov: This code is used only in Proton and here `shelljs` will not throw error and we do not need this `try/catch`

test/unit-tests/yok.ts
```diff
@@ -601,9 +601,12 @@ describe("yok", () => {
 	});
 
 	function deleteDirectory(directory: string): boolean {
-		shelljs.rm("-rf", directory);
-		let err = shelljs.error();
-		return err === null;
+		try {
+			shelljs.rm("-rf", directory);
```
TsvetanMilanov: I think here `shelljs.config.fatal` will be `false`. Can you check and if I am right we must keep the old logic.

Plamen5kov: I changed it to `shelljs.config.fatal = true` both places: https://github.com/NativeScript/nativescript-cli/pull/2364/commits/ae4937d1dada45e5e4204ed9f9c34bd78518c1a3#diff-95ad435cc796c681eb42182a4f10b038R6 and https://github.com/NativeScript/nativescript-cli/pull/2364/commits/ae4937d1dada45e5e4204ed9f9c34bd78518c1a3#diff-c095b997078fae6c5c43a57489e046b8R3

justcodebuilduser:
:heart:

test/unit-tests/yok.ts
TsvetanMilanov: OK. This is valid for NativeScript CLI, but not for AppBuilder CLI. I think you can use the old logic and the new one. This way this code will work in both CLI. You just have to return `shelljs.error() === null` after the `catch`. This way if shelljs throws an error it will return `false` in the `catch` and if it doesn't throw but there is an error it will still return `false`.

justcodebuilduser:
:heart:

PR #862, opened 2016-12-27
Add settlePromises method

rosen-vladimirov:
Add settlePromises method that will execute all promises instead of failing when the first promise is rejected. It will be used to replace `Future.wait(futures)`.
The method returns new promise that will be resolved when all promises are resolved. In case any of them is rejected, after all promises are completed, the returned promise will be rejected with an error, describing all errors during the execution.

### DIFF
--- a/helpers.ts
+++ b/helpers.ts
@@ -3,7 +3,42 @@ import * as Fiber from "fibers";
 import * as net from "net";
 let Table = require("cli-table");
 import Future = require("fibers/future");
-import { platform } from "os";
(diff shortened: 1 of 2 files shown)


justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

Mitko-Kerezov:
👍

PR #863, opened 2016-12-27
Implemented get/put file functionality in device file systems

tzraikov:
This functionality was necessary in order to implement predictable livesync as described here:

https://github.com/NativeScript/nativescript-cli/issues/2153
https://github.com/NativeScript/nativescript-cli/issues/2330

* Removed the watch option from options list. This option is specific for {N} CLI.

### DIFF
--- a/mobile/android/android-device-file-system.ts
+++ b/mobile/android/android-device-file-system.ts
@@ -23,12 +23,31 @@ export class AndroidDeviceFileSystem implements Mobile.IDeviceFileSystem {
 		return this.adb.executeShellCommand(listCommandArgs);
 	}
 
-	public getFile(deviceFilePath: string): IFuture<void> {
-		return Future.fromResult();
-	}
+	public getFile(deviceFilePath: string, outputPath?: string): IFuture<void> {
+		return (() => {
+			let stdout = !outputPath;
+			if (stdout) {
+				temp.track();
+				outputPath = temp.path({prefix: "sync", suffix: ".tmp"});
+			}
+			this.adb.executeCommand(["pull", deviceFilePath, outputPath]).wait();
+			if (stdout) {
+				let readStream = this.$fs.createReadStream(outputPath);
+				let future = new Future<void>();
+				readStream.pipe(process.stdout);
+				readStream.on("end", () => {
+					future.return();
+				});
+				readStream.on("error", (err: Error) => {
+					future.throw(err);
+				});
+				future.wait();
+			}
+		}).future<void>()();
+ 	}
 
 	public putFile(localFilePath: string, deviceFilePath: string): IFuture<void> {
-		return Future.fromResult();
+		return this.adb.executeCommand(["push", localFilePath, deviceFilePath]);
 	}
 
 	public transferFiles(deviceAppData: Mobile.IDeviceAppData, localToDevicePaths: Mobile.ILocalToDevicePathData[]): IFuture<void> {
--- a/mobile/ios/device/ios-device-file-system.ts
+++ b/mobile/ios/device/ios-device-file-system.ts
@@ -2,6 +2,7 @@ import * as iOSProxyServices from "./ios-proxy-services";
 import * as path from "path";
 import * as ref from "ref";
 import * as util from "util";
+import Future = require("fibers/future");
 
 export class IOSDeviceFileSystem implements Mobile.IDeviceFileSystem {
 	constructor(private device: Mobile.IiOSDevice,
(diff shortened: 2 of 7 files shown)


justcodebuilduser:
:heart:

mobile/android/android-device-file-system.ts
```diff
@@ -23,12 +24,35 @@ export class AndroidDeviceFileSystem implements Mobile.IDeviceFileSystem {
...
 	}
 
-	public getFile(deviceFilePath: string): IFuture<void> {
-		return Future.fromResult();
+	public getFile(deviceFilePath: string, outputPath?: string): IFuture<void> {
+		return (() => {
+			let stdout = false;
+			if (!outputPath) {
+				let fileName = path.basename(deviceFilePath);
+				outputPath = path.join(shelljs.tempdir(), this.$fs.getUniqueFileName(fileName));
```
rosen-vladimirov: you can use `temp.path()` to get a unique filename in the temp dir.

mobile/android/android-device-file-system.ts
```diff
@@ -23,12 +24,35 @@ export class AndroidDeviceFileSystem implements Mobile.IDeviceFileSystem {
...
+				stdout = true;
+			}
+			this.adb.executeCommand(["pull", deviceFilePath, outputPath]).wait();
+			if (stdout) {
+				let fileToRead = this.$fs.createReadStream(outputPath);
+				let fileToWrite = process.stdout;
```
rosen-vladimirov: Instead of using `while`, you can just pipe the code to readable stream to `process.stdout`, for example:
```TypeScript
const future = new Future<void>();
readStream.pipe(process.stdout);
readStream.on("end", () => {
	future.return();
});
readStream.on("error", (err: Error) => {
	future.throw(err);
});

future.wait();
```

mobile/ios/device/ios-device-file-system.ts
```diff
@@ -42,27 +43,32 @@ export class IOSDeviceFileSystem implements Mobile.IDeviceFileSystem {
...
+		let fileToRead = afcClient.open(deviceFilePath, "r");
+		let fileToWrite = outputFilePath ? this.$fs.createWriteStream(outputFilePath) : process.stdout;
+		if (outputFilePath) {
+			fileToWrite.on("close", () => {
+				future.return();
+			});
+		}
+		let dataSizeToRead = 8192;
+		let size = 0;
+		while(true) {
```
rosen-vladimirov: Same as above, you can pipe the readable stream to writeable stream:
Instead of using `while`, you can just pipe the code to readable stream to `process.stdout`, for example:
```TypeScript
const future = new Future<void>();
let afcClient = this.resolveAfc();
let fileToRead = afcClient.open(deviceFilePath, "r");
let fileToWrite = outputFilePath ? this.$fs.createWriteStream(outputFilePath) : process.stdout;

fileToRead.pipe(fileToWrite);
fileToRead.on("end", () => {
	future.return();
});
fileToRead.on("error", (err: Error) => {
	future.throw(err);
});

return future.
```

options.ts
```diff
@@ -50,7 +50,6 @@ export class OptionsBase {
 		return {
 			all: { type: OptionType.Boolean },
 			json: { type: OptionType.Boolean },
-			watch: { type: OptionType.Boolean },
```
rosen-vladimirov: Personally I think it's better to return `stream.Readable` from `getFile` method and allow the consumers to decide if the output should be in a file or on stdout. This will allow calling the method from the code and using the stream directly. Consider the case where you need to check the content of a file on device. The current implementation will download the file, so you'll have to read it again in the code. In case you return a Readable stream, you can check the content directly. 
The command (`tns device get-file`) can check the passed arguments and decide whether the content should be shown on the terminal or saved in a file.

However this is just an opinion and I do not insist to implement it this way.

rosen-vladimirov:
Personally I think it's better to return `stream.Readable` from `getFile` method and allow the consumers to decide if the output should be in a file or on stdout. This will allow calling the method from the code and using the stream directly. Consider the case where you need to check the content of a file on device. The current implementation will download the file, so you'll have to read it again in the code. In case you return a Readable stream, you can check the content directly. 
The command (`tns device get-file`) can check the passed arguments and decide whether the content should be shown on the terminal or saved in a file.

However this is just an opinion.

mobile/android/android-device-file-system.ts
rosen-vladimirov: In case you use `temp.path()` in combination with `temp.track()`, the downloaded file will be automatically removed after the process termination. This way users's temp directories will be cleaned from our files.

options.ts
tzraikov: I tought this is used only in CLI. The watch option is back.

justcodebuilduser:
:heart:

mobile/android/android-device-file-system.ts
tzraikov: hmm, is temp.path better than getUniqueFileName?

tzraikov: done

tzraikov: 👍

mobile/android/android-device-file-system.ts
tzraikov: done 👍

mobile/ios/device/ios-device-file-system.ts
tzraikov: This will not work because afcClient.open doesn't return a stream.

tzraikov:
Well, I agree, that would be great.. probably. However, it is more appropriate to be a part of a future refactoring. Currently I am more focused on a major task, namely - predictable livesync and this is just a fix of a broken API.

PS: It works now with stdout.

justcodebuilduser:
:broken_heart:

justcodebuilduser:
:broken_heart:

justcodebuilduser:
:heart:

options.ts
tzraikov: Note that I changed its default value to **true** because this should be its default value in CLI now. Is this OK with appbuilder?

tzraikov: Note that I changed its default value to true because this should be its default value in CLI now. Is this OK with appbuilder?

justcodebuilduser:
:broken_heart:

tzraikov:
run ci

justcodebuilduser:
:broken_heart:

justcodebuilduser:
:heart:

PR #867, opened 2017-01-05
added logging when executing system commands

Plamen5kov:
Cli executes system commands, but when they are not critical for the system the OS works fine, but cli fails because it can't find the necessary commands it expects to be there by default. Added some logging for easier tracing in the future. 
related: https://github.com/NativeScript/nativescript-cli/issues/1922


justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

Plamen5kov:
@rosen-vladimirov the changes you requested are done.

PR #864, opened 2016-12-28
Fixed: console.log outputs on the same line in android

tzraikov:

PR #868, opened 2017-01-09
Fix unit tests on Windows

rosen-vladimirov:
On Windows the unit tests are failing as the end of line is `\r\n`, but android-log-filter tests expect only `\n`.
Use `os.EOL` instead of `\n`, so the tests will pass on all OS.

Issue is introduced in https://github.com/telerik/mobile-cli-lib/pull/864


justcodebuilduser:
:heart:

PR #871, opened 2017-01-10
Fix problem with non-existent cli file

rosen-vladimirov:
CLI uses special file ("cli") in the TEMP in order to detect when the process should be killed in cases a new installation is in progress.
However it looks like Nodejs chmodSync sometimes fails on Windows with ENOENT error.
As the chmod operation is required only for UNIX and Linux based OSes, do not try to execute it on Windows.


rosen-vladimirov:
Approved here: https://github.com/telerik/mobile-cli-lib/pull/870

justcodebuilduser:
:heart:

PR #873, opened 2017-01-13
Fix cache decorator

TsvetanMilanov:
We need to cache the result of the decorated methods for each instance not for all instances. Also we need to add the name of the cached member to the cache property becasue if we do not add it we can't cache other members.

Also we need to execute build and deploy on multiple devices synchronously because the build will fail if the actions are executed for each device in parallel.

There were forgotten console.logs from debugging which are removed.


justcodebuilduser:
:heart:

PR #875, opened 2017-01-16
Make isEmulator a get property

Mitko-Kerezov:
Instead of a function - it doesn't perform any slow operations and can be a get property.
isEmulator was previously a property but was subsequently re-written as a function, however this is no longer necessary.


mobile/ios/device/ios-proxy-services.ts
```diff
@@ -269,6 +269,7 @@ export class InstallationProxyClient {
 		private $errors: IErrors) { }
 
 	public async deployApplication(packageFile: string): Promise<void> {
+		debugger;
```
rosen-vladimirov: no

justcodebuilduser:
:broken_heart:

justcodebuilduser:
:heart:

PR #877, opened 2017-01-18
Fix prompter

TsvetanMilanov:
After removing the fibers we forgot to await the result from the prompt before calling the unmuteStdout method. After this method is invoked the input from the terminal is not accepted.


justcodebuilduser:
:broken_heart:

justcodebuilduser:
:heart:

PR #879, opened 2017-01-19
Add default message when a node.js version is deprecated

rosen-vladimirov:
In method `verifyNodeVersion`, which is called at the beginning of each CLI, sometimes we have to inform the user that we'll stop support of specific range of Node.js versions in the next release.
So add the logic in the method and each CLI can specify deprecated versions in case it  has to.


justcodebuilduser:
:heart:

PR #880, opened 2017-01-20
Fix livesync and unable to resolve callstack

rosen-vladimirov:
Fix livesync errors

`deviceProjectRootPath` had been renamed to `getDeviceProjectRootPath` in one of previous commits. It's been fixed in the base class, but the classes that extend the base class were still using the old name.
Rename the method everywhere.

Fix `unable to resolve callstack error`

When an error is raised we are trying to resolve the callstack, but there's no source map for TypeScript generated methods for `async` (`__awaiter`). This breaks our logic. In such case just print the line in `.js` instead of `.ts` file.
Remove `source-map.js` from `vendor` and use the npm package.


justcodebuilduser:
:heart:

PR #881, opened 2017-01-20
Change the resolve constructor logic in yok

TsvetanMilanov:
Right now we can't resolve classes with names starting with small letter. With this change we keep the logic for resolving hooks and hierarchical commands.


justcodebuilduser:
:heart:

TsvetanMilanov:
Closing this because @rosen-vladimirov was right - some hooks are not working.

PR #883, opened 2017-01-24
Use node.d.ts from @types

rosen-vladimirov:
Currently we are using `node.d.ts` from version `0.11.x`. In order to get latest updates, remove our own version of node.d.ts and use the one from `@types`. Use the version for Node 6.
Fix code according to new definitions. Introduce ICliGlobal interface in order to describe the global object that's used from CLI (we add some properties to the global object).
Add definition file for `xmlhttprequest` module.
Add definitions for Eqatec Analytics.


justcodebuilduser:
:heart:

PR #884, opened 2017-01-25
Fix ensure all platform assets

TsvetanMilanov:
We do not need to add .png asset when the project has the 9 patch png version of this asset.


justcodebuilduser:
:heart:

ndinev:
:+1:

PR #885, opened 2017-01-31
Fixes for devices publicAPI

rosen-vladimirov:
### Allow re-require of injector dependencies

By default CLI does not allow requiring injector dependencies more than once, i.e. you cannot use:
```TypeScript
$injector.require("a", "path to file where a is registered");
$injector.require("a", "path to different file where a is registered");
```

However due to extenisbility model requirements we would like to be able to change some of the dependencies. Introduce new property in IInjector that will allow redefining of the modules.
Its default value is false, so we'll keep current behavior.

### Stop calling initialize automatically when resolving module from Public API
When resolving some module from Public API, CLI's yok automatically calls its initialize method. However this is no longer possible as it will require async getter. Remove the magic - the `initialize` method will not be called automatically anymore. In case it has to be called, the caller should take care of it, not CLI itself.
Due to this change, the following changes are applied:
* Delete some unit tests
* Expose devicesService initialize method, as it should be called from Proton, Fusion, etc.
* Modify deviceEmitter's initialize method - it's no longer async - remove the async parts as they should be called from other callers. Instead of this, just attach to required events. Call the initialize directly in the constructor of deviceEmitter.
* Fix unit tests and remove mockClassesWithInitializeMethod.ts - they are no longer required.

### Fix issue when callers are unable to find Android devices
In case there's some error in iTunes, CLI (or Proton) will fail when detecting iOS devices and will not continue with Android and iOS Simulators. Fix `detectCurrentlyAttachedDevices` by try-catching each call separately.


justcodebuilduser:
:broken_heart:

justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

PR #891, opened 2017-02-12
Fix working with android devices in some cases

rosen-vladimirov:
Fix working with Android devices:
- incorrect error is raise when adb writes multiple times to stderr.
Instead of handling all errors when the process exists, we've rejected
the promise on the first write to stderr flag. Add unit-test for this.
- An error is raised during construction of AndroidDevice object when
the device is not authorized. In this case we should not throw error -
just set the status to Unreachable. The caller of AndroidDevice must
handle the case when the status is Unreachable.
- Fix case when some adb messages are reported as AndroidDevices, for
example `* daemon started successfully *`. Filter such messages in the
androidDeviceDiscovery. Add unit tests for such known messages.


justcodebuilduser:
:heart:

PR #892, opened 2017-02-16
Fix ENOENT when sync to iOS simulator

rosen-vladimirov:
In some cases we are not able to sync files to iOS Simulator as chokidar does not raise correct events on macOS when directory is renamed immediately after it's being created.
After adding a file to this directory we throw ENOENT, as such dir does not exist in Simulator's dir. In order to fix this, ensure the directory exist in the simulator.


justcodebuilduser:
:heart:

PR #895, opened 2017-02-21
Fix getting info from registry.npmjs.org for specific version

rosen-vladimirov:
CLI tries to get information for specific version of a package by calling registry.npmjs.org.
It looks like recently the registry.npmjs.org has been modified and when passing version as query parameter, it just disregards it. Now the version is part of the URL itself.
Old: `http://registry.npmjs.org/rosen-test-npm-package?version=1.0.0`
New: `http://registry.npmjs.org/rosen-test-npm-package/1.0.0`

So fix CLI's code where we are generating the URL.

### DIFF
--- a/appbuilder/services/npm-service.ts
+++ b/appbuilder/services/npm-service.ts
@@ -184,7 +184,7 @@ export class NpmService implements INpmService {
 			registryUrl += "/";
 		}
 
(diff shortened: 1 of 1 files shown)


justcodebuilduser:
:heart:

PR #896, opened 2017-02-21
Fix getting package.json from npm registry (again)

rosen-vladimirov:
CLI needs information for specific versions of packages. However registry.npmjs.org have broken something and the URL we've been using (`http://registry.npmjs.org/<package_name>?version=<version>`) does not work anymore.
Instead it returns information for all versions.
We've tried calling `http://registry.npmjs.org/<package_name>/<version>`, which seemed to work fine, but it turned out it has a lot of issues with scoped dependencies.
Issues are describe [here](https://github.com/npm/registry/issues/130).
In order to fix CLI and allow installation of all plugins, get the full information from the registry and find only the info we need for the specified version.

### DIFF
--- a/appbuilder/services/npm-service.ts
+++ b/appbuilder/services/npm-service.ts
@@ -54,7 +54,7 @@ export class NpmService implements INpmService {
 				npmInstallResult.error = err;
 			}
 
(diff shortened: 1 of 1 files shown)


justcodebuilduser:
:heart:

PR #899, opened 2017-02-24
Update grunt-shell

TsvetanMilanov:
We started to receive `stdout maxBuffer exceeded Use --force to continue` in our PR builds. This happens because we use old version of grunt-shell.


justcodebuilduser:
:heart:

PR #904, opened 2017-02-28
emulator name was undefined because we missed an await

Plamen5kov:
See more info: https://github.com/NativeScript/nativescript-cli/issues/2576

_Problem_
When `tns run/debug android` is ran, with no running emulator, the emulator was started, the app was installed, but not launched. The problem is after the emulator was started we couldn't get the name of the emulator.

Related issue: https://github.com/NativeScript/nativescript-cli/issues/2486

### DIFF
--- a/mobile/android/android-emulator-services.ts
+++ b/mobile/android/android-emulator-services.ts
@@ -190,7 +190,6 @@ class AndroidEmulatorServices implements Mobile.IAndroidEmulatorServices {
 				return emulatorId;
 			}
 		}
-		this.$errors.failWithoutHelp("Couldn't find emulator id");
 	}
 
 	@invokeInit()
--- a/mobile/android/android-emulator-services.ts
+++ b/mobile/android/android-emulator-services.ts
@@ -184,9 +184,13 @@ class AndroidEmulatorServices implements Mobile.IAndroidEmulatorServices {
 
 		// if we get here, there's at least one running emulator
 		let getNameFunction = this.$options.geny ? this.getNameFromGenymotionEmulatorId : this.getNameFromSDKEmulatorId;
-		let emulatorId = await _(runningEmulators).find(async emulator => await getNameFunction.apply(this, [emulator]) === image);
-
-		return emulatorId;
+		for (let emulatorId of runningEmulators) {
+			const currentEmulatorName = await getNameFunction.apply(this, [emulatorId]);
+			if (currentEmulatorName === image) {
+				return emulatorId;
+			}
+		}
+		this.$errors.failWithoutHelp("Couldn't find emulator id");
 	}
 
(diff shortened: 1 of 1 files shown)


justcodebuilduser:
:heart:

mobile/android/android-emulator-services.ts
```diff
@@ -184,7 +184,7 @@ class AndroidEmulatorServices implements Mobile.IAndroidEmulatorServices {
 
 		// if we get here, there's at least one running emulator
 		let getNameFunction = this.$options.geny ? this.getNameFromGenymotionEmulatorId : this.getNameFromSDKEmulatorId;
-		let emulatorId = await _(runningEmulators).find(emulator => getNameFunction.apply(this, [emulator]) === image);
+		let emulatorId = await _(runningEmulators).find(async emulator => await getNameFunction.apply(this, [emulator]) === image);
```
rosen-vladimirov: this code will not wait for the execution of the lambdas inside `find`
It should be something like:
```TypeScript
for (let emulator of runningEmulators) {
    const currentEmulatorName = await getNameFunction.apply(this, [emulator]);
    if (currentEmulatorName === image) {
        return emulator;
    }
}
```

Plamen5kov: As I looked at the generated javascript, I came to the same conclusion. Will fix.

justcodebuilduser:
:heart:

mobile/android/android-emulator-services.ts
```diff
@@ -184,9 +184,13 @@ class AndroidEmulatorServices implements Mobile.IAndroidEmulatorServices {
...
-		let emulatorId = await _(runningEmulators).find(emulator => getNameFunction.apply(this, [emulator]) === image);
-
-		return emulatorId;
+		for (let emulatorId of runningEmulators) {
+			const currentEmulatorName = await getNameFunction.apply(this, [emulatorId]);
+			if (currentEmulatorName === image) {
+				return emulatorId;
+			}
+		}
+		this.$errors.failWithoutHelp("Couldn't find emulator id");
```
rosen-vladimirov: This changes the behavior of the method. Previously it's been returning undefined. And the call to the method relies on this undefined value:
https://github.com/telerik/mobile-cli-lib/blob/master/mobile/android/android-emulator-services.ts#L269

Is there a particular reason to add this failure here?

Plamen5kov: Forgot it there, while testing.

justcodebuilduser:
:heart:

Plamen5kov:
related to: https://github.com/NativeScript/nativescript-cli/issues/2210

>It happen always when emulator is not started before tns run android
this is fixed.

PR #905, opened 2017-03-14
Fix device log filter

TsvetanMilanov:
We have added one parameter in the filter method which we do not need and we do not respect it when we call the method. The solution is to remove it.

### DIFF
--- a/definitions/mobile.d.ts
+++ b/definitions/mobile.d.ts
@@ -205,12 +205,11 @@ declare module Mobile {
 		 * Filters data for specified platform.
 		 * @param {string} platform The platform for which is the device log.
 		 * @param {string} data The input data for filtering.
-		 * @param {string} projectDir The project's root directory.
 		 * @param {string} pid @optional The application PID for this device.
 		 * @param {string} logLevel @optional The logging level based on which input data will be filtered.
 		 * @return {string} The filtered result based on the input or null when the input data shouldn't be shown.
 		 */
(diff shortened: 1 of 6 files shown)


justcodebuilduser:
:broken_heart:

justcodebuilduser:
:heart:

definitions/mobile.d.ts
```diff
@@ -205,12 +205,11 @@ declare module Mobile {
 		 * Filters data for specified platform.
 		 * @param {string} platform The platform for which is the device log.
 		 * @param {string} data The input data for filtering.
-		 * @param {string} projectDir The project's root directory.
 		 * @param {string} pid @optional The application PID for this device.
 		 * @param {string} logLevel @optional The logging level based on which input data will be filtered.
 		 * @return {string} The filtered result based on the input or null when the input data shouldn't be shown.
 		 */
-		filterData(platform: string, data: string, projectDir: string, pid?: string, logLevel?: string): string;
```
rosen-vladimirov: This is required for NativeScript CLI:
https://github.com/NativeScript/nativescript-cli/blob/master/lib/services/ios-log-filter.ts#L11

TsvetanMilanov: The `projectDirectory` is required only in the `ios-log-filter` in the NS CLI. That's why we will remove the parameter from the method and add fix in the NS CLI.

PR #907, opened 2017-03-14
Fix getting info for Android tools

rosen-vladimirov:
Due to changes in Android SDK, we have to update the checks in CLI. While gathering system information, we check the android executable, which is no longer returning correct results.
In order to fix this, we rely on the emulator executable, which is the real thing we need as it is the one that allows us to work with Android Emulators.
Fix sys-info checks and get correct path to emulator according to latest changes.

### DIFF
--- a/sys-info-base.ts
+++ b/sys-info-base.ts
@@ -1,14 +1,15 @@
 import * as os from "os";
 import * as osenv from "osenv";
 import * as path from "path";
-import {quoteString} from "./helpers";
+import { quoteString } from "./helpers";
 
 export class SysInfoBase implements ISysInfo {
 	constructor(protected $childProcess: IChildProcess,
-				protected $hostInfo: IHostInfo,
-				protected $iTunesValidator: Mobile.IiTunesValidator,
-				protected $logger: ILogger,
-				protected $winreg: IWinReg) { }
+		protected $hostInfo: IHostInfo,
+		protected $iTunesValidator: Mobile.IiTunesValidator,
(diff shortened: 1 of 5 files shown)


justcodebuilduser:
:heart:

sys-info-base.ts
```diff
@@ -208,38 +208,37 @@ export class SysInfoBase implements ISysInfo {
...
-				}
-			} catch(err) {
-				this.$logger.trace(`Error while checking is ${pathToAndroid} installed. Error is: ${err.messge}`);
+				// emulator -help exits with code 1 on Windows, so we should parse the output.
+				// First line of it should be:
+				// Android Emulator usage: emulator [options] [-qemu args]
+				const emulatorHelp = this.$childProcess.spawnFromEvent(this.$androidEmulatorServices.pathToEmulatorExecutable, ["-help"], "close", {}, { throwError: false }).wait();
+				result = emulatorHelp.stdout.indexOf("usage: emulator") !== -1;
+			} catch (err) {
+				this.$logger.trace(`Error while checking is emulator installed. Error is: ${err.messge}`);
```
TsvetanMilanov: I think we will never trace the error message from `emulator -help`, if there is one, because we set `throwError: false` on spawnFromEvent. You can check if `result` is `false` and trace the message from `emulatorHelp.stdout` or `emulatorHelp.stderr`.

justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

PR #908, opened 2017-03-16
Fixes for Cloud Builds

rosen-vladimirov:
### Pass excludedProjectDirsAndFiles from getProjectFiles
The parameter is not passed to the isFileExcluded, so you can never ignore additional dirs.
Update semver definitions and package to latest one.

### Fix hooks execution when using CLI as lib
When NS CLI is used as library, project hooks are not executed as hooksService gets the project dir from `$options` or current working dir.
Fix this by getting the projectDir from the hookArgs object. The object is constructed from method parameters. We check if there's `projectDir` or `projectData.projectDir` arg.
Also fix the caching of hooks - each time when we try to execute them, we'll search the directories for hooks. This is required for long living process.


justcodebuilduser:
:heart:

PR #909, opened 2017-03-20
Do not check for application updates for Unreachable devices

rosen-vladimirov:
When devices are not with status Connected, we should not check for application updates on these devices as the code will fail.
Fix this in the deviceDetectionInterval logic, so long living processes will work correctly.


justcodebuilduser:
:heart:

PR #910, opened 2017-03-20
Fix copy of MobileApp.app build file after creation

yyosifov:
When the --copy-to flag in the CLI is set, we use the CopyFile method from FileSystem to copy the built file - AppName.app to the destination. Mac OS resolves this file as a directory in the terminal and the cp command fails with "omitting directory" because it's missing the -R flag.


justcodebuilduser:
:heart:

PR #911, opened 2017-03-20
Enable authenticated proxy

Mitko-Kerezov:

The intended workflow is as follows:

```BASH
$ appbuilder provision
Your proxy requires a username and password. You can run
        appbuilder proxy set <hostname> <port> <username> <password>
In order to supply appbuilder with the credentials needed.
$ appbuilder proxy set 127.0.0.1 8888 user pass
Sucessfully setup proxy.
$ appbuilder provision
1. Provision 1
2. Provision 2
```

Apart from `setting` proxy data a user can `get` and `clear` it


justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

http-client.ts
```diff
@@ -176,7 +183,7 @@ export class HttpClient implements Server.IHttpClient {
 		return response;
 	}
 
-	private async setResponseResult(result: IPromiseActions<Server.IResponse>, timerId: number, resultData: { response?: Server.IRequestResponseData, body?: string, err?: Error,  }): Promise<void> {
+	private async setResponseResult(result: IPromiseActions<Server.IResponse>, timerId: number, resultData: { response?: Server.IRequestResponseData, body?: string, err?: Error, }): Promise<void> {
```
TsvetanMilanov: I think we don't need the `,` after `err?: Error` in the `resultData` type definition.

services/credentials-service.ts
```diff
@@ -0,0 +1,62 @@
...
+		private $hostInfo: IHostInfo) {
+		this.pathToWindowsHelperScript = path.join(__dirname, "..", "vendor", "WinCredentialsHelper.ps1");
+	}
+
+	public async setCredentials(key: string, credentials: ICredentials): Promise<ICredentials> {
+		if (this.$hostInfo.isWindows) {
+			await this.$childProcess.spawnFromEvent("powershell.exe", [this.pathToWindowsHelperScript, "-AddCred", "-Target", key, "-User", credentials.username, "-Pass", this.encrypt(credentials.password)], "close");
+			return credentials;
+		} else {
+			throw new Error("Not implemented");
```
TsvetanMilanov: I think the message here should be something like `Storing credentials is not supported on {osName} yet.`

services/proxy-cache-service.ts
```diff
@@ -0,0 +1,39 @@
...
+		return this.$fs.exists(this.proxyCacheFilePath) && this.$fs.readJson(this.proxyCacheFilePath);
+	}
+
+	public clearCache(): void {
+		this.$fs.deleteFile(this.proxyCacheFilePath);
+		this.$credentialsService.clearCredentials(this.credentialsKey);
+	}
+
+	public async getCredentials(): Promise<ICredentials> {
+		return this.$credentialsService.getCredentials(this.credentialsKey);
```
TsvetanMilanov: I think `credentialsKey` can be property in the `staticConfig` and remove the `get/setCredentials` from the `ProxyCacheService` and use the `CredentialsService`.

commands/proxy.ts
```diff
@@ -0,0 +1,102 @@
...
+
+	constructor(private $injector: IInjector,
+		private $logger: ILogger,
+		private $prompter: IPrompter,
+		private $proxyCacheService: IProxyCacheService) {
+	}
+
+	public async execute(args: string[]): Promise<void> {
+		let hostname = args[0],
+			port = args[1],
```
Mitko-Kerezov: `--` flags have a semantic meaning of altering a command's behavior which is not the case here. I'm reluctant to use `--` flags in this particular case and I'd rather stick with command-line parameters

services/proxy-cache-service.ts
Mitko-Kerezov: I think it is logically more appropriate that way, because if we have say `this.$staticConfig.PROXY_CREDENTIALS_KEY` that means that whoever want to get/set proxy credential details would have to know about `$staticConfig` as well as `$credentialsService`

Basically what I'm trying to say is - I think it's better to abstract all proxy-related things in one service instead of leaking logic to the code that intends to work with any proxy-related stuff.

I hope that makes sense

Mitko-Kerezov:
Ping @TsvetanMilanov with fixed comments

justcodebuilduser:
:heart:

services/proxy-cache-service.ts
TsvetanMilanov: You are right. Do you think it will be OK to rename the `ProxyCacheService` to `ProxyService`? I suggested to remove the `get/setCredentials` from the `ProxyCacheService` because I thought this service will only handle the caching part of the proxy operations.

Mitko-Kerezov: Seems reasonable

justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

commands/proxy.ts
```diff
@@ -0,0 +1,103 @@
...
+		}
+
+		const proxyCache: IProxyCache = {
+			PROXY_HOSTNAME: hostname,
+			PROXY_PORT: +port,
+			USE_PROXY: true
+		};
+
+		this.$proxyService.setCache(proxyCache);
+		this.$logger.out("Sucessfully setup proxy.");
```
rosen-vladimirov: The command requires disabling of analytics, however I would like to know how many of the customers are using it. Can you add some feature tracking here. You can follow the syntax for commands or introduce new `feature` in the analytics.

commands/proxy.ts
```diff
@@ -0,0 +1,103 @@
...
+		let message = "";
+		const proxyCache: IProxyCache = this.$proxyService.getCache();
+		if (proxyCache) {
+			const proxyCredentials = await this.$proxyService.getCredentials();
+			message = `Hostname: ${proxyCache.PROXY_HOSTNAME}${EOL}Port: ${proxyCache.PROXY_PORT}`;
+			if (proxyCredentials && proxyCredentials.username) {
+				message += `${EOL}Username: ${proxyCredentials.username}`;
+			}
+
+			message += `${EOL}PROXY IS ${proxyCache.USE_PROXY ? "ENABLED" : "DISABLED"}`;
```
rosen-vladimirov: Do we need capital letters here?

commands/proxy.ts
```diff
@@ -0,0 +1,103 @@
...
+		} else {
+			message = "No proxy set";
+		}
+
+		this.$logger.out(message);
```
rosen-vladimirov: Maybe we can use `printMarkdown` for better readability of the proxy settings, but I do not insist for this.

commands/proxy.ts
```diff
@@ -0,0 +1,103 @@
...
+	public disableAnalytics = true;
+
+	public allowedParameters: ICommandParameter[] = [];
+
+	constructor(private $logger: ILogger,
+		private $proxyService: IProxyService) {
+	}
+
+	public async execute(args: string[]): Promise<void> {
+		// TODO: Clear credentials from vault
```
rosen-vladimirov: still TODO?

http-client.ts
```diff
@@ -40,17 +41,23 @@ export class HttpClient implements Server.IHttpClient {
 		let pipeTo = options.pipeTo;
 		delete options.pipeTo;
 
-		let proto = this.$config.USE_PROXY ? "http" : requestProto;
+		const proxyCache = this.$proxyService.getCache();
+		let proto = proxyCache && proxyCache.USE_PROXY ? "http" : requestProto;
```
rosen-vladimirov: shouldn't you respect the `proxySettings` here?

http-client.ts
```diff
@@ -229,7 +236,10 @@ export class HttpClient implements Server.IHttpClient {
 	}
 
 	private getErrorMessage(response: Server.IRequestResponseData, body: string): string {
-		if (response.statusCode === 402) {
+		if (response.statusCode === 407) {
+			const clientNameLowerCase = this.$staticConfig.CLIENT_NAME.toLowerCase();
+			return `Your proxy requires a username and password. You can run \n\t${clientNameLowerCase} proxy set <hostname> <port> <username> <password>.\nIn order to supply ${clientNameLowerCase} with the credentials needed.`;
```
rosen-vladimirov: `os.EOL` instead of `\n` maybe?

services/credentials-service.ts
```diff
@@ -0,0 +1,63 @@
...
+			await this.$childProcess.spawnFromEvent("powershell.exe", [this.pathToWindowsHelperScript, "-AddCred", "-Target", key, "-User", credentials.username, "-Pass", this.encrypt(credentials.password)], "close");
+			return credentials;
+		} else {
+			throw new Error(`Storing credentials is not supported on ${platform()} yet.`);
+		}
+	}
+
+	public async getCredentials(key: string): Promise<ICredentials> {
+		if (this.$hostInfo.isWindows) {
+			let credentialsSpawnResult = await this.$childProcess.spawnFromEvent("powershell.exe", [this.pathToWindowsHelperScript, "-GetCred", "-Target", key], "close");
```
rosen-vladimirov: const for all `let`s

commands/proxy.ts
Mitko-Kerezov: Implemented

commands/proxy.ts
Mitko-Kerezov: I think I prefer it without markdown - it displays three simple properties no need to be fancy

Mitko-Kerezov:
Ping @rosen-vladimirov with fixed comments

justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

commands/proxy.ts
```diff
@@ -0,0 +1,145 @@
...
+
+	public abstract allowedParameters: ICommandParameter[];
+
+	public abstract execute(args: string[]): Promise<void>;
+
+	protected async tryTrackUsage() {
+		try {
+			await this.$analyticsService.trackFeature(this.commandName);
+		} catch (ex) {
+			// Swallow exception - there might be something wrong with the proxy or its credentials
```
rosen-vladimirov: I do not like swallowing anything... let's trace it.

commands/proxy.ts
```diff
@@ -0,0 +1,145 @@
...
+		}
+	}
+}
+
+export class ProxySetCommand extends ProxyCommandBase {
+	public disableAnalytics = true;
```
rosen-vladimirov: you do not need this, base command sets it

commands/proxy.ts
```diff
@@ -0,0 +1,145 @@
...
+		this.$proxyService.setCache(proxyCache);
+		this.$logger.out("Sucessfully setup proxy.");
+		await this.tryTrackUsage();
+	}
+}
+
+$injector.registerCommand(proxySetCommandName, ProxySetCommand);
+
+export class ProxyGetCommand extends ProxyCommandBase {
+	public allowedParameters: ICommandParameter[] = [];
```
rosen-vladimirov: you can set it to [] in base command and omit it here.

commands/proxy.ts
```diff
@@ -0,0 +1,145 @@
...
+
+		this.$logger.out(message);
+		await this.tryTrackUsage();
+	}
+}
+
+$injector.registerCommand(proxyGetCommandName, ProxyGetCommand);
+
+export class ProxyClearCommand extends ProxyCommandBase {
+	public disableAnalytics = true;
```
rosen-vladimirov: you do not need this, base command sets it

commands/proxy.ts
```diff
@@ -0,0 +1,145 @@
...
+
+	public allowedParameters: ICommandParameter[] = [];
```
rosen-vladimirov: you can set it to [] in base command and omit it here.

justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

PR #912, opened 2017-03-21
Throw correct error when unable to install app on iOS

rosen-vladimirov:
When we are unable to install app on iOS device, we show warning and we continue the execution. However we have to throw error in such cases.


justcodebuilduser:
:heart:

PR #913, opened 2017-03-21
Correct pipe name when emitting in spawnFromEvent

Mitko-Kerezov:


justcodebuilduser:
:heart:

PR #914, opened 2017-03-22
Fix device detection interval deadlock

rosen-vladimirov:
The exportedDecorator promise may have postAction. In case the postAction returns a promise, we'll not wait for its execution and we'll return the result to the caller.
This leads to problem when there are two consecutive calls to deployOnDevices - after the first one finishes, we start the deviceDetectioninterval, but we do not wait for it to finish the first execution.
When the second deploy starts it sees there's a running interval, clears it and waits for the promise to be resolved. However we've never started the execution, so the promise is never started, so it is never resolved.
Fix this by:
 - clearing the interval after waiting for current promise to be resolved
 - fix the exportedPromise decorator to wait for the first execution of the startDeviceDetection interval to finish

### DIFF
--- a/decorators.ts
+++ b/decorators.ts
@@ -72,7 +72,7 @@ export function invokeInit(): any {
 	return invokeBefore("init");
 }
 
-export function exportedPromise(moduleName: string, postAction?: () => void): any {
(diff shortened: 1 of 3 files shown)


justcodebuilduser:
:heart:

PR #915, opened 2017-03-22
provide a better message

Plamen5kov:
_Problem_
Provide a better message when the device is connected but not paired:

_Solution_
Message is more verbose. Add suggestion for the user.

related issue: https://github.com/NativeScript/nativescript-cli/issues/1882#issuecomment-288433597


justcodebuilduser:
:heart:

PR #916, opened 2017-03-23
Fix stop application on iOS simulator

TsvetanMilanov:
We need to pass the project name instead of the app identifier to kill the application on the simulator.


justcodebuilduser:
:heart:

PR #917, opened 2017-03-23
Stop stopping device detection interval

Mitko-Kerezov:
Delete `@exportedPromise` decorator as we no longer need it
Stop stopping device detection interval when deploying on iOS devices. These operations no longer conflict as the code is located in an external dependency.

### DIFF
--- a/appbuilder/services/livesync/livesync-service.ts
+++ b/appbuilder/services/livesync/livesync-service.ts
@@ -1,4 +1,4 @@
-import { exportedPromise } from "../../../decorators";
+import { exported } from "../../../decorators";
 
 export class ProtonLiveSyncService implements IProtonLiveSyncService {
 	private excludedProjectDirsAndFiles = ["app_resources", "plugins", ".*.tmp", ".ab"];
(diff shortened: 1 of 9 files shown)


justcodebuilduser:
:broken_heart:

justcodebuilduser:
:heart:

appbuilder/services/livesync/livesync-service.ts
```diff
@@ -14,16 +14,13 @@ export class ProtonLiveSyncService implements IProtonLiveSyncService {
 		private $logger: ILogger,
 		private $companionAppsService: ICompanionAppsService) { }
 
-	@exportedPromise("liveSyncService", async function () {
-		await this.$devicesService.startDeviceDetectionInterval();
-	})
 	public livesync(deviceDescriptors: IDeviceLiveSyncInfo[], projectDir: string, filePaths?: string[]): Promise<IDeviceLiveSyncResult>[] {
```
rosen-vladimirov: why this method is not exposed?

test/unit-tests/decorators.ts
```diff
@@ -23,40 +18,56 @@ describe("decorators", () => {
 		$injector = new Yok();
 	});
 
-	describe("exportedPromise", () => {
+	describe.only("exported", () => {
```
rosen-vladimirov: only 😭

rosen-vladimirov:
Also I believe we need the `stopDeviceDetectionInterval` method. In some cases CLI does not need it, but it's started as `skipInferPlatform` is passed to devicesService.initialize.
The other option is to add new option to devicesService's initialize method (doNotStartDeviceDetection) and do not start the interval in case it is passed.

justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

PR #920, opened 2017-03-27
Enable authenticated proxy

rosen-vladimirov:
Cherry-pick https://github.com/telerik/mobile-cli-lib/pull/911 in release-appbuilder


commands/proxy.ts
```diff
@@ -0,0 +1,137 @@
+import * as commandParams from "../command-params";
+import { isInteractive } from "../helpers";
+import { EOL } from "os";
+
+const proxyGetCommandName = "proxy|*get";
+const proxySetCommandName = "proxy|set";
+const proxyClearCommandName = "proxy|clear";
+
+export abstract class ProxyCommandBase implements ICommand {
+	public disableAnalytics = true;
```
yyosifov: Why do we want to disable the analytics here? Is this an internal command only?

justcodebuilduser:
:heart:

commands/proxy.ts
```diff
@@ -0,0 +1,137 @@
...
+			await this.$analyticsService.trackFeature(this.commandName);
+		} catch (ex) {
+			this.$logger.trace("Error in trying to track proxy command usage:");
+			this.$logger.trace(ex);
+		}
+	}
+}
+
+export class ProxySetCommand extends ProxyCommandBase {
+	public allowedParameters = [new commandParams.StringCommandParameter(this.$injector), new commandParams.StringCommandParameter(this.$injector), new commandParams.StringCommandParameter(this.$injector), new commandParams.StringCommandParameter(this.$injector)];
```
yyosifov: quite a long line I think - 265 chars

commands/proxy.ts
```diff
@@ -0,0 +1,137 @@
...
+			port = args[1],
+			username = args[2],
+			password = args[3];
+
+		const noHostName = !hostname;
+		const noPort = !port;
+		const credentialsRequired = username && !password;
+
+		if (!isInteractive() && (noHostName || noPort || credentialsRequired)) {
+			this.$errors.fail("Console is not interactive - you need to supply all command parameters.");
```
yyosifov: Do you think it would be nicer to the user if we pin-point to the missing pieces? What's our common practice? Or shall we at least display the syntax of the command `tns proxy hostname port username password`

commands/proxy.ts
rosen-vladimirov: When this option is not set, the commands service tries to track the command usage before executing the command itself. This will fail when the proxy is still not set (for example when callling `<cli name> proxy set`. That's why we disable the tracking here and track the command at the end of the execution method.

commands/proxy.ts
rosen-vladimirov: `$erros.fail` will display the command help, that's why the message is generic one - the output here will be:
```
<the generic message for missing parameter>
usage of the command
```

commands/proxy.ts
```diff
@@ -0,0 +1,137 @@
...
+
+		if (noPort) {
+			port = await this.$prompter.getString("Port", { allowEmpty: false });
+		}
+
+		if (!username) {
+			username = await this.$prompter.getString("Username (Leave empty to omit.)", { defaultAction: () => "" });
+		}
+
+		if (credentialsRequired) {
```
yyosifov: At this point, this variable's value can be incorrect. If the `username` hasn't been passed, and the user passes it this would be false.

commands/proxy.ts
```diff
@@ -0,0 +1,137 @@
...
+		if (username && password) {
+			await this.$proxyService.setCredentials({
+				password,
+				username
+			});
+		}
+
+		const proxyCache: IProxyCache = {
+			PROXY_HOSTNAME: hostname,
+			PROXY_PORT: +port,
```
yyosifov: do we need to validate the Port? What if it's -13/0

commands/proxy.ts
```diff
@@ -0,0 +1,137 @@
...
+			USE_PROXY: true
+		};
+
+		this.$proxyService.setCache(proxyCache);
+		this.$logger.out("Sucessfully setup proxy.");
```
yyosifov: small typo: successfully

commands/proxy.ts
```diff
@@ -0,0 +1,137 @@
...
+		await this.tryTrackUsage();
+	}
+}
+
+$injector.registerCommand(proxySetCommandName, ProxySetCommand);
+
+export class ProxyGetCommand extends ProxyCommandBase {
```
yyosifov: I guess this is the common practice, but when I navigate to classes if there are more than one in a file, it doesn't work correctly. What is more as a whole I've been used to the 1 file = 1 class rule.

commands/proxy.ts
```diff
@@ -0,0 +1,137 @@
...
+export class ProxyClearCommand extends ProxyCommandBase {
+	constructor(protected $analyticsService: IAnalyticsService,
+		protected $logger: ILogger,
+		protected $proxyService: IProxyService) {
+		super($analyticsService, $logger, $proxyService, proxyClearCommandName);
+	}
+
+	public async execute(args: string[]): Promise<void> {
+		this.$proxyService.clearCache();
+		this.$logger.out("Sucessfully cleared proxy.");
```
yyosifov: same typo

http-client.ts
```diff
@@ -40,17 +42,23 @@ export class HttpClient implements Server.IHttpClient {
...
+		if (proxySettings || (proxyCache && proxyCache.USE_PROXY)) {
+			const proxyCredentials = await this.$proxyService.getCredentials();
 			options.path = requestProto + "://" + options.host + options.path;
 			headers.Host = options.host;
-			options.host = (proxySettings && proxySettings.hostname) || this.$config.PROXY_HOSTNAME;
-			options.port = (proxySettings && proxySettings.port) || this.$config.PROXY_PORT;
+			options.host = (proxySettings && proxySettings.hostname) || proxyCache.PROXY_HOSTNAME;
+			options.port = (proxySettings && proxySettings.port) || proxyCache.PROXY_PORT;
+			if (proxyCredentials && proxyCredentials.username && proxyCredentials.password) {
+				headers["Proxy-Authorization"] = "Basic " + new Buffer(`${proxyCredentials.username}:${proxyCredentials.password}`).toString('base64');
```
yyosifov: Do we want to support cases for non-Basic authorization? Or we can note it in the documentation that we support only Basic Auth.

http-client.ts
```diff
@@ -40,17 +42,23 @@ export class HttpClient implements Server.IHttpClient {
...
-		if (proxySettings || this.$config.USE_PROXY) {
+		if (proxySettings || (proxyCache && proxyCache.USE_PROXY)) {
+			const proxyCredentials = await this.$proxyService.getCredentials();
 			options.path = requestProto + "://" + options.host + options.path;
 			headers.Host = options.host;
-			options.host = (proxySettings && proxySettings.hostname) || this.$config.PROXY_HOSTNAME;
-			options.port = (proxySettings && proxySettings.port) || this.$config.PROXY_PORT;
+			options.host = (proxySettings && proxySettings.hostname) || proxyCache.PROXY_HOSTNAME;
+			options.port = (proxySettings && proxySettings.port) || proxyCache.PROXY_PORT;
+			if (proxyCredentials && proxyCredentials.username && proxyCredentials.password) {
```
yyosifov: I would move the `const proxyCredentials` just above the if, as it's the first & only place it's used and it will be easier to read for me. There are too many proxy* variables anyway.

http-client.ts
```diff
@@ -40,17 +42,23 @@ export class HttpClient implements Server.IHttpClient {
...
-		let proto = this.$config.USE_PROXY ? "http" : requestProto;
+		const proxyCache = this.$proxyService.getCache();
+		let proto = proxyCache && proxyCache.USE_PROXY ? "http" : requestProto;
 		let http = require(proto);
 
 		options.headers = options.headers || {};
 		let headers = options.headers;
 
-		if (proxySettings || this.$config.USE_PROXY) {
+		if (proxySettings || (proxyCache && proxyCache.USE_PROXY)) {
```
yyosifov: We can extract this to it's own method? So that we can make unit tests for it and validate that we don't have some regressions?

http-client.ts
```diff
@@ -229,7 +237,10 @@ export class HttpClient implements Server.IHttpClient {
 	}
 
 	private getErrorMessage(response: Server.IRequestResponseData, body: string): string {
-		if (response.statusCode === 402) {
+		if (response.statusCode === 407) {
```
yyosifov: I'd suggest we add some constants with the frequently-used Http Status code for easier readability

services/credentials-service.ts
```diff
@@ -0,0 +1,60 @@
...
+		private $hostInfo: IHostInfo) {
+		this.pathToWindowsCredentialsManager = path.join(__dirname, "..", "vendor", platform(), "CredentialsManager.exe");
+	}
+
+	public async setCredentials(key: string, credentials: ICredentials): Promise<ICredentials> {
+		if (this.$hostInfo.isWindows) {
+			await this.$childProcess.spawnFromEvent(this.pathToWindowsCredentialsManager, ["set", key, credentials.username, this.encrypt(credentials.password)], "close");
+			return credentials;
+		} else {
+			throw new Error(`Storing credentials is not supported on ${platform()} yet.`);
```
yyosifov: So, this will work only for Windows?

rosen-vladimirov: At the moment - yes. You'll be able to set proxy on other OSes (as it is at the moment), but you'll not be able to use authenticated proxy.

http-client.ts
```diff
@@ -8,9 +9,10 @@ import filesize = require("filesize");
 export class HttpClient implements Server.IHttpClient {
```
Plamen5kov: httpclient has new behavior as it now throws `Storing credentials is not supported on <platform>` exception. Consider introducing tests for this behavior.

yok.ts
```diff
@@ -192,6 +192,7 @@ export class Yok implements IInjector {
 	private createHierarchicalCommand(name: string) {
 		let factory = () => {
 			return {
+				disableAnalytics: true,
```
yyosifov: why is this change needed and connected to the proxies?

services/credentials-service.ts
yyosifov: I was wondering, why do we have to support both ways to set "hostname", "port" if we don't think of the Authenticated scenario which should be an addition to the standard "proxy" scenario? As I understood I can now have two different proxies - one in config, one in cache. In this case it will use the one from the config + {username,pass} from the cache which is confusing? Have you considered to store only the username/pass in the Credentials Storage and leave the hostname/port as is and manipulate them in the config.json via the command?

commands/proxy.ts
rosen-vladimirov: interestingly why the linter did not caught this

http-client.ts
rosen-vladimirov: At the moment only Basic Authentication will be supported.

services/credentials-service.ts
rosen-vladimirov: I'm not sure what you mean here. In all cases the hostname, port and USE_PROXY values are persisted in a file inside a special folder (known as profileDir in CLI). The options from config.json are deleted and they are no longer supported. Only the credentials are stored in a special location.

yok.ts
rosen-vladimirov: The `createHierarchicalCommand` is a little magic for commands defined as `<part1>|<part2>` - during parsing of the command line, CLI tries to check each argument for a valid command and if there's no such command, it will fail.
So in case you write `<cli name> part1 part2` CLI gets the part1 and tries to resolve a command called part1. You have not defined such in the code. This is resolved with the current code. It's been mistakenly forgotten that such commands should not be tracked.

As per the relation with proxy commands - again - when you execute `<cli name> proxy set ...`, CLI will first resolve this command (during evaluating the proxy part of the command line. As it does not have disabledAnalytics, CLI will try to track that this command is used and the request will fail (as the proxy is still not set).

http-client.ts
rosen-vladimirov: done

justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

PR #921, opened 2017-03-27
Fix ios-device-operations dispose

TsvetanMilanov:
We need to check if the justlaunch flag is passed before we set the should dispose property to false. If we set it to false unconditionally when we pass justlaunch flag the process will never exit.


justcodebuilduser:
:heart:

PR #922, opened 2017-03-27
skip the hidden files when livesyncing

yyosifov:
Add filter for files which are hidden in linux/Mac to be skipped when livesync-ing the app. Some IDEs create them to backup the currently edited file and this causes a re-deploy/re-run of the Application although it's not necessary.

### DIFF
--- a/services/livesync-service-base.ts
+++ b/services/livesync-service-base.ts
@@ -45,7 +45,17 @@ class LiveSyncServiceBase implements ILiveSyncServiceBase {
 			}
 		});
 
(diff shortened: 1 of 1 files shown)


justcodebuilduser:
:heart:

yyosifov:
ping @rosen-vladimirov do you think it's safe to merge? It's 2 weeks old already, I'd prefer if we either merge it or close it

rosen-vladimirov:
Closing this as in some cases in AppBuilder we want to sync hidden files (the project file is called `.abproject` and in some cases we want to upload it to device.

PR #925, opened 2017-03-28
Add filter by platform when searching device

yyosifov:
When in debug mode we should respect the passed platform and search only for devices for the respective platform. If we have devices from both android and iOS we recognized them both as valid devices and started an emulator.

### DIFF
--- a/test/unit-tests/mobile/devices-service.ts
+++ b/test/unit-tests/mobile/devices-service.ts
@@ -236,14 +236,7 @@ describe("devicesService", () => {
 		testInjector: IInjector,
 		devicesService: Mobile.IDevicesService,
 		androidEmulatorServices: any,
-		logger: CommonLoggerStub,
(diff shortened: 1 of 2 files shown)


justcodebuilduser:
:broken_heart:

test/unit-tests/mobile/devices-service.ts
```diff
@@ -607,7 +600,6 @@ describe("devicesService", () => {
 				assert.deepEqual(counter, 1, "The action must be executed on only one device.");
 				counter = 0;
 				androidDeviceDiscovery.emit("deviceLost", tempDevice);
-				await assertAndroidEmulatorIsStarted({ platform: "android" });
```
Plamen5kov: this is the second time the `devicesService.initialize()` is called, and it's expected to not start emulator

justcodebuilduser:
:heart:

PR #926, opened 2017-03-28
added alias for device command

Plamen5kov:
made alias for `tns device` command. Now it also works with `devices`.


justcodebuilduser:
:heart:

PR #927, opened 2017-03-30
Add debug method to the public API

TsvetanMilanov:
We need to add debug service to the public API of the {N} CLI. This service should have method which starts debugging and returns the chrome devtools url. With this service it will be possible to use the CLI as a library and start debugging.
This PR includes:
- Return the results from the action in devicesService.execute(action)

### DIFF
--- a/helpers.ts
+++ b/helpers.ts
@@ -388,49 +388,6 @@ export async function connectEventuallyUntilTimeout(factory: () => net.Socket, t
 	});
 }
 
-export function isPortAvailable(candidatePort: number): Promise<boolean> {
-	return new Promise<boolean>((resolve, reject) => {
-		let isResolved = false;
-		let server = net.createServer();
-
-		server.on("error", (err: Error) => {
-			if (!isResolved) {
-				isResolved = true;
-				resolve(false);
-			}
-		});
-
-		server.once("close", () => {
-			if (!isResolved) { // "close" will be emitted right after "error"
-				isResolved = true;
-				resolve(true);
-			}
-		});
-
-		server.on("listening", (err: Error) => {
-			if (err && !isResolved) {
-				isResolved = true;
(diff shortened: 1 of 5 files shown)


justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

helpers.ts
```diff
@@ -388,6 +388,49 @@ export async function connectEventuallyUntilTimeout(factory: () => net.Socket, t
...
+			}
+
+			server.close();
+		});
+
+		server.listen(candidatePort, "localhost");
+	});
+}
+
+export async function getAvailablePort(port: number): Promise<number> {
```
rosen-vladimirov: the correct way to get available port is by starting server on port 0 and waiting for it's listening event, where you'll receive the real free port. So I suggest either renaming the method or use 
https://github.com/telerik/mobile-cli-lib/blob/master/services/net-service.ts#L4

helpers.ts
```diff
@@ -388,6 +388,49 @@ export async function connectEventuallyUntilTimeout(factory: () => net.Socket, t
 	});
 }
 
+export function isPortAvailable(candidatePort: number): Promise<boolean> {
```
rosen-vladimirov: this method should be moved to `netService`

helpers.ts
TsvetanMilanov: We want to have a method which can get available port in some range. In the debug service in {N} CLI we want to check for free port starting from a specific port because we want to start chrome devtools on one port every time if it is possible. This way the users will be able to just refresh the chrome devtools page when using the debug command again.

justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

PR #929, opened 2017-03-31
Fix proxy support

Mitko-Kerezov:
* Use `request` npm module which has out-of-the-box support for proxies
* Allow the user to specify proxy protocol - up until now we only had `http`
* Rename `proxy-<action>-command.ts` files to `proxy-<action>.ts`


justcodebuilduser:
:broken_heart:

justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

commands/proxy/proxy-set.ts
```diff
@@ -22,22 +23,30 @@ export class ProxySetCommand extends ProxyCommandBase {
...
+			if (!isInteractive()) {
+				this.$errors.fail("Console is not interactive - you need to supply all command parameters.");
+			} else {
+				urlString = await this.$prompter.getString("Url", { allowEmpty: false });
+			}
+		}
 
-		const noHostName = !hostname;
+		const urlObj = parse(urlString);
+		let port = urlObj.port && +urlObj.port || HttpProtocolToPort[urlObj.protocol];
```
rosen-vladimirov: is there any way for port to be undefined after this line? In case not, please fix the next line

commands/proxy/proxy-set.ts
```diff
@@ -22,22 +23,30 @@ export class ProxySetCommand extends ProxyCommandBase {
...
 		const noPort = !port || !this.isValidPort(port);
+		const authCredentials = this.getCredentialsFromAuth(urlObj.auth || "");
+		username = username || authCredentials.username;
```
rosen-vladimirov: I believe we should fail in case you specify username in both ways, for example:
`appbuilder proxy set http://username:password@127.0.0.1:8888 username1 password1`
should fail. In case the username and password on both places are the same, we should not fail.

http-client.ts
```diff
@@ -97,65 +95,77 @@ export class HttpClient implements Server.IHttpClient {
...
-			});
-
-			request.on("error", (err: Error) => {
-				this.setResponseResult(promiseActions, timerId, { err });
-			});
+			options.url = options.url || `${options.proto}://${options.host}${options.path}`;
+			options.encoding = null;
+			options.followAllRedirects = true;
+
+			let requestObj = request(options);
```
rosen-vladimirov: `const` maybe

http-client.ts
```diff
@@ -97,65 +95,77 @@ export class HttpClient implements Server.IHttpClient {
...
+
+			requestObj
+				.on("error", (err: Error) => {
+					this.$logger.trace("An error occurred while sending the request:", err);
+					const errorMessageMatch = err.message.match(HttpClient.STATUS_CODE_REGEX);
+					const errorMessageStatusCode = errorMessageMatch && errorMessageMatch[1] && +errorMessageMatch[1];
+					let errorMessage = this.getErrorMessage(errorMessageStatusCode, null);
```
rosen-vladimirov: const

http-client.ts
```diff
@@ -97,65 +95,77 @@ export class HttpClient implements Server.IHttpClient {
 
 			this.$logger.trace("httpRequest: %s", util.inspect(options));
```
rosen-vladimirov: this trace must be immediately before the request - currently it will not print all options that we pass and we may be confused when asking customers for logs.

http-client.ts
```diff
@@ -266,16 +277,19 @@ export class HttpClient implements Server.IHttpClient {
...
+			let credentialsPart = "";
 			const proxyCredentials = await this.$proxyService.getCredentials();
 			if (proxyCredentials && proxyCredentials.username && proxyCredentials.password) {
-				headers["Proxy-Authorization"] = "Basic " + new Buffer(`${proxyCredentials.username}:${proxyCredentials.password}`).toString('base64');
+				credentialsPart = `${proxyCredentials.username}:${proxyCredentials.password}@`;
 			}
 
+			// Note that proto ends with :
+			options.proxy = `${proto}//${credentialsPart}${host}:${port}`;
+			options.rejectUnauthorized = false;
```
rosen-vladimirov: Why is this required? Maybe this should be an option in the proxy configuration.

commands/proxy/proxy-set.ts
```diff
@@ -22,22 +23,30 @@ export class ProxySetCommand extends ProxyCommandBase {
...
+		password = password || authCredentials.password;
 
-		if (!isInteractive() && (noHostName || noPort || this.isPasswordRequired(username, password))) {
+		if (!isInteractive() && (noPort || this.isPasswordRequired(username, password))) {
```
rosen-vladimirov: I know I've written this part, but now I see an incorrect behavior here - In case I've specified port, but it's not valid and the terminal is not interactive, we'll receive error "Console is not interactive - you need to supply all command parameters."
Can you fix this by providing correct message in such case?

http-client.ts
```diff
@@ -97,65 +95,77 @@ export class HttpClient implements Server.IHttpClient {
...
+			options.url = options.url || `${options.proto}://${options.host}${options.path}`;
+			options.encoding = null;
+			options.followAllRedirects = true;
+
+			let requestObj = request(options);
+
+			requestObj
+				.on("error", (err: Error) => {
+					this.$logger.trace("An error occurred while sending the request:", err);
+					const errorMessageMatch = err.message.match(HttpClient.STATUS_CODE_REGEX);
```
rosen-vladimirov: Can you please add comment why we need this regex? You can place [this url](https://github.com/request/tunnel-agent/blob/eb2b1b19e09ee0e6a2b54eb2612755731b7301dc/index.js#L166) in the comment, so in case the response is added to the error object in a later version, we'll know where to search for it.

http-client.ts
```diff
@@ -97,65 +95,77 @@ export class HttpClient implements Server.IHttpClient {
...
+					err.message = errorMessage || err.message;
+					this.setResponseResult(promiseActions, timerId, { err });
+				})
+				.on("response", (response: Server.IRequestResponseData) => {
+					let successful = helpers.isRequestSuccessful(response);
+					if (!successful) {
+						pipeTo = undefined;
+					}
+
+					let responseStream = response;
```
rosen-vladimirov: const

http-client.ts
```diff
@@ -97,65 +95,77 @@ export class HttpClient implements Server.IHttpClient {
...
+				.on("error", (err: Error) => {
+					this.$logger.trace("An error occurred while sending the request:", err);
+					const errorMessageMatch = err.message.match(HttpClient.STATUS_CODE_REGEX);
+					const errorMessageStatusCode = errorMessageMatch && errorMessageMatch[1] && +errorMessageMatch[1];
+					let errorMessage = this.getErrorMessage(errorMessageStatusCode, null);
+					err.message = errorMessage || err.message;
+					this.setResponseResult(promiseActions, timerId, { err });
+				})
+				.on("response", (response: Server.IRequestResponseData) => {
+					let successful = helpers.isRequestSuccessful(response);
```
rosen-vladimirov: I'm not sure if it's possible to come here with something that's not 2.x.x? Have you seen such behavior?

http-client.ts
```diff
@@ -97,65 +95,77 @@ export class HttpClient implements Server.IHttpClient {
...
+					} else {
+						let data: string[] = [];
+
+						responseStream.on("data", (chunk: string) => {
+							data.push(chunk);
+						});
+
+						responseStream.on("end", () => {
+							this.$logger.trace("httpRequest: Done. code = %d", response.statusCode.toString());
+							let responseBody = data.join("");
```
rosen-vladimirov: const

commands/proxy/proxy-set.ts
Mitko-Kerezov: Theoretically there is a way for `port` to be undefined if `HttpProtocolToPort[urlObj.protocol]` returns `undefined`. This would happen in the unlikely case where the user enters an url address which makes use of neither `http` nor `https` protocol.

http-client.ts
Mitko-Kerezov: Is disputable because calling `.end()` alters the object's state but nonetheless point taken - will change it to `const`

http-client.ts
Mitko-Kerezov: `responseStream` is reassigned in case one of the `switch` cases is met on the following lines

http-client.ts
Mitko-Kerezov: Theoretically it should be possible (1xx responses for example) but I haven't seen it myself. I prefer to keep the code this way _better safe than sorry_

http-client.ts
Mitko-Kerezov: Talked offline and decided to expose a `--` flag to control this behavior

justcodebuilduser:
:broken_heart:

justcodebuilduser:
:heart:

commands/proxy/proxy-set.ts
```diff
@@ -17,25 +18,42 @@ export class ProxySetCommand extends ProxyCommandBase {
...
 		}
-
-		if (noHostName) {
-			hostname = await this.$prompter.getString("Hostname", { allowEmpty: false });
+		username = username || authCredentials.username;
+		password = password || authCredentials.password;
+
+		if (!isInteractive()) {
+			if (noPort) {
+				this.$errors.failWithoutHelp(`The port you have specified (${port || "none"}) is not valid.`);
```
rosen-vladimirov: `The port you have specified none is not valid`

This seems like a strange error message.

Mitko-Kerezov: It'll actually say
`The port you have specified (none) is not valid`

Moreover it'll be very difficult to get to this error message 😄

PR #930, opened 2017-04-03
Increase discover iOS devices timeout

yyosifov:
For older iOS devices, they cannot be enlisted by the:

`tns device ios`

command, as they're discovered slower. I've tested with 3.0 seconds which is double the current value and it seems to have worked every time for an old iPad of ours. 

Best,
Yosif


justcodebuilduser:
:heart:

PR #931, opened 2017-04-04
Adding tick-tack mechanism to incrementally wait for devices

yyosifov:
In order to improve the hard-coded "wait for 5 seconds" timeout and then check if some iOS Devices have been detected, we change to iteratively checking for iOS Device results on every 0.5 seconds, so that the 5 seconds would be a worst case for very very slow iOS devices or no iOS devices connected at all.


justcodebuilduser:
:broken_heart:

justcodebuilduser:
:broken_heart:

justcodebuilduser:
:heart:

PR #932, opened 2017-04-04
Validate proxy url

Mitko-Kerezov:
* Prompt the user for a full url in case an incomplete or incorrect one is passed (for example 127.0.0.1 instead of http://127.0.0.1
* Fix UI bug in `proxy get` command where `:` is displayed twice after protocol (i.e. http:://127.0.0.1)
* Print proxy info after `proxy set`

### DIFF
--- a/services/proxy-service.ts
+++ b/services/proxy-service.ts
@@ -1,4 +1,5 @@
 import * as path from "path";
+import { EOL } from "os";
 import { Proxy } from "../constants";
 
 export class ProxyService implements IProxyService {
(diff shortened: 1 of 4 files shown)


justcodebuilduser:
:heart:

services/proxy-service.ts
```diff
@@ -27,6 +28,24 @@ export class ProxyService implements IProxyService {
 		this.$credentialsService.clearCredentials(this.credentialsKey);
 	}
 
+	public async getInfo(): Promise<string> {
```
rosen-vladimirov: I would expect getInfo to return JSON object with information, not a string. This way we can return it to Fusion for example.

rosen-vladimirov: You can rename it or make it return an object

Mitko-Kerezov: The object you are talking about which you'd like to return to fusion can be obtained through `getCache` method.
I can rename the `getInfo` to `getStringRepresentation` for example if the name troubles you but I see no point in this method returning an object because `getCache` does exactly that.
Moreover, if it returns an object I'd have to duplicate the printing logic (i.e. printing credentials if necessary)

rosen-vladimirov: I would name this `getInfoMessage` but I do not insist. Go, go, go

ndinev:
👍

PR #933, opened 2017-04-04
Fix proxy usage on macOS and Linux and fix using npm proxy

rosen-vladimirov:
Fix using proxy on macOS and Linux - the getCredentials method should not fail - it should return null on OSes that are not supported.
Fix using npm proxy - CLI automatically detects the proxy and uses it for calls to npm registry. However we have not set the rejectUnauthorized variable and all requests to our internal npm proxy were failing. So in such cases set the rejectUnauthorized to the value of strict-ssl npm config variable.

### DIFF
--- a/appbuilder/services/npm-service.ts
+++ b/appbuilder/services/npm-service.ts
@@ -1,7 +1,7 @@
 import * as path from "path";
 import * as os from "os";
 import * as constants from "../../constants";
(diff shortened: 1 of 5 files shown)


justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

ndinev:
👍

PR #935, opened 2017-04-05
Fix proxy clear error on non Windows OS

Mitko-Kerezov:
Trace the error instead of throwing an exception.

Ping @rosen-vladimirov


justcodebuilduser:
:heart:

ndinev:
👍

PR #937, opened 2017-04-05
Fix AppBuilder unit tests

Mitko-Kerezov:
* Throw more meaningful error when trying to start ios simulator on non-Darwin OS
* Do not throw inside `resolveEmulatorServices` method - one should still be able to resolve on non-Darwin as long as one does not start an emulator

Ping @rosen-vladimirov @Plamen5kov


justcodebuilduser:
:heart:

PR #938, opened 2017-04-05
Introduce QR service

Mitko-Kerezov:
This is needed by fusion as we need to display QR codes there


justcodebuilduser:
:heart:

PR #940, opened 2017-04-07
Add event emitter helper method

Mitko-Kerezov:
Its purpose is to attach an event handler, await an action and detach the event handler

### DIFF
--- a/helpers.ts
+++ b/helpers.ts
@@ -338,13 +338,13 @@ export function isPromise(candidateFuture: any): boolean {
 	return !!(candidateFuture && typeof (candidateFuture.then) === "function");
 }
 
-export async function attachAwaitDetach(eventName: string, ee: EventEmitter, cb: Function, operation: Promise<any>) {
-	ee.on(eventName, cb);
+export async function attachAwaitDetach(eventName: string, eventEmitter: EventEmitter, eventHandler: Function, operation: Promise<any>) {
+	eventEmitter.on(eventName, eventHandler);
 
 	try {
 		await operation;
 	} finally {
-		ee.removeListener(eventName, cb);
+		eventEmitter.removeListener(eventName, eventHandler);
 	}
 }
 
--- a/helpers.ts
+++ b/helpers.ts
@@ -3,6 +3,7 @@ import * as net from "net";
 let Table = require("cli-table");
 import { platform, EOL } from "os";
 import { ReadStream } from "tty";
+import { EventEmitter } from "events";
 
 export function deferPromise<T>(): IDeferPromise<T> {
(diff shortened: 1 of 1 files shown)


justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

helpers.ts
```diff
@@ -337,6 +338,16 @@ export function isPromise(candidateFuture: any): boolean {
 	return !!(candidateFuture && typeof (candidateFuture.then) === "function");
 }
 
+export async function attachAwaitDetach(eventName: string, ee: EventEmitter, cb: Function, operation: Promise<any>) {
```
yyosifov: for readability I wouldn't name the variables like this in an exported function - ee/cb

helpers.ts
```diff
@@ -337,6 +338,16 @@ export function isPromise(candidateFuture: any): boolean {
...
+	ee.on(eventName, cb);
```
yyosifov: I was wondering, this looks a lot like the `using(x: IDisposable) { }` pattern in C# and `try-with-resources` pattern in Java, do you think it would be appropriate for us to add a Utility method which can handle it following the same pattern?

I imagine usage be like:

`
using(new DisposableEventsWrapper(eventName, eventsCallback), (platformData.projectRoot, projectData, buildConfig) => {
    await platformData.platformProjectService.buildProject(platformData.projectRoot, projectData, buildConfig))
}
`

The purpose would be to be more-readable and reusable? We'll have to add IDisposable/using patterns and implement a DisposableEventsWrapper for generic event handling.

What do you think? It can be perceived as an over-engineering in the current case?

I've found something similar here: [https://gist.github.com/dsherret/cf5d6bec3d0f791cef00](https://gist.github.com/dsherret/cf5d6bec3d0f791cef00)

justcodebuilduser:
:heart:

helpers.ts
Mitko-Kerezov: I think this might be a bit of an overkill for this particular task. I really do like the idea though, in fact I've considered something similar in the past but never got around to implementing it.

A concern of mine is the fact that we already have `IDisposable` interface and a mechanism in our dependency injection for calling `.dispose()`. However you certainly wouldn't want to call `.dispose()` on the classes that currently implement `IDisposable` yourself. That would create a bit of confusion IMO.

helpers.ts
Mitko-Kerezov: Fixed

Mitko-Kerezov:
Ping @yyosifov with fixed comments.

Had to push force the branch because of a minor merge conflict - sorry for that

justcodebuilduser:
:heart:

PR #941, opened 2017-04-07
Pass promise to connectEventually method

rosen-vladimirov:
The connectEventually method should use a promise for its factory parameter. Make the method async and pass a method that returns a promise.


justcodebuilduser:
:heart:

PR #942, opened 2017-04-10
Add helper method for getting PID of app from iOS Sim logs

rosen-vladimirov:
Add new helper method to get the process ID of an application from iOS Simulator Logs. Whenever we start an application on iOS Simulator, in its logs we can find the PID of the process in the following format:
```
<app id>: <app id>: <PID>
```
Add unit tests for the method.


justcodebuilduser:
:heart:

PR #943, opened 2017-04-12
Chmod parent dirs when syncing files on Android

Pip3r4o:
Pushing files as part of the livesync procedure on Android doesn't always update parent directories' permissions on the device unless the changes require a rebuild of the application, at which point they will be included in a list of files and directories to be chmod`d.

I propose that files being pushed in that temporary directory always have their parent's permissions explicitly set to 777 to allow their removal when necessary.

### DIFF
--- a/mobile/android/android-device-file-system.ts
+++ b/mobile/android/android-device-file-system.ts
@@ -54,8 +54,11 @@ export class AndroidDeviceFileSystem implements Mobile.IDeviceFileSystem {
 		await Promise.all(
 			_(localToDevicePaths)
 				.filter(localToDevicePathData => this.$fs.getFsStats(localToDevicePathData.getLocalPath()).isFile())
-				.map(async localToDevicePathData =>
-					await this.adb.executeCommand(["push", localToDevicePathData.getLocalPath(), localToDevicePathData.getDevicePath()])
+				.map(async localToDevicePathData => {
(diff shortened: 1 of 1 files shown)


justcodebuilduser:
:heart:

mobile/android/android-device-file-system.ts
```diff
@@ -54,8 +54,13 @@ export class AndroidDeviceFileSystem implements Mobile.IDeviceFileSystem {
 		await Promise.all(
 			_(localToDevicePaths)
 				.filter(localToDevicePathData => this.$fs.getFsStats(localToDevicePathData.getLocalPath()).isFile())
-				.map(async localToDevicePathData =>
-					await this.adb.executeCommand(["push", localToDevicePathData.getLocalPath(), localToDevicePathData.getDevicePath()])
+				.map(async localToDevicePathData => {
+					const devicePath = localToDevicePathData.getDevicePath();
+					let sections = devicePath.split("/");
+					let parentDir = sections.slice(0, sections.length - 1);
```
rosen-vladimirov: instead of splitting and joining, I believe you can use `path.dirname(devicePath)` - have you tried it?

Pip3r4o: I have not, thanks for making me aware of it!

justcodebuilduser:
:heart:

PR #946, opened 2017-04-19
Introduce utimes to fs

Mitko-Kerezov:

### DIFF
--- a/file-system.ts
+++ b/file-system.ts
@@ -55,6 +55,10 @@ export class FileSystem implements IFileSystem {
 		});
 	}
 
+	public utimes(path: string, atime: Date, mtime: Date): void {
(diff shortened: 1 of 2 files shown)


file-system.ts
```diff
@@ -55,6 +55,10 @@ export class FileSystem implements IFileSystem {
 		});
 	}
 
+	public utimes(path: string, atime: Date, mtime: Date): void {
+		return fs.utimesSync(path, atime, mtime);
```
rosen-vladimirov: return + void type seems strange

Mitko-Kerezov: this is done as a guard against future behavior changes - in case the original `fs.utimesSync` returns something on previous or future versions of node.js we needn't change the code here

justcodebuilduser:
:heart:

PR #947, opened 2017-04-21
Fix error messages to be consistent.

yyosifov:
Fix the errors for not found device to be consistent everywhere.


justcodebuilduser:
:heart:

PR #948, opened 2017-04-24
Expose service for configuration

Mitko-Kerezov:
Currently only the user agent is configurable.

Ping @rosen-vladimirov
P.S. I'm open to suggestions regarding the name of the service

### DIFF
--- a/declarations.d.ts
+++ b/declarations.d.ts
@@ -63,6 +63,29 @@ interface IiTunesConnectApplication {
 }
 
 /**
+ * Describes configuration settings that modify the behavior of some methods.
+ */
+interface IConfigurationSettings {
+	/**
+	 * This string will be used when constructing the UserAgent http header.
+	 * @type {string}
+	 */
+	userAgentName: string;
+}
+
+/**
+ * Describes service used to confugure various settings.
+ */
+interface ISettingsService {
+	/**
+	 * Used to set various settings in order to modify the behavior of some methods.
+	 * @param {IConfigurationSettings} settings Settings which will modify the behaviour of some methods.
+	 * @returns {void}
+	 */
+	setSettings(settings: IConfigurationSettings): void;
+}
+
+/**
  * Describes data returned from querying itunes' Content Delivery api
  */
 interface IContentDeliveryBody {
--- a/services/settings-service.ts
+++ b/services/settings-service.ts
@@ -1,0 +1,14 @@
+import { exported } from "../decorators";
+
+export class SettingsService implements ISettingsService {
+	constructor(private $staticConfig: Config.IStaticConfig) { }
+
+	@exported("settingsService")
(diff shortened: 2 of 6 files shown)


justcodebuilduser:
:heart:

services/configuration-service.ts
```diff
@@ -0,0 +1,12 @@
+import { exported } from "../decorators";
+
+export class ConfigurationService implements IConfigurationService {
+	constructor(private $staticConfig: Config.IStaticConfig) { }
+
+	@exported("configurationService")
+	set(settings: IConfigurationSettings): void {
+		this.$staticConfig.USER_AGENT_NAME = settings.userAgentName;
```
rosen-vladimirov: what if `userAgentName` is undefined? Should you set it?

declarations.d.ts
```diff
@@ -63,6 +63,29 @@ interface IiTunesConnectApplication {
...
 
 /**
+ * Describes configuration settings that modify the behavior of some methods.
+ */
+interface IConfigurationSettings {
+	/**
+	 * This string will be used when constructing the UserAgent http header.
+	 * @type {string}
+	 */
+	userAgentName: string;
```
rosen-vladimirov: it's used for constructing `User-Agent` but the usage suppose you'll pass `name` of the client, shouldn't we change it to `clientName`?
For example in the future we may use this setting for other messages

services/configuration-service.ts
```diff
@@ -0,0 +1,12 @@
...
+	}
+}
+
+$injector.register("configurationService", ConfigurationService);
```
rosen-vladimirov: Can we name this `settings`? The usage will be:
```JavaScript
require("nativescript").settings.set({ userAgentName: "myFancyClient" })
```

justcodebuilduser:
:heart:

declarations.d.ts
Mitko-Kerezov: I value the semantics over the usage - semantically you're setting part of the User-Agent http header and not the clientName.
Besides we've already got a `CLIENT_NAME` property in the `$staticConfig` and it would be weird to reuse the same lexical word for a different semantic meaning

services/settings-service.ts
```diff
@@ -0,0 +1,17 @@
...
+	constructor(private $staticConfig: Config.IStaticConfig,
+		private $httpClient: Server.IHttpClient) { }
+
+	@exported("settingsService")
+	setSettings(settings: IConfigurationSettings): void {
+		if (settings.userAgentName) {
+			this.$staticConfig.USER_AGENT_NAME = settings.userAgentName;
+		}
+
+		this.$httpClient.httpRequest({ url: "http://registry.npmjs.org/appbuilder" });
```
rosen-vladimirov: uhm... ? what ?

justcodebuilduser:
:heart:

PR #951, opened 2017-04-25
Add generic method to injector's .d.ts

rosen-vladimirov:
Add Generics for two of the methods in the injector, which will allow us to skip explicit casts. Most useful in tests. For examle:
Before:
```TypeScript
const a: ISomeInterface = testInejctor.resolve("a");
```

Now:
```TypeScript
const a = testInejctor.resolve<ISomeInterface>("a");
```


justcodebuilduser:
:heart:

PR #952, opened 2017-05-04
Try catch printCommandHelp

Mitko-Kerezov:
In case `printCommandHelp` fails (when for example CLI doesn't have permissions to write to the `--profile-dir`) we get an uncaught exception.

Ping @yyosifov @rosen-vladimirov

### DIFF
--- a/declarations.d.ts
+++ b/declarations.d.ts
@@ -422,6 +422,7 @@ interface ICommandOptions {
 }
 
 declare const enum ErrorCodes {
+	UNCAUGHT = 120,
 	UNKNOWN = 127,
 	INVALID_ARGUMENT = 128,
 	RESOURCE_PROBLEM = 129,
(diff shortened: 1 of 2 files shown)


justcodebuilduser:
:heart:

errors.ts
```diff
@@ -154,7 +154,11 @@ export class Errors implements IErrors {
 				: "\x1B[31;1m" + ex.message + "\x1B[0m");
 
 			if (!ex.suppressCommandHelp) {
-				await printCommandHelp();
+				try {
+					await printCommandHelp();
+				} catch (printHelpException) {
+					console.log("Failed to display command help", printHelpException);
```
yyosifov: shall we use console.error in this case?

declarations.d.ts
```diff
@@ -422,6 +422,7 @@ interface ICommandOptions {
 }
 
 declare const enum ErrorCodes {
+	UNCAUGHT = 120,
```
yyosifov: This will be used in fusion/cli?

justcodebuilduser:
:heart:

PR #953, opened 2017-05-04
Pass correct args to ios-sim-portable's stopApplication

rosen-vladimirov:
When Xcode 8 or later is installed, ios-sim-portable will spawn `xcrun simctl terminate <device id> <app id>` in order to stop the application.
For earlier version the command for stopping the app is `killall <app name>`. So pass correct args to the method.

NOTE: Merge only after the ios-sim-portable version is bumped (i.e. after merging https://github.com/telerik/ios-sim-portable/pull/90).


justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

PR #956, opened 2017-05-10
Use qr-image for qr code generation

Mitko-Kerezov:
It provides a larger variety of options, has no dependencies and is overall more robust

Try and maintain the same behavior regarding the `qr` service.


justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

PR #957, opened 2017-05-11
Add file.exe

nadyaA:
We need this, so we can use it to examine
for static and dynamic libs in the binaries while preparing


rosen-vladimirov:
This should be in NativeScript CLI only.

nadyaA:
Okay closing this fabulous PR 🙍

justcodebuilduser:
:heart:

PR #958, opened 2017-05-11
Fix UserId passing to the Eqatec

yyosifov:
When passing UserId to the Eqatec we're currently passing the Promise. We should pass the awaited promise though, as the eqatec library works with string/guid and doesn't work with the promise, therefore all our 3.x users are saved under the same UserId currently.

### DIFF
--- a/services/analytics-service-base.ts
+++ b/services/analytics-service-base.ts
@@ -194,7 +194,7 @@ export class AnalyticsServiceBase implements IAnalyticsService {
 		this._eqatecMonitor.setInstallationID(guid);
 
 		try {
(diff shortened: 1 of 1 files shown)


yyosifov:
pasted image at 2017_05_11 11_45 am

justcodebuilduser:
:heart:

yyosifov:
@rosen-vladimirov I think we should include this in 3.0.1? Although, we have Installation Id which is unique, the AnonymousId might be used for "new users" statistic which would be wrong now.

rosen-vladimirov:
You can target release branch

yyosifov:
@rosen-vladimirov for `release` branch you mean in cli, right?

rosen-vladimirov:
no, release branch here

PR #959, opened 2017-05-11
Fix UserId passing to the Eqatec

yyosifov:
When passing UserId to the Eqatec we're currently passing the Promise. We should pass the awaited promise though, as the eqatec library works with string/guid and doesn't work with the promise, therefore all our 3.x users are saved under the same UserId currently.


justcodebuilduser:
:heart:

PR #960, opened 2017-05-11
Transpile with project's tsc

Mitko-Kerezov:
In case the project has `typescript` as a dependency use it for transpilation in lieu of our own.

### DIFF
--- a/services/typescript-service.ts
+++ b/services/typescript-service.ts
@@ -4,6 +4,7 @@ import temp = require("temp");
 import { exported } from "../decorators";
 import { NODE_MODULES_DIR_NAME, FileExtensions } from "../constants";
 import { ChildProcess } from "child_process";
+import { gte } from "semver";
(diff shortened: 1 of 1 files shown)


justcodebuilduser:
:heart:

services/typescript-service.ts
```diff
@@ -169,6 +181,9 @@ export class TypeScriptService implements ITypeScriptService {
 			.concat(this.getTypeScriptCompilerOptionsAsArguments(options.compilerOptions) || [])
 			.value();
 
+		// HACK: if we're using the project's own typescript module, do not pass any additional arguments
```
TsvetanMilanov: Why?

Mitko-Kerezov: The aim of this is that in the future we'll try and not pass any arguments other than the `cwd`.
So this is temporary until we get around to that

PR #964, opened 2017-05-18
Fix posting notification to iOS Simulator

rosen-vladimirov:
In case there are spaces in path to the executable (node) or in the args, we are unable to send notification. Replace the usage of childProcess.exec with childProcess.spawn, which handles the spaces correctly.


justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

PR #965, opened 2017-05-19
Emit events comming from custom device discoveries

Mitko-Kerezov:
Make `DevicesService` extend `EventEmitter` and emit `deviceLost`/`deviceFound` events so that `DeviceEmitter` can emit them ahead.


justcodebuilduser:
:heart:

PR #966, opened 2017-05-19
Adding isBuildForDevice flag

yyosifov:
We need this flag, so that we don't fire up the Simulator/Emulator when we are executing:

`tns build {platform} --forDevice`

as obviously we don't need the Simulator/Emulator and we need to build ipa/apk for real device.


justcodebuilduser:
:heart:

PR #967, opened 2017-05-23
Remove .bash_profile alter for autocomplete #1358

yyosifov:
1. When we turn the autocomplete ON, for users which don't have `.bash_profile` we create one and add `source ~/.tnsrc` there.
2. This breaks users without `.bash_profile` files, as their Terminal has automatically been picking `.bashrc` and now it's not `sourced` in `.bash_profile`, therefore they lose their environment variables which is bad.
3. So we're left sourcing the .tnsrc file only in `.bashrc` and `.zshrc`. In case user has bash_profile, he would have sourced the `.bashrc` there.


justcodebuilduser:
:heart:

PR #968, opened 2017-05-25
Fix deviceLogs emitted for all attached ios devices

Mitko-Kerezov:
Check if device identifier matches and emit only if it does.


justcodebuilduser:
:heart:

PR #969, opened 2017-05-25
Save all unexistent folders and chown them after creation

yyosifov:
If the user didn't have ~/.local/share/.nativescript-cli we create it for them and if the setup is ran with `sudo` the owner of all these directories would be root.

Now we track what will be created and after the creation we `chown` them.

### DIFF
--- a/services/user-settings-service.ts
+++ b/services/user-settings-service.ts
@@ -49,7 +49,7 @@ export class UserSettingsServiceBase implements IUserSettingsService {
 
 				this.$fs.writeFile(this.userSettingsFilePath, null);
 
-				// when running under 'sudo' we create the /.local/share dir with root as owner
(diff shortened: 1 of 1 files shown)


justcodebuilduser:
:heart:

yyosifov:
@rosen-vladimirov I hate to write `while (true) {`. Do you think it's OK here?

services/user-settings-service.ts
```diff
@@ -43,10 +45,34 @@ export class UserSettingsServiceBase implements IUserSettingsService {
 	public async loadUserSettingsFile(): Promise<void> {
 		if (!this.userSettingsData) {
 			if (!this.$fs.exists(this.userSettingsFilePath)) {
+				let unexistingDirs = this.getUnexistingDirectories(this.userSettingsFilePath);
```
rosen-vladimirov: Ignore my comment, we have to check before we create the file...

justcodebuilduser:
:heart:

PR #970, opened 2017-05-26
Do not create debug & release config files by default

nadyaA:
* This is breaking {N} CLI
	* Not necessary for Cordova as they come from templates. Also if
one deletes them, it is expected to get error.

http://teampulse.telerik.com/view#item/345348

ping @Mitko-Kerezov @TsvetanMilanov


rosen-vladimirov:
If we want this fix in next AppBuilder CLI release, you should target release-appbuilder branch.

justcodebuilduser:
:heart:

PR #972, opened 2017-06-01
Fix analytics

rosen-vladimirov:
Fix analytics, so we'll no longer track information for users, who have not allowed us to do so. There are several problems:
* Whenever the terminal is not interactive and we do not know if user allows us to track them, we set the value of TrackExceptions to true. Also we immediately send information to Analytics that we've set the value to true. So all CI builds are tracked as new users, as we start a session for them and send the value of TrackExceptions (enabled).
* `setStatus` method almost always sends data to Analytics. This method should not send information. It does not care if we have allowed to be tracked - it always sends data. Remove this logic.
* `tns usage-reporting <value>` and `tns error-reporting <values>` always send information to Analytics - in fact the commands are calling "setStatus", that's why they send information. In order to track them, use the "trackFeature" method which will track information ONLY when user had allowed this.
* Fix `getStatus` method which had incorrect behavior - the if has been using the values of the enum (0,1,2) and we were getting in the body of the if statement in a very inaccurate cases.
* Remove unit tests for `setStatus` and add new ones for checkConsent logic.

### DIFF
--- a/test/unit-tests/analytics-service.ts
+++ b/test/unit-tests/analytics-service.ts
@@ -348,41 +348,6 @@ describe("analytics-service", () => {
 			assert.isTrue(isEqatecStopCalled);
 		});
 
-		it("tracks that user had disabled feature tracking", async () => {
-			let testInjector = createTestInjector(baseTestScenario);
-			service = testInjector.resolve("analyticsService");
-			let staticConfig: Config.IStaticConfig = testInjector.resolve("staticConfig");
-			await service.setStatus(staticConfig.TRACK_FEATURE_USAGE_SETTING_NAME, false);
-			assert.isTrue(savedSettingNamesAndValues.indexOf(`${staticConfig.TRACK_FEATURE_USAGE_SETTING_NAME}.false`) !== -1);
-			assert.isTrue(trackedFeatureNamesAndValues.indexOf(`${staticConfig.TRACK_FEATURE_USAGE_SETTING_NAME}.disabled`) !== -1);
-		});
-
-		it("tracks that user had enabled feature tracking", async () => {
-			let testInjector = createTestInjector(baseTestScenario);
-			service = testInjector.resolve("analyticsService");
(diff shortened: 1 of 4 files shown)


rosen-vladimirov:
run ci

justcodebuilduser:
:heart:

justcodebuilduser:
:heart:

test/unit-tests/analytics-service.ts
```diff
@@ -447,6 +412,17 @@ describe("analytics-service", () => {
 			assert.isTrue(savedSettingNamesAndValues.indexOf(`${staticConfig.TRACK_FEATURE_USAGE_SETTING_NAME}.true`) !== -1);
 		});
 
+		it("enables exception tracking when user confirms feature tracking and exception trackin is not set before that", async () => {
```
Mitko-Kerezov: `trackin` -> `tracking`

test/unit-tests/analytics-service.ts
```diff
@@ -458,25 +434,39 @@ describe("analytics-service", () => {
 			assert.isTrue(savedSettingNamesAndValues.indexOf(`${staticConfig.TRACK_FEATURE_USAGE_SETTING_NAME}.false`) !== -1);
 		});
 
-		it("enables exception tracking when it is not set", async () => {
-			baseTestScenario.featureTracking = false;
+		it("disables exception tracking when user confirms feature tracking and exception tracking is not set before that", async () => {
```
Mitko-Kerezov: I believe the user actually rejects feature tracking here

test/unit-tests/analytics-service.ts
```diff
@@ -485,6 +475,15 @@ describe("analytics-service", () => {
 			assert.deepEqual(savedSettingNamesAndValues, "");
 		});
 
+		it("does nothing when values are not set, but console is not interactive", async () => {
```
Mitko-Kerezov: the `but` here could be an `and`

justcodebuilduser:
:heart:

PR #977, opened 2017-06-20
Fix not listing iOS devices and confusing messages during reinstall app

rosen-vladimirov:
## Fix `tns devices` not listing attached iOS devices
When `skipDeviceDetectionInterval` option is not passed to `devicesService`, it does not wait the default timeout (6 seconds) for looking for iOS devices.
It takes some time to detect the devices, but we resolve the promise immediately (as the deviceDetection interval will report the new devices later).
However `tns devices` command is a short-living process and it exits before receiving the iOS devices. Fix the command by passing `skipDeviceDetectionInterval` option.

## Do not call uninstall applicaiton when app is not installed
In reinstallApplication we always call uninstallApplication even when the app is not installed. This produces confusing warnings on stdout.
Add check and call uninstallApplication only when app is installed.

### DIFF
--- a/commands/device/list-devices.ts
+++ b/commands/device/list-devices.ts
@@ -17,7 +17,7 @@ export class ListDevicesCommand implements ICommand {
 
 		this.$logger.out("\nConnected devices & emulators");
 		let index = 1;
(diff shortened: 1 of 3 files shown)


justcodebuilduser:
:heart:

PR #978, opened 2017-06-21
Add build method to plist module

rosen-vladimirov:
plist module has build event that is missing from our .d.ts


justcodebuilduser:
:heart:

PR #979, opened 2017-06-23
Rename isBuildForDevice option to skipEmulatorStart

rosen-vladimirov:
DevicesService.initialize method has option called `isBuildForDevice`. Its real purpose is to disable start of simulator, so rename it and use it on correct places.

### DIFF
--- a/definitions/mobile.d.ts
+++ b/definitions/mobile.d.ts
@@ -355,9 +355,9 @@ declare module Mobile {
 		 */
 		skipDeviceDetectionInterval?: boolean;
 		/**
-		 * Specifies whether this is a ForDevice build.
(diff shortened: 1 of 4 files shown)


definitions/mobile.d.ts
```diff
@@ -357,7 +357,7 @@ declare module Mobile {
 		/**
 		 * Specifies whether this is a ForDevice build.
```
TsvetanMilanov: I think we should change the description too.

rosen-vladimirov:
run ci

justcodebuilduser:
:broken_heart:

justcodebuilduser:
:broken_heart:

PR #980, opened 2017-06-27
Update lodash types definition due to update of TS

nadyaA:
We need this custom file as the default definition requires import every
time we use lodash. Having it declared as a module allows us to use it as
global variable, required once.


justcodebuilduser:
:broken_heart:

justcodebuilduser:
:broken_heart:

justcodebuilduser:
:broken_heart:

justcodebuilduser:
:broken_heart:

PR #987, opened 2017-07-09
Make devicesService.getDevice public

rosen-vladimirov:
`devicesService.getDevice` method returns device instance based on specified device identifier or number (from the list of devices).
Make it public, so it can be used in `tns debug <platform>` command.

### DIFF
--- a/mobile/mobile-core/devices-service.ts
+++ b/mobile/mobile-core/devices-service.ts
@@ -255,7 +255,7 @@ export class DevicesService extends EventEmitter implements Mobile.IDevicesServi
 	 * Method expects running devices.
 	 * @param identifier parameter passed by the user to --device flag
 	 */
-	private async getDevice(deviceOption: string): Promise<Mobile.IDevice> {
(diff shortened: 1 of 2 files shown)


justcodebuilduser:
:broken_heart:

mobile/mobile-core/devices-service.ts
```diff
@@ -255,7 +255,7 @@ export class DevicesService extends EventEmitter implements Mobile.IDevicesServi
 	 * Method expects running devices.
 	 * @param identifier parameter passed by the user to --device flag
 	 */
-	private async getDevice(deviceOption: string): Promise<Mobile.IDevice> {
+	public async getDevice(deviceOption: string): Promise<Mobile.IDevice> {
```
nadyaA: Since the method becomes public, we may consider adding tests for it.

rosen-vladimirov: I agree that the method needs separate tests, but currently it is already covered with tests, mainly [here](https://github.com/telerik/mobile-cli-lib/blob/master/test/unit-tests/mobile/devices-service.ts#L505-L527)

PR #988, opened 2017-07-11
Do not attach to the device log on every application start.

TomaNikolov:
Cache the startDeviceLog method to ensure that we are attached only once to the device log.

[#2941](https://github.com/NativeScript/nativescript-cli/issues/2941)


justcodebuilduser:
:broken_heart:

PR #990, opened 2017-07-24
Filter android logs by process id

Pip3r4o:
### Problem
https://github.com/NativeScript/nativescript-cli/issues/2818 - When an app was started the CLI's logging filter would parse each incoming adb line and check it against a regex that either `JS:` was the tag used, or that any of the [accepted tags](https://github.com/telerik/mobile-cli-lib/compare/pete/log-by-pid-android?expand=1#diff-db4bc9bdec0f0e3438155588e7c23c1aR38)'s string was contained within the logcat line.
That would at times return incorrect, or incomplete logs.

### Proposed solution
Read the process id for the application process from adb, then filter logs by the pid.

### DIFF
--- a/mobile/android/android-application-manager.ts
+++ b/mobile/android/android-application-manager.ts
@@ -11,6 +11,7 @@ export class AndroidApplicationManager extends ApplicationManagerBase {
 		private $logcatHelper: Mobile.ILogcatHelper,
 		private $androidProcessService: Mobile.IAndroidProcessService,
 		private $httpClient: Server.IHttpClient,
+		private $deviceLogProvider: Mobile.IDeviceLogProvider,
 		$logger: ILogger,
 		$hooksService: IHooksService) {
 		super($logger, $hooksService);
@@ -49,6 +50,12 @@ export class AndroidApplicationManager extends ApplicationManagerBase {
 			"1"]);
 
 		if (!this.$options.justlaunch) {
+			const deviceIdentifier = this.identifier;
+			const processIdentifier = await this.$androidProcessService.getAppProcessId(deviceIdentifier, appIdentifier);
+			if (processIdentifier) {
+				this.$deviceLogProvider.setApplicationPidForDevice(deviceIdentifier, processIdentifier);
+			}
+
 			await this.$logcatHelper.start(this.identifier);
 		}
 	}
--- a/mobile/android/android-log-filter.ts
+++ b/mobile/android/android-log-filter.ts
@@ -7,14 +7,14 @@ export class AndroidLogFilter implements Mobile.IPlatformLogFilter {
 
 	// sample line is "11-23 12:39:07.310  1584  1597 I art     : Background sticky concurrent mark sweep GC freed 21966(1780KB) AllocSpace objects, 4(80KB) LOS objects, 77% free, 840KB/3MB, paused 4.018ms total 158.629ms"
 	// or '12-28 10:45:08.020  3329  3329 W chromium: [WARNING:data_reduction_proxy_settings.cc(328)] SPDY proxy OFF at startup'
-	private static API_LEVEL_23_LINE_REGEX = /.+?\s+?(?:[A-Z]\s+?)([A-Za-z ]+?)\s*?\: (.*)/;
+	private static API_LEVEL_23_LINE_REGEX = /.+?\s+?(?:[A-Z]\s+?)([A-Za-z \.]+?)\s*?\: (.*)/;
 
 	constructor(private $loggingLevels: Mobile.ILoggingLevels) { }
 
-	public filterData(data: string, logLevel: string): string {
-		let specifiedLogLevel = (logLevel || '').toUpperCase();
+	public filterData(data: string, logLevel: string, pid: string): string {
+		const specifiedLogLevel = (logLevel || '').toUpperCase();
 		if (specifiedLogLevel === this.$loggingLevels.info) {
-			let log = this.getConsoleLogFromLine(data);
+			const log = this.getConsoleLogFromLine(data, pid);
 			if (log) {
 				if (log.tag) {
 					return `${log.tag}: ${log.message}` + os.EOL;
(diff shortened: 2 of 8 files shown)


justcodebuilduser:
:broken_heart:

mobile/android/android-application-manager.ts
```diff
@@ -49,10 +50,39 @@ export class AndroidApplicationManager extends ApplicationManagerBase {
 			"1"]);
 
 		if (!this.$options.justlaunch) {
+			let processIdentifier = await this.getProcessId(appIdentifier);
+			if (processIdentifier) {
+				let deviceIdentifier = this.adb.identifier;
```
rosen-vladimirov: why do you use `this.adb.identifier` instead of `this.identifier`

mobile/android/android-application-manager.ts
```diff
@@ -49,10 +50,39 @@ export class AndroidApplicationManager extends ApplicationManagerBase {
...
 			await this.$logcatHelper.start(this.identifier);
 		}
 	}
 
+	private async getProcessId(appId: string): Promise<string> {
+		// Read the process id of the application that was just started
+		// Information is returned in the following format
+		// USER      PID   PPID  VSIZE  RSS   WCHAN            PC  NAME
+		// u0_a64    4831  488   851380 94888    ep_poll f1e43bb9 S org.nativescript.myApplication
+		let processes = await this.adb.executeShellCommand(["ps"]);
```
rosen-vladimirov: Instead of executing this and parsing the result, can't you use the `$androidProcessService`: https://github.com/telerik/mobile-cli-lib/blob/master/mobile/mobile-core/android-process-service.ts#L206-L218

mobile/android/android-log-filter.ts
```diff
@@ -7,14 +7,14 @@ export class AndroidLogFilter implements Mobile.IPlatformLogFilter {
 
 	// sample line is "11-23 12:39:07.310  1584  1597 I art     : Background sticky concurrent mark sweep GC freed 21966(1780KB) AllocSpace objects, 4(80KB) LOS objects, 77% free, 840KB/3MB, paused 4.018ms total 158.629ms"
 	// or '12-28 10:45:08.020  3329  3329 W chromium: [WARNING:data_reduction_proxy_settings.cc(328)] SPDY proxy OFF at startup'
-	private static API_LEVEL_23_LINE_REGEX = /.+?\s+?(?:[A-Z]\s+?)([A-Za-z ]+?)\s*?\: (.*)/;
+	private static API_LEVEL_23_LINE_REGEX = /.+?\s+?(?:[A-Z]\s+?)([A-Za-z \.]+?)\s*?\: (.*)/;
```
rosen-vladimirov: Why is this changed? I couldn't understand it from the commit message.

mobile/android/android-log-filter.ts
```diff
@@ -29,15 +29,20 @@ export class AndroidLogFilter implements Mobile.IPlatformLogFilter {
 		return data + os.EOL;
 	}
 
-	private getConsoleLogFromLine(lineText: string): any {
-		let acceptedTags = ["chromium", "Web Console", "JS"];
+	private getConsoleLogFromLine(lineText: string, pid: string): any {
+		// filter log line if it does not belong to the current application process id
+		if (lineText.indexOf(pid) < 0) {
```
rosen-vladimirov: So in case we do not have pid, we'll always return null. But in case we have PID, we'll still apply the incorrect filters?
For iOS Simulator we have similar case, but the logic there is:
- in case there's a pid, find it in the logs and show each line that contains it
- in case we do not have a PID, apply the old filtering

Can't we do the same here?

mobile/android/android-log-filter.ts
```diff
@@ -29,15 +29,20 @@ export class AndroidLogFilter implements Mobile.IPlatformLogFilter {
...
+			return null;
+		}
+
+		let acceptedTags = ["chromium", "Web Console", "JS", "ActivityManager", "System.err"];
 		let match = lineText.match(AndroidLogFilter.LINE_REGEX) || lineText.match(AndroidLogFilter.API_LEVEL_23_LINE_REGEX);
 
 		if (match && acceptedTags.indexOf(match[1].trim()) !== -1) {
 			return { tag: match[1].trim(), message: match[2] };
 		}
-		let matchingTag = _.some(acceptedTags, (tag: string) => { return lineText.indexOf(tag) !== -1; });
```
rosen-vladimirov: Why is this line deleted? Its idea is to save us in cases where the logs are changed and our regex fails, but we still check if the line contains "chromium" for example.

mobile/android/device-android-debug-bridge.ts
```diff
@@ -6,7 +6,7 @@ interface IComposeCommandResult {
 }
 
 export class DeviceAndroidDebugBridge extends AndroidDebugBridge implements Mobile.IDeviceAndroidDebugBridge {
-	constructor(private identifier: string,
+	constructor(public identifier: string,
```
rosen-vladimirov: I believe you do not need this change as the only place where you are using the property outside of the current class is on a place, where you already have the identifier.

justcodebuilduser:
:broken_heart:

mobile/android/android-application-manager.ts
```diff
@@ -49,10 +50,39 @@ export class AndroidApplicationManager extends ApplicationManagerBase {
...
+		processes = processes.trim()
+			.split("\n")
+			.map((line: string) => line.replace(/\s+/g, ' ')
+				.split(" ")
+				.filter((entry: string) => /\S/.test(entry)));
+
+		let currentProcessLine = processes.find((line: Array<string>) => line[line.length - 1] === appId);
+
+		if (!currentProcessLine) {
+			return Promise.resolve(null);
```
rosen-vladimirov: you do not need to return Promise here, `async` keyword does this for you. 
```TypeScript
return null;
```
is enough

justcodebuilduser:
:broken_heart:

justcodebuilduser:
:broken_heart:

mobile/android/android-application-manager.ts
```diff
@@ -49,6 +50,12 @@ export class AndroidApplicationManager extends ApplicationManagerBase {
 			"1"]);
 
 		if (!this.$options.justlaunch) {
+			let deviceIdentifier = this.identifier;
```
nadyaA: As I see you use 'let' in almost all cases, if there is no specific reason (if so what), you may consider updating your code. As you know declaring the variables with 'let' indicates the'll be changed and most of these are not. 😸

Pip3r4o: Noted, thank you!

justcodebuilduser:
:broken_heart:

justcodebuilduser:
:broken_heart:

mobile/android/android-log-filter.ts
rosen-vladimirov: After personal discussion - the reason is that we have added a new tag, that has `.` in it.

test/unit-tests/android-log-filter.ts
```diff
@@ -133,6 +246,21 @@ describe("androidLogFilter", () => {
...
+			it("when API level 23 or later is used, and application process matches logcat pids", () => {
+				const appPid = "8141";
+				_.each(androidApiLevel23MapForPid8141, testData => {
+					assertFiltering(testData.input, testData.output ? testData.output + EOL : testData.output, logLevel, appPid);
+				});
+			});
+
+			it("when API level 23 or later is used, and application pid doesn't match any logcat pids", () => {
+				const appPid = "99999";
+				const expectedOutputMap = androidApiLevel23MapForPid8141.map((testData) => { return { input: testData.input, output: null } });
```
rosen-vladimirov: you can shorten this:
```TypeScript
const expectedOutputMap = androidApiLevel23MapForPid8141.map(testData => ({ input: testData.input, output: null }) );
```

justcodebuilduser:
:broken_heart:

justcodebuilduser:
:broken_heart:

PR #991, opened 2017-07-25
Print message when successfully setting the tns proxy configuration

Pip3r4o:
Log a message instructing the user to run the `proxy set` help command to find out how to setup the `npm` and `Gradle` proxy configurations.

Implemented as suggested in https://github.com/NativeScript/nativescript-cli/issues/2816

Related: https://github.com/NativeScript/nativescript-cli/pull/3004

### DIFF
--- a/commands/proxy/proxy-set.ts
+++ b/commands/proxy/proxy-set.ts
@@ -18,6 +18,7 @@ export class ProxySetCommand extends ProxyCommandBase {
 		private $injector: IInjector,
 		private $prompter: IPrompter,
 		private $hostInfo: IHostInfo,
+		private $staticConfig: Config.IStaticConfig,
 		protected $analyticsService: IAnalyticsService,
 		protected $logger: ILogger,
 		protected $options: ICommonOptions,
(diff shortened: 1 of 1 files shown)


justcodebuilduser:
:broken_heart:

commands/proxy/proxy-set.ts
```diff
@@ -103,6 +103,8 @@ export class ProxySetCommand extends ProxyCommandBase {
 			this.$logger.warn(`Note that storing credentials is not supported on ${platform()} yet.`);
 		}
 
+		this.$logger.warn("Note that 'npm' and 'Gradle' need to be configured separately to work with a proxy. Run 'tns proxy set --help' for more information.");
```
rosen-vladimirov: As this code is shared between AppBuilder and NativeScript CLI, you should use `$staticConfig.CLIENT_NAME` instead of `tns` in the message.
Also `Gradle`  is not used in AppBuilder CLI.
Maybe we can place some if here to print the message only when we are in NativeScript CLI scenario.

Pip3r4o: I did not consider that scenario, you are right!

PR #992, opened 2017-07-25
Return the real transferred files when transferDirectory of device fs is called

rosen-vladimirov:
When uploading files on devices/emulators, we are executing either `transferDirectory` or `transferFiles` methods of the specific device instance.
However for Android devices, the transferDirectory method has some logic to transfer only modified files (based on shasums from previous execution). So it doesn't really transfer the directory, but just several files.
For LiveSync purposes it is important to know which are the tranferred files, so change the return type of `transferDirectory` method to return the exact files that we've sent to device.

Try-catch the stop of iOS Application as it may fail in some rare cases - in this case just retry it.

### DIFF
--- a/mobile/android/android-device-file-system.ts
+++ b/mobile/android/android-device-file-system.ts
@@ -81,9 +81,9 @@ export class AndroidDeviceFileSystem implements Mobile.IDeviceFileSystem {
 		}
 	}
 
-	public async transferDirectory(deviceAppData: Mobile.IDeviceAppData, localToDevicePaths: Mobile.ILocalToDevicePathData[], projectFilesPath: string): Promise<void> {
-		let devicePaths: string[] = [],
-			currentShasums: IStringDictionary = {};
+	public async transferDirectory(deviceAppData: Mobile.IDeviceAppData, localToDevicePaths: Mobile.ILocalToDevicePathData[], projectFilesPath: string): Promise<Mobile.ILocalToDevicePathData[]> {
+		const devicePaths: string[] = [];
+		const currentShasums: IStringDictionary = {};
 
 		await Promise.all(
 			localToDevicePaths.map(async localToDevicePathData => {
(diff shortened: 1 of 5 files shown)


justcodebuilduser:
:broken_heart:

mobile/android/android-device-file-system.ts
```diff
@@ -102,28 +102,27 @@ export class AndroidDeviceFileSystem implements Mobile.IDeviceFileSystem {
 
 		let deviceHashService = this.getDeviceHashService(deviceAppData.appIdentifier);
 		let filesToChmodOnDevice: string[] = devicePaths;
-		if (this.$options.force) {
+		let tranferredFiles: Mobile.ILocalToDevicePathData[] = [];
+		let oldShasums = await deviceHashService.getShasumsFromDevice();
```
nadyaA: const?

mobile/android/android-device-file-system.ts
```diff
@@ -102,28 +102,27 @@ export class AndroidDeviceFileSystem implements Mobile.IDeviceFileSystem {
...
-						.map(localToDevicePathData => {
-							filesToChmodOnDevice.push(`"${localToDevicePathData.getDevicePath()}"`);
-							return this.transferFile(localToDevicePathData.getLocalPath(), localToDevicePathData.getDevicePath());
-						})
-						.value()
-				);
-			} else {
-				await this.adb.executeCommand(["push", projectFilesPath, await deviceAppData.getDeviceProjectRootPath()]);
-			}
+			let changedShasums: any = _.omitBy(currentShasums, (hash: string, pathToFile: string) => !!_.find(oldShasums, (oldHash: string, oldPath: string) => pathToFile === oldPath && hash === oldHash));
```
nadyaA: const?

justcodebuilduser:
:broken_heart:

justcodebuilduser:
:broken_heart:

PR #993, opened 2017-07-26
Fix simulator started on run with fake device id.

KristianDD:
This fixes https://github.com/NativeScript/nativescript-cli/issues/2728. Before a simulator with default preference was started when using  `tns run ios --device "fake"` and nothing else happens. Now a descriptive error is show to the user.

SHOULD NOT BE MERGED BEFORE - https://github.com/telerik/ios-sim-portable/pull/92

### DIFF
--- a/mobile/mobile-core/devices-service.ts
+++ b/mobile/mobile-core/devices-service.ts
@@ -409,7 +409,7 @@ export class DevicesService extends EventEmitter implements Mobile.IDevicesServi
 	}
 
 	private async _startEmulatorIfNecessary(data?: Mobile.IDevicesServicesInitializationOptions): Promise<void> {
-			let deviceInstances = this.getDeviceInstances();
+			const deviceInstances = this.getDeviceInstances();
 
 			//if no --device is passed and no devices are found, the default emulator is started
 			if (!data.deviceId && _.isEmpty(deviceInstances)) {
@@ -688,14 +688,16 @@ export class DevicesService extends EventEmitter implements Mobile.IDevicesServi
 	}
 
 	private getEmulatorError(error: Error, platform: string): string {
-		let emulatorName = "Emulator";
+		let emulatorName = constants.DeviceTypes.Emulator;
+
 		if (this.$mobileHelper.isiOSPlatform(platform)) {
-			emulatorName = "Simulator";
+			emulatorName = constants.DeviceTypes.Simulator;
 		}
 
 		return `Cannot find connected devices.${EOL}` +
 			`${emulatorName} start failed with: ${error.message}${EOL}` +
-			`To list currently connected devices and verify that the specified identifier exists, run '${this.$staticConfig.CLIENT_NAME.toLowerCase()} device'.`;
+			`To list currently connected devices and verify that the specified identifier exists, run '${this.$staticConfig.CLIENT_NAME.toLowerCase()} device'.${EOL}` +
+			`To list available ${emulatorName.toLowerCase()} images, run '${this.$staticConfig.CLIENT_NAME.toLowerCase()} device <Platform> --available-devices'.`;
 	}
 }
 
--- a/mobile/mobile-core/devices-service.ts
+++ b/mobile/mobile-core/devices-service.ts
@@ -392,19 +392,34 @@ export class DevicesService extends EventEmitter implements Mobile.IDevicesServi
 			}
 			let deviceInstances = this.getDeviceInstances();
 
-			//if no --device is passed and no devices are found, the default emulator is started
 			if (!data.deviceId && _.isEmpty(deviceInstances)) {
 				if (!this.$hostInfo.isDarwin && this.$mobileHelper.isiOSPlatform(data.platform)) {
 					this.$errors.failWithoutHelp(constants.ERROR_NO_DEVICES_CANT_USE_IOS_SIMULATOR);
 				}
+			}
 
+			try {
+				await this._startEmulatorIfNecessary(data);
+			} catch (err) {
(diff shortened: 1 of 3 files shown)


justcodebuilduser:
:broken_heart:

justcodebuilduser:
:broken_heart:

justcodebuilduser:
:broken_heart:

mobile/mobile-core/devices-service.ts
```diff
@@ -392,19 +392,34 @@ export class DevicesService extends EventEmitter implements Mobile.IDevicesServi
...
 			if (!data.deviceId && _.isEmpty(deviceInstances)) {
 				if (!this.$hostInfo.isDarwin && this.$mobileHelper.isiOSPlatform(data.platform)) {
 					this.$errors.failWithoutHelp(constants.ERROR_NO_DEVICES_CANT_USE_IOS_SIMULATOR);
 				}
+			}
+
+			try {
+				await this._startEmulatorIfNecessary(data);
+			} catch (err) {
+				const errorMessage = this.getEmulatorError(err, data.platform);
```
Plamen5kov: Please fix the method description, because you changed the method behavior. (now it fails without help)

mobile/mobile-core/devices-service.ts
```diff
@@ -392,19 +392,34 @@ export class DevicesService extends EventEmitter implements Mobile.IDevicesServi
...
 
+				this.$errors.failWithoutHelp(errorMessage);
+			}
+		}
+	}
+
+	private async _startEmulatorIfNecessary(data?: Mobile.IDevicesServicesInitializationOptions): Promise<void> {
+			let deviceInstances = this.getDeviceInstances();
```
Plamen5kov: `deviceInstances` is already called once on line 393, so there's no need to call it again here.

mobile/mobile-core/devices-service.ts
```diff
@@ -672,6 +686,17 @@ export class DevicesService extends EventEmitter implements Mobile.IDevicesServi
 		shouldReturnImmediateResult = shouldReturnImmediateResult || false;
 		return { shouldReturnImmediateResult: shouldReturnImmediateResult, platform: platform };
 	}
+
+	private getEmulatorError(error: Error, platform: string): string {
+		let emulatorName = "Emulator";
```
Plamen5kov: please use constants.ts / DeviceTypes and add "Simulator" there

mobile/mobile-core/devices-service.ts
```diff
@@ -672,6 +686,17 @@ export class DevicesService extends EventEmitter implements Mobile.IDevicesServi
...
+		if (this.$mobileHelper.isiOSPlatform(platform)) {
+			emulatorName = "Simulator";
+		}
+
+		return `Cannot find connected devices.${EOL}` +
+			`${emulatorName} start failed with: ${error.message}${EOL}` +
+			`To list currently connected devices and verify that the specified identifier exists, run '${this.$staticConfig.CLIENT_NAME.toLowerCase()} device'.`;
```
Plamen5kov: Suggestion: `tns device` only provides the user with the information about the running devices, but the `--device` also accepts emulator image names. By that logic we can either tell the user to run `tns device android --available-devices` or make the `tns device` command to show the available emulator images.

mobile/mobile-core/devices-service.ts
KristianDD: Please help me find the place where to update the description. I can't see where inside the method description is included the way it fails. There are a few places inside the method that also fail without help (that`s the reason I made it fail without help here also).

mobile/mobile-core/devices-service.ts
KristianDD: I'd prefer that instead of passing it as an argument, but if you insist I can change it...

Plamen5kov: Please do, because it's easier to read and there's no duplicate code.

mobile/mobile-core/devices-service.ts
Plamen5kov: You're right, just describe as best you can what the method does so it's easy to understand.

mobile/mobile-core/devices-service.ts
KristianDD: I think it will be good to add t`ns devices --available-devices` as it is not visible inside specific platform. Moreover as `tns devices` lists active devices for both platform, and `tns devices ios ` for specific platform it seems consistent that both should accept `--available-devices`. @rosen-vladimirov what do you think?

mobile/mobile-core/devices-service.ts
rosen-vladimirov: I'm also a fan of getting the deviceInstances here, I see no benefit of additional args of the method in this case.
Just make it `const` instaed of `let`

mobile/mobile-core/devices-service.ts
rosen-vladimirov: I agree with the change, but you can implement it in another PR :)

mobile/mobile-core/devices-service.ts
rosen-vladimirov: I'm not a big fan of getting 10 rows of description what the method does. The name and the code should talk for itself. Some basic info should be enough. But describing that in case A the situation 1 will happen, and in case B situation 2 will happen is not correct, IMO. In case you are interested in all cases - you can always check the implementation.

mobile/mobile-core/devices-service.ts
KristianDD: Created an issue/feature request https://github.com/NativeScript/nativescript-cli/issues/3023

PR #995, opened 2017-07-31
use am start instead of monkey to start application

Plamen5kov:
_problem_
`monkey` is a tool used for testing, not for production use and we shouldn't use it since it's a tool used and built for stress testing. It has random behavior and it's not recommended for production use and therefore we are replacing `monkey` with the recommended `am` tool.
Fix Issue: https://github.com/NativeScript/NativeScript/issues/4591

_solution_
Use `am` tool instead to start the application by it's activity identifier which is a combination of the package name and the default activity. 
Example for such an identifier would be: `org.nativescript.appname/com.tns.MainActivity`

### DIFF
--- a/mobile/android/android-application-manager.ts
+++ b/mobile/android/android-application-manager.ts
@@ -2,6 +2,7 @@ import { EOL } from "os";
 import { ApplicationManagerBase } from "../application-manager-base";
 import { LiveSyncConstants, TARGET_FRAMEWORK_IDENTIFIERS } from "../../constants";
 import { hook } from "../../helpers";
+import { cache } from "../../decorators";
 
 export class AndroidApplicationManager extends ApplicationManagerBase {
 
@@ -61,11 +62,7 @@ export class AndroidApplicationManager extends ApplicationManagerBase {
 		const pmDumpOutput = await this.adb.executeShellCommand(["pm", "dump", appIdentifier, "|", "grep", "-A", "1", "MAIN"]);
 		const activityMatch = this.getFullyQualifiedActivityRegex();
 		const match = activityMatch.exec(pmDumpOutput);
-		let possibleIdentifier = "";
-
-		if (match && match.length > 0) {
-			possibleIdentifier = match[0];
-		}
+		const possibleIdentifier = match && match[0];
 
 		if (possibleIdentifier) {
 			await this.adb.executeShellCommand(["am", "start", "-n", possibleIdentifier]);
@@ -128,6 +125,7 @@ export class AndroidApplicationManager extends ApplicationManagerBase {
 		return applicationViews;
 	}
 
+	@cache()
 	private getFullyQualifiedActivityRegex(): RegExp {
 		const androidPackageName = "([A-Za-z]{1}[A-Za-z\\d_]*\\.)*[A-Za-z][A-Za-z\\d_]*";
 		const packageActivitySeparator = "\\/";
--- a/mobile/android/android-application-manager.ts
+++ b/mobile/android/android-application-manager.ts
@@ -45,14 +45,19 @@ export class AndroidApplicationManager extends ApplicationManagerBase {
 
 	public async startApplication(appIdentifier: string): Promise<void> {
 		const pmDumpOutput = await this.adb.executeShellCommand(["pm", "dump", appIdentifier, "|", "grep", "-A", "1", "MAIN"]);
-		const fullActivityNameRegExp = /([A-Za-z]{1}[A-Za-z\d_]*\.)*[A-Za-z][A-Za-z\d_]*\/([a-z][a-z_0-9]*\.)*[A-Z_]($[A-Z_]|[\w_])*/;
+		const fullActivityNameRegExp = this.getFullyQualifiedActivityRegex();
 		const activityMatch = new RegExp(fullActivityNameRegExp, "m");
-		const possibleIdentifier = activityMatch.exec(pmDumpOutput)[0];
+		const match = activityMatch.exec(pmDumpOutput);
+		let possibleIdentifier = "";
+
+		if(match && match.length > 0) {
+			possibleIdentifier = match[0]
+		}
 
 		if (possibleIdentifier) {
 			await this.adb.executeShellCommand(["am", "start", "-n", possibleIdentifier]);
 		} else {
-			this.$logger.trace(`Tried starting activity: ${possibleIdentifier}, but failed`);
+			this.$logger.trace(`Tried starting activity for: ${appIdentifier}, but failed`);
 		}
 
 		if (!this.$options.justlaunch) {
@@ -107,5 +112,9 @@ export class AndroidApplicationManager extends ApplicationManagerBase {
 		}));
 
 		return applicationViews;
+	}
+
+	public getFullyQualifiedActivityRegex(): RegExp {
+		return /([A-Za-z]{1}[A-Za-z\d_]*\.)*[A-Za-z][A-Za-z\d_]*\/([a-z][a-z_0-9]*\.)*[A-Z_$]($[A-Z_$]|[$_\w_])*/;
 	}
 }
--- a/mobile/android/android-application-manager.ts
+++ b/mobile/android/android-application-manager.ts
@@ -44,10 +44,16 @@ export class AndroidApplicationManager extends ApplicationManagerBase {
 	}
 
 	public async startApplication(appIdentifier: string): Promise<void> {
-		await this.adb.executeShellCommand(["monkey",
-			"-p", appIdentifier,
-			"-c", "android.intent.category.LAUNCHER",
-			"1"]);
+		const pmDumpOutput = await this.adb.executeShellCommand(["pm", "dump", appIdentifier, "|", "grep", "-A", "1", "MAIN"]);
+		const fullActivityNameRegExp = /([A-Za-z]{1}[A-Za-z\d_]*\.)*[A-Za-z][A-Za-z\d_]*\/([a-z][a-z_0-9]*\.)*[A-Z_]($[A-Z_]|[\w_])*/;
+		const activityMatch = new RegExp(fullActivityNameRegExp, "m");
+		const possibleIdentifier = activityMatch.exec(pmDumpOutput)[0];
+
+		if (possibleIdentifier) {
+			await this.adb.executeShellCommand(["am", "start", "-n", possibleIdentifier]);
+		} else {
+			this.$logger.trace(`Tried starting activity: ${possibleIdentifier}, but failed`);
+		}
 
 		if (!this.$options.justlaunch) {
 			const deviceIdentifier = this.identifier;
--- a/mobile/android/android-application-manager.ts
+++ b/mobile/android/android-application-manager.ts
@@ -44,9 +44,22 @@ export class AndroidApplicationManager extends ApplicationManagerBase {
 	}
 
 	public async startApplication(appIdentifier: string): Promise<void> {
+
+		/*
+		Example "pm dump <app_identifier> | grep -A 1 MAIN" output"
+			android.intent.action.MAIN:
+			3b2df03 org.nativescript.cliapp/com.tns.NativeScriptActivity filter 50dd82e
+			Action: "android.intent.action.MAIN"
+			Category: "android.intent.category.LAUNCHER"
+			--
+			intent={act=android.intent.action.MAIN cat=[android.intent.category.LAUNCHER] flg=0x10200000 cmp=org.nativescript.cliapp/com.tns.NativeScriptActivity}
+			realActivity=org.nativescript.cliapp/com.tns.NativeScriptActivity
+			--
+			Intent { act=android.intent.action.MAIN cat=[android.intent.category.LAUNCHER] flg=0x10200000 cmp=org.nativescript.cliapp/com.tns.NativeScriptActivity }
+			frontOfTask=true task=TaskRecord{fe592ac #449 A=org.nativescript.cliapp U=0 StackId=1 sz=1}
+		*/
 		const pmDumpOutput = await this.adb.executeShellCommand(["pm", "dump", appIdentifier, "|", "grep", "-A", "1", "MAIN"]);
-		const fullActivityNameRegExp = this.getFullyQualifiedActivityRegex();
-		const activityMatch = new RegExp(fullActivityNameRegExp, "m");
+		const activityMatch = this.getFullyQualifiedActivityRegex();
 		const match = activityMatch.exec(pmDumpOutput);
 		let possibleIdentifier = "";
 
@@ -115,7 +128,11 @@ export class AndroidApplicationManager extends ApplicationManagerBase {
 		return applicationViews;
 	}
 
-	public getFullyQualifiedActivityRegex(): RegExp {
-		return /([A-Za-z]{1}[A-Za-z\d_]*\.)*[A-Za-z][A-Za-z\d_]*\/([a-z][a-z_0-9]*\.)*[A-Z_$]($[A-Z_$]|[$_\w_])*/;
+	private getFullyQualifiedActivityRegex(): RegExp {
+		const androidPackageName = "([A-Za-z]{1}[A-Za-z\\d_]*\\.)*[A-Za-z][A-Za-z\\d_]*";
+		const packageActivitySeparator = "\\/";
+		const fullJavaClassName = "([a-z][a-z_0-9]*\\.)*[A-Z_$]($[A-Z_$]|[$_\\w_])*";
+
+		return new RegExp(`${androidPackageName}${packageActivitySeparator}${fullJavaClassName}`, `m`);
 	}
 }
--- a/mobile/android/android-application-manager.ts
+++ b/mobile/android/android-application-manager.ts
@@ -50,14 +50,15 @@ export class AndroidApplicationManager extends ApplicationManagerBase {
 		const match = activityMatch.exec(pmDumpOutput);
 		let possibleIdentifier = "";
 
-		if(match && match.length > 0) {
-			possibleIdentifier = match[0]
+		if (match && match.length > 0) {
+			possibleIdentifier = match[0];
 		}
 
 		if (possibleIdentifier) {
 			await this.adb.executeShellCommand(["am", "start", "-n", possibleIdentifier]);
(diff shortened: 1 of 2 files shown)


mobile/android/android-application-manager.ts
```diff
@@ -44,10 +44,16 @@ export class AndroidApplicationManager extends ApplicationManagerBase {
 	}
 
 	public async startApplication(appIdentifier: string): Promise<void> {
-		await this.adb.executeShellCommand(["monkey",
-			"-p", appIdentifier,
-			"-c", "android.intent.category.LAUNCHER",
-			"1"]);
+		const pmDumpOutput = await this.adb.executeShellCommand(["pm", "dump", appIdentifier, "|", "grep", "-A", "1", "MAIN"]);
+		const fullActivityNameRegExp = /([A-Za-z]{1}[A-Za-z\d_]*\.)*[A-Za-z][A-Za-z\d_]*\/([a-z][a-z_0-9]*\.)*[A-Z_]($[A-Z_]|[\w_])*/;
```
TsvetanMilanov: Can you provide some example output of `pm dump`?

mobile/android/android-application-manager.ts
```diff
@@ -44,10 +44,16 @@ export class AndroidApplicationManager extends ApplicationManagerBase {
...
+		const activityMatch = new RegExp(fullActivityNameRegExp, "m");
+		const possibleIdentifier = activityMatch.exec(pmDumpOutput)[0];
```
TsvetanMilanov: `activityMatch.exec(pmDumpOutput)` will return `null` if nothing matches the RegExp. I think it will be better to check the result before we try to get the first match from it.

TsvetanMilanov: Also I think if the result is `null` we should try to start the default NativeScript/AppBuilder activity. This way if the output from `pm dump` is changed we will be able to start applications which use the default activity.

mobile/android/android-application-manager.ts
```diff
@@ -44,10 +44,16 @@ export class AndroidApplicationManager extends ApplicationManagerBase {
...
+
+		if (possibleIdentifier) {
+			await this.adb.executeShellCommand(["am", "start", "-n", possibleIdentifier]);
+		} else {
+			this.$logger.trace(`Tried starting activity: ${possibleIdentifier}, but failed`);
```
TsvetanMilanov: We will get here if we don't have `possibleIdentifier`. This means that we will log `null/undefined/""`. Maybe we should log the `appIdentifier`.

mobile/android/android-application-manager.ts
rosen-vladimirov: Also this regex is really complicated. Is it from official documentation? In any case we need tests for it :)

Plamen5kov: One regex is for verified Android application package and the second part is for a fully qualified java class name. Will write the tests!

Plamen5kov: Here's an example of the `pm dump | grep -A 1 MAIN`:
```
       android.intent.action.MAIN:
          3b2df03 org.nativescript.cliapp/com.tns.NativeScriptActivity filter 50dd82e
            Action: "android.intent.action.MAIN"
            Category: "android.intent.category.LAUNCHER"
--
        intent={act=android.intent.action.MAIN cat=[android.intent.category.LAUNCHER] flg=0x10200000 cmp=org.nativescript.cliapp/com.tns.NativeScriptActivity}
        realActivity=org.nativescript.cliapp/com.tns.NativeScriptActivity
--
            Intent { act=android.intent.action.MAIN cat=[android.intent.category.LAUNCHER] flg=0x10200000 cmp=org.nativescript.cliapp/com.tns.NativeScriptActivity }
            frontOfTask=true task=TaskRecord{fe592ac #449 A=org.nativescript.cliapp U=0 StackId=1 sz=1}
```
and the regex finds the full path to the activity name, which in this case is: `org.nativescript.cliapp/com.tns.NativeScriptActivity`

mobile/android/android-application-manager.ts
Plamen5kov: I prefer we don't start Activities that are not classified with the "MAIN" tag. The activity may be in the manifest, but not as a "MAIN" activity, and it will be an unexpected behavior to start an activity not meant as main.

Plamen5kov: will do!

mobile/android/android-application-manager.ts
Plamen5kov: good catch will fix to show app identifier

mobile/android/android-application-manager.ts
TsvetanMilanov: Yes, it will be unexpected behavior. My point is that if we start the default main activity, after we have failed to get it from pm dump, we will not break the CLI for the users who use the default main activity.
When we use the monkey tool we know that it will start the main activity, no matter how it's named. With the new logic we search for the main activity name with `grep -A 1` and I think it is possible that on different Android versions the output can be different. Also if we hit other problems with the new approach we will not be able to start any application if we don't use the default main activity.

Plamen5kov: I understand your point, but we can't start a hardcoded activity and hope it's the main one in the manifest.

Plamen5kov:
The method you're referring to `runApplicationOnEmulatorCore` in the `android-emulator-services` is not used since we removed the `emulate` command. We can remove the whole emulator service, but it let it be in another PR.

TsvetanMilanov:
@Plamen5kov Can you check if `androidEmulatorServices` is used in the AppBuilder CLI before you remove it?

test/unit-tests/android-application-manager.ts
```diff
@@ -0,0 +1,63 @@
...
+
+			for (let i = 0; i < validTestInput.length; i += 1) {
+				let validInput = validTestInput[i];
+				const match = activityMatch.exec(validInput);
+				let expectedElement = expectedValidTestInput[i];
+
+				assert.isArray(match);
+				assert.isTrue(expectedElement === match[0]);
+			}
+			assert.isTrue(true);
```
Plamen5kov: must have forgotten it when testing, will remove it

test/unit-tests/android-application-manager.ts
```diff
@@ -0,0 +1,63 @@
+import { AndroidApplicationManager } from "../../mobile/android/android-application-manager";
+import { Yok } from "../../yok";
+import { assert } from "chai";
+
+describe("android-application-manager", () => {
+
+	let testInjector: IInjector,
+		validTestInput: Array<string>,
+		expectedValidTestInput: Array<string>;
```
Plamen5kov: why does it matter if they are string[] or Array<string>?

mobile/android/android-application-manager.ts
```diff
@@ -44,10 +44,22 @@ export class AndroidApplicationManager extends ApplicationManagerBase {
 	}
 
 	public async startApplication(appIdentifier: string): Promise<void> {
-		await this.adb.executeShellCommand(["monkey",
-			"-p", appIdentifier,
-			"-c", "android.intent.category.LAUNCHER",
-			"1"]);
+		const pmDumpOutput = await this.adb.executeShellCommand(["pm", "dump", appIdentifier, "|", "grep", "-A", "1", "MAIN"]);
+		const fullActivityNameRegExp = this.getFullyQualifiedActivityRegex();
+		const activityMatch = new RegExp(fullActivityNameRegExp, "m");
```
Plamen5kov: I do need the "m" flag because we are searching for "MAIN", but the actual string we're matching is on the next line.

Plamen5kov:
Sure, when I make another PR about removing the androidEmulatorService, I'll check.

mobile/android/android-application-manager.ts
```diff
@@ -102,4 +114,8 @@ export class AndroidApplicationManager extends ApplicationManagerBase {
 
 		return applicationViews;
 	}
+
+	public getFullyQualifiedActivityRegex(): RegExp {
+		return /([A-Za-z]{1}[A-Za-z\d_]*\.)*[A-Za-z][A-Za-z\d_]*\/([a-z][a-z_0-9]*\.)*[A-Z_$]($[A-Z_$]|[$_\w_])*/;
```
Plamen5kov: I didn't want to split the regex because of the escaping, but if it's done like so in other places, I can also do it here.

test/unit-tests/android-application-manager.ts
TsvetanMilanov: There is no difference between `Array<string>` and `string[]`. I prefer to use `string[]` for consistency with the other code.

mobile/android/android-application-manager.ts
```diff
@@ -44,10 +44,35 @@ export class AndroidApplicationManager extends ApplicationManagerBase {
...
+			Action: "android.intent.action.MAIN"
+			Category: "android.intent.category.LAUNCHER"
+			--
+			intent={act=android.intent.action.MAIN cat=[android.intent.category.LAUNCHER] flg=0x10200000 cmp=org.nativescript.cliapp/com.tns.NativeScriptActivity}
+			realActivity=org.nativescript.cliapp/com.tns.NativeScriptActivity
+			--
+			Intent { act=android.intent.action.MAIN cat=[android.intent.category.LAUNCHER] flg=0x10200000 cmp=org.nativescript.cliapp/com.tns.NativeScriptActivity }
+			frontOfTask=true task=TaskRecord{fe592ac #449 A=org.nativescript.cliapp U=0 StackId=1 sz=1}
+		*/
+		const pmDumpOutput = await this.adb.executeShellCommand(["pm", "dump", appIdentifier, "|", "grep", "-A", "1", "MAIN"]);
```
rosen-vladimirov: [Minor] Do we really need this `grep`? First of all, is `grep -A` available on all Android devices? Also we are getting part of the output of `pm dump` here and later we are searching for something via regular expression. I'm wondering, is it possible to make the regular expression smart enough to parse the full result.

mobile/android/android-application-manager.ts
```diff
@@ -44,10 +44,35 @@ export class AndroidApplicationManager extends ApplicationManagerBase {
...
+		const activityMatch = this.getFullyQualifiedActivityRegex();
+		const match = activityMatch.exec(pmDumpOutput);
+		let possibleIdentifier = "";
```
rosen-vladimirov: this can be:
```TypeScript
const possibleIdentifier = match && match[0];
```

test/unit-tests/android-application-manager.ts
```diff
@@ -0,0 +1,129 @@
...
+				}
+			} else {
+				this.startedWithActivityManager = this.checkIfStartedWithActivityManager(args);
+				if (this.startedWithActivityManager) {
+					this.validIdentifierPassed = this.checkIfValidIdentifierPassed(args);
+				}
+			}
+		}
+		AndroidDebugBridgeStub.methodCallCount++;
+		Promise.resolve();
```
rosen-vladimirov: you do not need this

test/unit-tests/android-application-manager.ts
```diff
@@ -0,0 +1,129 @@
...
+			case "monkey": return false;
+			default: return false;
+		}
+	}
+
+	private checkIfValidIdentifierPassed(args: string[]): Boolean {
+		if (args && args.length) {
+			const possibleIdentifier = args[args.length - 1];
+			let validTestString = this.expectedValidTestInput[AndroidDebugBridgeStub.methodCallCount];
+			if (possibleIdentifier === validTestString) {
```
rosen-vladimirov:
```TypeScript
return possibleIdentifier === validTestString;
```

test/unit-tests/android-application-manager.ts
```diff
@@ -0,0 +1,129 @@
...
+			case "am": return true;
+			case "monkey": return false;
+			default: return false;
+		}
+	}
+
+	private checkIfValidIdentifierPassed(args: string[]): Boolean {
+		if (args && args.length) {
+			const possibleIdentifier = args[args.length - 1];
+			let validTestString = this.expectedValidTestInput[AndroidDebugBridgeStub.methodCallCount];
```
rosen-vladimirov: const

test/unit-tests/android-application-manager.ts
```diff
@@ -0,0 +1,129 @@
...
+		androidDebugBridge:AndroidDebugBridgeStub;
+
+	beforeEach(() => {
+		testInjector = createTestInjector();
+		androidApplicationManager = testInjector.resolve("androidApplicationManager");
+		androidDebugBridge = testInjector.resolve("adb");
+	});
+	describe("startApplication", () => {
+		it("fires up the right application", async () => {
+			for (let i = 0; i < androidDebugBridge.getInputLength(); i += 1) {
```
rosen-vladimirov: `i +=1` -> `i++`

mobile/android/android-application-manager.ts
```diff
@@ -102,4 +127,12 @@ export class AndroidApplicationManager extends ApplicationManagerBase {
 
 		return applicationViews;
 	}
+
+	private getFullyQualifiedActivityRegex(): RegExp {
+		const androidPackageName = "([A-Za-z]{1}[A-Za-z\\d_]*\\.)*[A-Za-z][A-Za-z\\d_]*";
+		const packageActivitySeparator = "\\/";
+		const fullJavaClassName = "([a-z][a-z_0-9]*\\.)*[A-Z_$]($[A-Z_$]|[$_\\w_])*";
+
+		return new RegExp(`${androidPackageName}${packageActivitySeparator}${fullJavaClassName}`, `m`);
```
Plamen5kov: Will put `@cache` as you suggested, thank you!
I thought about using the app identifier for the first part of the regex, but I prefer to keep the general implementation since the app identifier is sometimes overridden by the user in `app.gradle`.

test/unit-tests/android-application-manager.ts
Plamen5kov: `i += 1` is faster, but if you insist I have no problem.

rosen-vladimirov: Well, that's not entirely correct. Consider the following example.
```JavaScript
const iterations = 100*1000*1000;

console.time("i+=1");
for(let i = 0; i < iterations; i+=1) {
}
console.timeEnd("i+=1");

console.time("i++");
for(let i = 0; i < iterations; i++) {
}
console.timeEnd("i++");
```

And the result is:
```
$ node b.js
i+=1: 1118.667ms
i++: 887.539ms

$ node b.js
i+=1: 1116.609ms
i++: 877.461ms

$ node b.js
i+=1: 1106.636ms
i++: 882.537ms
```

But anyway, the codebase contains many lines with `<variable name>++` , so I prefer consistency here.

mobile/android/android-application-manager.ts
Plamen5kov: `So what's the expected behavior when the application has more than one MAIN activity`
The first main activity will be taken. I've not seen a user case where there's more than one MAIN activity, plus the user can specify a DEFAULT main activity, but I haven't seen that either, so I didn't want to complicate the implementation if there's no need.

`Also, shouldn't we find the activity with Launcher category?`
I also thought about this, but then I found out that some part of the usual Android applications don't specify a launcher category. Read more [here](https://chris.orr.me.uk/android-apps-info-category/). Keeping this in mind I thought best not to discriminate against such apps.
Regarding the example of the two MAIN activities you provided, I haven't thought about it, because up until now, I hadn't seen this scenario, but I'm open to discussing all possible scenarios.

`Isn't it faster to call adb shell dumpsys package <appIdentifier>`
To be honest I used `pm dump` instead of `dumpsys` because when I saw the content of `pm` it seemed more reliable because it was part of the `framework.jar`. `dumpsys` command is just a binary that is somewhere in the `/system/bin` folder, and I couldn't find any information on its support and reliability.
```
generic_x86:/ # cat /system/bin/pm                                                                                                                                   
# Script to start "pm" on the device, which has a very rudimentary
# shell.
#
base=/system
export CLASSPATH=$base/framework/pm.jar
exec app_process $base/bin com.android.commands.pm.Pm "$@"

generic_x86:/ # cat /system/bin/am                                                                                                                                   
#!/system/bin/sh
#
# Script to start "am" on the device, which has a very rudimentary
# shell.
#
```
`Do we really need this grep`
I've tested this and the regex will work with the full result of the `pm dump` in case `-A 1` isn't available, but the regex works faster in a shorter amount of text so I left it with the `-A 1` option.

test/unit-tests/android-application-manager.ts
Plamen5kov: Ha, cool they fixed it. Will change it.

mobile/android/android-application-manager.ts
rosen-vladimirov: @Plamen5kov , thanks for the detailed explanations. I'm not familiar with Android apps with more than one MAIN activities, there are such samples in Android Studio. It looks like the only way to understand if the current approach is fine is to merge it and see if such case will ever arise.

test/unit-tests/android-application-manager.ts
```diff
@@ -0,0 +1,127 @@
...
+			Intent { act=android.intent.action.MAIN cat=[android.intent.category.LAUNCHER] flg=0x10200000 cmp=org.nativescript.testApp/com.tns.NativeScriptActivity } \
+			frontOfTask=true task=TaskRecord{fe592ac #449 A=org.nativescript.testApp U=0 StackId=1 sz=1}"
+		];
+
+	public async executeShellCommand(args: string[]): Promise<any> {
+		if (args && args.length > 0) {
+			if (args[0] === "pm") {
+				const passedIdentifier = args[2];
+				if (passedIdentifier === invalidIdentifier) {
+					return Promise.resolve("invalid output string");
```
rosen-vladimirov: No need for `Promise.resolve` here - just `return "invalid output string";`

test/unit-tests/android-application-manager.ts
```diff
@@ -0,0 +1,127 @@
...
+				} else {
+					const testString = this.validTestInput[AndroidDebugBridgeStub.methodCallCount];
+					return Promise.resolve(testString);
```
rosen-vladimirov: No need for `Promise.resolve` here - just `return testString;`

Plamen5kov:
ping @TsvetanMilanov

PR #996, opened 2017-07-31
fix getxcodeversion usage

Plamen5kov:
_problem_
related issue: https://github.com/NativeScript/nativescript-cli/issues/2810
`getxcodeversion` returning null isn't handled correctly in `tns doctor`

_solution_
handle erroneous case and remove unnecessary code

### DIFF
--- a/services/xcode-select-service.ts
+++ b/services/xcode-select-service.ts
@@ -30,8 +30,11 @@ export class XcodeSelectService implements IXcodeSelectService {
 	@cache()
 	public async getXcodeVersion(): Promise<IVersionData> {
 		let sysInfoBase = this.$injector.resolve("sysInfoBase");
-		let xcodeVer = await sysInfoBase.getXCodeVersion(),
-			xcodeVersionMatch = xcodeVer.match(/Xcode (.*)/),
+		let xcodeVer = await sysInfoBase.getXCodeVersion();
+		if (!xcodeVer) {
+			this.$errors.fail("xcodebuild execution failed. Make sure that you have latest Xcode and tools installed.");
+		}
+		let xcodeVersionMatch = xcodeVer.match(/Xcode (.*)/),
 			xcodeVersionGroup = xcodeVersionMatch && xcodeVersionMatch[1],
 			xcodeVersionSplit = xcodeVersionGroup && xcodeVersionGroup.split(".");
 
--- a/services/xcode-select-service.ts
+++ b/services/xcode-select-service.ts
@@ -32,7 +32,7 @@ export class XcodeSelectService implements IXcodeSelectService {
 		let sysInfoBase = this.$injector.resolve("sysInfoBase");
 		let xcodeVer = await sysInfoBase.getXCodeVersion();
 		if (!xcodeVer) {
-			this.$errors.fail("xcodebuild execution failed. Make sure that you have latest Xcode and tools installed.");
+			this.$errors.failWithoutHelp("xcodebuild execution failed. Make sure that you have latest Xcode and tools installed.");
 		}
 
 		let xcodeVersionMatch = xcodeVer.match(/Xcode (.*)/),
--- a/services/xcode-select-service.ts
+++ b/services/xcode-select-service.ts
@@ -34,6 +34,7 @@ export class XcodeSelectService implements IXcodeSelectService {
 		if (!xcodeVer) {
 			this.$errors.fail("xcodebuild execution failed. Make sure that you have latest Xcode and tools installed.");
 		}
+
 		let xcodeVersionMatch = xcodeVer.match(/Xcode (.*)/),
 			xcodeVersionGroup = xcodeVersionMatch && xcodeVersionMatch[1],
 			xcodeVersionSplit = xcodeVersionGroup && xcodeVersionGroup.split(".");
--- a/sys-info-base.ts
+++ b/sys-info-base.ts
@@ -57,18 +57,8 @@ export class SysInfoBase implements ISysInfo {
 		return this.javaCompilerVerCache;
 	}
 
-	private xCodeVerCache: string = null;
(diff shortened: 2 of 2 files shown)


services/xcode-select-service.ts
```diff
@@ -30,8 +30,11 @@ export class XcodeSelectService implements IXcodeSelectService {
 	@cache()
 	public async getXcodeVersion(): Promise<IVersionData> {
 		let sysInfoBase = this.$injector.resolve("sysInfoBase");
-		let xcodeVer = await sysInfoBase.getXCodeVersion(),
-			xcodeVersionMatch = xcodeVer.match(/Xcode (.*)/),
+		let xcodeVer = await sysInfoBase.getXCodeVersion();
+		if (!xcodeVer) {
+			this.$errors.fail("xcodebuild execution failed. Make sure that you have latest Xcode and tools installed.");
+		}
```
TsvetanMilanov: Can you add one new line after this `}`

sys-info-base.ts
```diff
@@ -57,18 +57,8 @@ export class SysInfoBase implements ISysInfo {
...
-			try {
-				this.xCodeVerCache = this.$hostInfo.isDarwin ? await this.exec("xcodebuild -version") : null;
-			} catch (e) {
-				this.$logger.trace(`Command "xcodebuild -version" failed: ${e}`);
-				this.xCodeVerCache = null;
-			}
-		}
-
-		return this.xCodeVerCache;
+		return this.$hostInfo.isDarwin ? await this.exec("xcodebuild -version") : null;
```
TsvetanMilanov: Why we don't want to have cache anymore?

sys-info-base.ts
```diff
@@ -157,9 +147,9 @@ export class SysInfoBase implements ISysInfo {
 			res.procArch = process.arch;
 			res.nodeVer = process.version;
 
-			res.npmVer = await this.getNpmVersion();
+			res.npmVer = await this.getNpmVersion(); //not used anywhere except for tests
```
TsvetanMilanov: Can you check if we use this in the AppBuilder CLI

sys-info-base.ts
```diff
@@ -157,9 +147,9 @@ export class SysInfoBase implements ISysInfo {
...
 
-			res.javaVer = await this.getJavaVersion();
+			res.javaVer = await this.getJavaVersion(); //not used anywhere except for tests
```
TsvetanMilanov: Can you check if we use this in the AppBuilder CLI

sys-info-base.ts
```diff
@@ -207,6 +197,7 @@ export class SysInfoBase implements ISysInfo {
 			}
 		} catch (e) {
 			// if we got an error, assume not working
+			this.$logger.trace(`Error while executing child process: ${e}`);
```
TsvetanMilanov: You can change `child process` to `${cmd}`. Also I think if we log `${e}` here, in the logs we will see `[object object]` instead of the error. You can log `e.message`.

sys-info-base.ts
Plamen5kov: i'll return the cache

sys-info-base.ts
Plamen5kov: will do

sys-info-base.ts
Plamen5kov: i got no private access to look at the repo, i'll remove the comments

sys-info-base.ts
Plamen5kov: i got no private access to look at the repo, i'll remove the comments

services/xcode-select-service.ts
```diff
@@ -30,8 +30,12 @@ export class XcodeSelectService implements IXcodeSelectService {
 	@cache()
 	public async getXcodeVersion(): Promise<IVersionData> {
 		let sysInfoBase = this.$injector.resolve("sysInfoBase");
-		let xcodeVer = await sysInfoBase.getXCodeVersion(),
-			xcodeVersionMatch = xcodeVer.match(/Xcode (.*)/),
+		let xcodeVer = await sysInfoBase.getXCodeVersion();
+		if (!xcodeVer) {
+			this.$errors.fail("xcodebuild execution failed. Make sure that you have latest Xcode and tools installed.");
```
TsvetanMilanov: Can we use `this.$errors.failWithoutHelp` here or we need to use `fail`?

Plamen5kov: I originally made it to failWithoutHelp, but for consistency with https://github.com/NativeScript/nativescript-cli/blob/1955028ac3ded6a012813476f08d0f4f0d89dddd/lib/services/ios-project-service.ts#L1190 i left it like this. If you insist I have no objection of making it fail without help.

TsvetanMilanov: I think it is better to use `this.$errors.failWithoutHelp` in the services.

rosen-vladimirov:
This code will not fix the mentioned issue, it will just fail with different error when executing `tns doctor`. 
By definition `tns doctor` should detect all issues and print them, it should not fail and stop execution once a single issue is detected.

@Plamen5kov am I missing something here, maybe there's another PR in {N} CLI that would fix the `doctor` failure?

Plamen5kov:
@rosen-vladimirov I thought failing was the expected behavior, my bad. I'll fix it in another PR.

rosen-vladimirov:
@Plamen5kov the failure in the current code (changed in this PR) is fine IMO. But `tns doctor` should catch this error and show warning or something like this. Or even - `tns doctor` should not get here in case we already know Xcode is not installed.

Plamen5kov:
Yes, my plan was to add a check to `tns doctor` on the expected default tools on mac, just like we do [here](https://github.com/NativeScript/nativescript-cli/blob/master/lib/services/doctor-service.ts#L61). Will fix as soon as I get the time

PR #997, opened 2017-08-09
Fix the error handling in hooks execution

TsvetanMilanov:
Some hooks can reject their promises without error.
We need to check if we have an error before we try to get some information from it.

### DIFF
--- a/services/hooks-service.ts
+++ b/services/hooks-service.ts
@@ -132,7 +132,7 @@ export class HooksService implements IHooksService {
 						await maybePromise;
 					} catch (err) {
 						if (err && _.isBoolean(err.stopExecution) && err.errorAsWarning === true) {
-							this.$logger.warn(err.message);
+							this.$logger.warn(err.message || err);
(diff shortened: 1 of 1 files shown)


services/hooks-service.ts
```diff
@@ -128,28 +128,25 @@ export class HooksService implements IHooksService {
...
-				this.$logger.trace("Executing %s hook at location %s with environment ", hookName, hook.fullPath, environment);
-
-				let output = await this.$childProcess.spawnFromEvent(command, [hook.fullPath], "close", environment, { throwError: false });
-				if (output.exitCode !== 0) {
-					throw new Error(output.stdout + output.stderr);
+					try {
+						await maybePromise;
+					} catch (err) {
+						if (err && _.isBoolean(err.stopExecution) && err.errorAsWarning === true) {
+							this.$logger.warn(err.message);
```
rosen-vladimirov: this should be `err.message || err`

PR #999, opened 2017-08-13
Add support for `<span>` tag in CLI's help

rosen-vladimirov:
Placing `<span>` tags in CLI's markdown files will work when generating HTML help pages, but the tags will also be shown in terminal. Skip the tags when generating command-line help.


rosen-vladimirov:
Note: Tests are failing as there's no `.travis.yml` in this branch :)

PR #1007, opened 2017-08-24
Allow setting eqatec-analytics service settings from child classes

rosen-vladimirov:
Currently the `AnalyticsServiceBase` is starting the eqatec monitor process and the only setting that callers may pass is the API key.
Expose a new protected method that allows starting the monitor with all settings specified. Keep the old private method `start` and when it is called get all required settings from user-settings-service and analytics-settings-service.
Remove code for restarting Analytics monitor when exception should be tracked. We'll no longer use it in {N} CLI, where we had such requirement.


services/analytics-service-base.ts
```diff
@@ -252,56 +310,16 @@ export class AnalyticsServiceBase implements IAnalyticsService {
...
 
 		return `${readableSettingName} is ${status}.`;
 	}
 
 	private async getJsonStatusMessage(settingName: string): Promise<string> {
-		let status = await this.getStatus(settingName);
-		let enabled = status === AnalyticsStatus.notConfirmed ? null : status === AnalyticsStatus.disabled ? false : true;
-		return JSON.stringify({ enabled: enabled });
+		const status = await this.getStatus(settingName);
+		const enabled = status === AnalyticsStatus.notConfirmed ? null : status === AnalyticsStatus.disabled ? false : true;
```
Mitko-Kerezov: I think this can be written a bit more readable

commands/analytics.ts
```diff
@@ -63,16 +63,17 @@ export class ErrorReportingCommand extends AnalyticsCommand {
...
 		$options: ICommonOptions,
-		private $staticConfig: Config.IStaticConfig) {
+		$staticConfig: Config.IStaticConfig
+	) {
 		super($analyticsService, $logger, $errors, $options, $staticConfig.ERROR_REPORT_SETTING_NAME, "Error reporting");
 	}
 
 	public async execute(args: string[]): Promise<void> {
-		const shouldRestartEqatecMonitor = this.$staticConfig.ANALYTICS_EXCEPTIONS_API_KEY && await this.$analyticsService.isEnabled(this.$staticConfig.TRACK_FEATURE_USAGE_SETTING_NAME);
+		// const shouldRestartEqatecMonitor = this.$staticConfig.ANALYTICS_EXCEPTIONS_API_KEY && await this.$analyticsService.isEnabled(this.$staticConfig.TRACK_FEATURE_USAGE_SETTING_NAME);
```
Mitko-Kerezov: In case you're only going to call the `super` in this method you'd better not override it at all

declarations.d.ts
```diff
@@ -467,7 +467,8 @@ declare const enum ErrorCodes {
 	INVALID_ARGUMENT = 128,
 	RESOURCE_PROBLEM = 129,
 	KARMA_FAIL = 130,
-	UNHANDLED_REJECTION_FAILURE = 131
+	UNHANDLED_REJECTION_FAILURE = 131,
+	FAILED_EQATEC_ANALYTICS_PROCESS = 132
```
rosen-vladimirov: this is no longer required and used, it should be deleted.

declarations.d.ts
```diff
@@ -558,14 +559,14 @@ interface IAnalyticsService {
...
 	/**
-	 * Restarts the monitor with a new API key.
-	 * @param {string} projectApiKey The API key for new Analytics project.
-	 * @returns {Promise<void>}
+	 * Tracks the answer of question if user allows to be tracked.
+	 * @param {{ acceptTrackFeatureUsage: boolean }} settings Object containing information about user's answer.
+	 * @return {Promise<void>}
 	 */
-	restartEqatecMonitor(projectApiKey: string): Promise<void>;
+	trackAcceptFeatureUsage(settings: { acceptTrackFeatureUsage: boolean }): Promise<void>;
```
Mitko-Kerezov: You could extract an interface here and reuse it in the {N} CLI PR

services/analytics-service-base.ts
```diff
@@ -56,6 +62,15 @@ export class AnalyticsServiceBase implements IAnalyticsService {
 		}
 	}
 
+	public async trackAcceptFeatureUsage(settings: { acceptTrackFeatureUsage: boolean }): Promise<void> {
+		try {
+			await this.sendDataToEqatecMonitors(this.acceptUsageReportingAPIKeys,
+				(eqatecMonitor: IEqatecMonitor) => eqatecMonitor.trackFeature(`${this.acceptTrackFeatureSetting}.${settings.acceptTrackFeatureUsage}`));
+		} catch (e) {
+			this.$logger.trace(`Analytics exception: '${e}'`);
```
Mitko-Kerezov: In case an exception is raised here we'd trace `Analytics exception: '[object Object]'`

services/analytics-service-base.ts
```diff
@@ -78,26 +93,11 @@ export class AnalyticsServiceBase implements IAnalyticsService {
...
-					// Sending the exception might take a while.
-					// As in most cases we exit immediately after exception is caught,
-					// wait for tracking the exception.
-					await this.$progressIndicator.showProgressIndicator(this.waitForSending(), 500);
-				}
+				await this.sendDataToEqatecMonitors(this.exceptionsTrackingAPIKeys,
+					(eqatecMonitor: IEqatecMonitor) => eqatecMonitor.trackException(exception, message));
 			} catch (e) {
-				this.$logger.trace("Analytics exception: '%s'", e.toString());
+				this.$logger.trace(`Analytics exception: '${e}'`);
```
Mitko-Kerezov: Same here - we'd better trace the `e` object as a separate argument

services/analytics-service-base.ts
```diff
@@ -134,29 +135,92 @@ export class AnalyticsServiceBase implements IAnalyticsService {
...
-				await this.start();
-				if (this._eqatecMonitor) {
-					this._eqatecMonitor.trackFeature(featureTrackString);
-					await this.waitForSending();
-				}
+				await this.sendDataToEqatecMonitors(this.featureTrackingAPIKeys, (eqatecMonitor: IEqatecMonitor) => eqatecMonitor.trackFeature(featureTrackString));
 			}
 		} catch (e) {
-			this.$logger.trace("Analytics exception: '%s'", e.toString());
+			this.$logger.trace(`Analytics exception: '${e}'`);
```
Mitko-Kerezov: Same here

services/analytics-service-base.ts
```diff
@@ -134,29 +135,92 @@ export class AnalyticsServiceBase implements IAnalyticsService {
...
+			if (this.getIsSending(eqatecMonitor)) {
+				this.$logger.trace(`Waiting for analytics to send information. Will check in ${intervalTime}ms.`);
+				const interval = setInterval(() => {
+					if (!this.getIsSending(eqatecMonitor) || remainingTime <= 0) {
+						clearInterval(interval);
+						resolve();
+					}
+
+					remainingTime -= intervalTime;
+					this.$logger.trace(`Waiting for analytics to send information. Will check in ${intervalTime}ms. Remaining time is: ${remainingTime}`);
```
Mitko-Kerezov: The first part of this message is duplicated in a `trace` above - can be extracted as a `const` variable

services/analytics-service-base.ts
```diff
@@ -134,29 +135,92 @@ export class AnalyticsServiceBase implements IAnalyticsService {
...
+			for (const settingName of settingsNames) {
+				await this.getStatus(settingName);
+			}
+
+			this.$logger.trace("Analytics statuses: ");
+			this.$logger.trace(this.analyticsStatuses);
+		}
+	}
+
+	private getIsSending(eqatecMonitor: any): boolean {
```
Mitko-Kerezov: Isn't the type here `IEqatecMonitor` instead of  `any`

services/analytics-service-base.ts
```diff
@@ -134,29 +135,92 @@ export class AnalyticsServiceBase implements IAnalyticsService {
...
+		return eqatecMonitor.status().isSending;
+	}
+
+	private waitForSending(eqatecMonitor: any): Promise<void> {
```
Mitko-Kerezov: Same for here IMO you could mark this as `eqatecMonitor: IEqatecMonitor`

PR #1009, opened 2017-08-25
Add prefer-const rule to tslint

rosen-vladimirov:
Enable `prefer-const` rule in tslint and fix the errors:
The rule requires that variable declarations use const instead of let and var if possible.

If a variable is only assigned to once when it is declared, it should be declared using `const`

The prefer-const rule checks the code and verifies the usage of let and const

### DIFF
--- a/commands/proxy/proxy-set.ts
+++ b/commands/proxy/proxy-set.ts
@@ -27,9 +27,9 @@ export class ProxySetCommand extends ProxyCommandBase {
 	}
 
 	public async execute(args: string[]): Promise<void> {
-		let urlString = args[0],
-			username = args[1],
-			password = args[2];
+		let urlString = args[0];
+		let username = args[1];
+		let password = args[2];
 
 		const noUrl = !urlString;
 		if (noUrl) {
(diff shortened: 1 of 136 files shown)


commands/proxy/proxy-set.ts
```diff
@@ -27,9 +27,9 @@ export class ProxySetCommand extends ProxyCommandBase {
 	}
 
 	public async execute(args: string[]): Promise<void> {
-		let urlString = args[0],
-			username = args[1],
-			password = args[2];
+		let urlString = args[0];
```
galexandrov: const?

helpers.ts
```diff
@@ -1,6 +1,6 @@
 import * as uuid from "uuid";
 import * as net from "net";
-let Table = require("cli-table");
+const Table = require("cli-table");
```
galexandrov: Can we move the const after the imports, seems more right :)

commands/proxy/proxy-set.ts
rosen-vladimirov: nope, it is set on line 39
Also now linter checks if any `const` can be used instead of `let` and shows error in this case.

PR #1011, opened 2017-08-31
detach logcat helper on stop application + tests

Plamen5kov:
_problem_
`logcatHelper` has inconsistent behavior between different API level emulators (logs missing on some emulators, depending on how fast the application starts)

_solution_
* Detach each time the application is stopped (no reason to keep streaming the logcat if the application is stopped)
* Start the `logcatHelper` each time the application is started
before we needed to do an `adb logcat -c` before attaching to the logcat, because the filtering by PID was no good, but that's no longer a problem.

### DIFF
--- a/mobile/android/logcat-helper.ts
+++ b/mobile/android/logcat-helper.ts
@@ -3,8 +3,8 @@ import { DeviceAndroidDebugBridge } from "./device-android-debug-bridge";
 import { ChildProcess } from "child_process";
 
 interface IDeviceLoggingData {
-    loggingProcess: ChildProcess;
-    lineStream: any;
+	loggingProcess: ChildProcess;
+	lineStream: any;
 }
 
 export class LogcatHelper implements Mobile.ILogcatHelper {
@@ -27,7 +27,7 @@ export class LogcatHelper implements Mobile.ILogcatHelper {
 			this.mapDevicesLoggingData[deviceIdentifier] = {
 				loggingProcess: logcatStream,
 				lineStream: lineStream
-			}
+			};
 
 			logcatStream.stderr.on("data", (data: NodeBuffer) => {
 				this.$logger.trace("ADB logcat stderr: " + data.toString());
@@ -54,7 +54,7 @@ export class LogcatHelper implements Mobile.ILogcatHelper {
 	}
 
 	public stop(deviceIdentifier: string): void {
-		if(this.mapDevicesLoggingData[deviceIdentifier]) {
+		if (this.mapDevicesLoggingData[deviceIdentifier]) {
 			this.mapDevicesLoggingData[deviceIdentifier].loggingProcess.removeAllListeners();
 			this.mapDevicesLoggingData[deviceIdentifier].lineStream.removeAllListeners();
 		}
--- a/mobile/android/logcat-helper.ts
+++ b/mobile/android/logcat-helper.ts
@@ -3,8 +3,8 @@ import { DeviceAndroidDebugBridge } from "./device-android-debug-bridge";
 
 export class LogcatHelper implements Mobile.ILogcatHelper {
 	private mapDeviceToLoggingStarted: IDictionary<boolean>;
-	private adbLogcat:any;
-	private lineStream:any;
+	private mapDeviceToLogcat:any;
+	private mapDeviceToLineStream:any;
 
 	constructor(private $deviceLogProvider: Mobile.IDeviceLogProvider,
 		private $devicePlatformsConstants: Mobile.IDevicePlatformsConstants,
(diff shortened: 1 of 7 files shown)


mobile/android/logcat-helper.ts
```diff
@@ -16,17 +18,14 @@ export class LogcatHelper implements Mobile.ILogcatHelper {
 		if (deviceIdentifier && !this.mapDeviceToLoggingStarted[deviceIdentifier]) {
 			const adb: Mobile.IDeviceAndroidDebugBridge = this.$injector.resolve(DeviceAndroidDebugBridge, { identifier: deviceIdentifier });
 
-			// remove cached logs:
-			await adb.executeCommand(["logcat", "-c"]);
+			this.adbLogcat = await adb.executeCommand(["logcat"], { returnChildProcess: true });
```
rosen-vladimirov: In case you have more than one device, `$logcatHelper` will be called for each of them and you'll cache the `adb logcat` instance of the last device. So after that you'll kill it only, restart only this instance, etc. Is this expected?

mobile/android/logcat-helper.ts
```diff
@@ -16,17 +18,14 @@ export class LogcatHelper implements Mobile.ILogcatHelper {
...
+			this.lineStream = byline(this.adbLogcat.stdout);
```
rosen-vladimirov: Same remark as the one for `adbLogcat` instance

test/unit-tests/mobile/android/logcat-helper.ts
```diff
@@ -0,0 +1,125 @@
...
+
+class ChildProcessStub {
+	public static methodCallCount = 0;
+	private isWin = /^win/.test(process.platform);
+
+	public spawn(command: string, args?: string[], options?: any): childProcess.ChildProcess {
+		ChildProcessStub.methodCallCount++;
+		let pathToExecutable = "";
+		if (this.isWin) {
+			pathToExecutable = "getcontent.bat";
```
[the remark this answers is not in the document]
Plamen5kov: As far as I tested it out there is no common command to use between unix machines and windows to print out a file content to the stdout.
`type` is windows specific and `cat` is unix speciffic -> [read more](http://www.yolinux.com/TUTORIALS/unix_for_dos_users.html)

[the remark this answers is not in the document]
rosen-vladimirov: I fully agree with this, my point is that you do not need to use `getcontent` executable. Just set the `pathToExecutable` to `type` in case you are on Windows and `cat` otherwise.

mobile/android/logcat-helper.ts
Plamen5kov: didn't think about that, thank you. will fix

mobile/android/logcat-helper.ts
```diff
@@ -3,30 +3,31 @@ import { DeviceAndroidDebugBridge } from "./device-android-debug-bridge";
...
+	private mapDeviceToLogcat:any;
+	private mapDeviceToLineStream:any;
 
 	constructor(private $deviceLogProvider: Mobile.IDeviceLogProvider,
 		private $devicePlatformsConstants: Mobile.IDevicePlatformsConstants,
 		private $logger: ILogger,
 		private $injector: IInjector,
 		private $processService: IProcessService) {
 		this.mapDeviceToLoggingStarted = Object.create(null);
+		this.mapDeviceToLogcat = Object.create(null);
```
rosen-vladimirov: Instead of creating three different dictionaries with the same key (`deviceIdentifier`), you can create a single one with two properties:
```TypeScript
interface IDeviceLoggingData {
    loggingProcess: ChildProcess; // (or any in case you define the interface in .d.ts)
    lineStream: any;
}

private devicesLoggingData: IDicitionary<ILoggingData>;
```

This way in `start` method you can use:
```TypeScript
const deviceLoggingData = this.devicesLoggingData[deviceIdentifier];
if (deviceIdentifier && !deviceLoggingData) {
...
}
```

And in the `stop` method, after removing all listeners, just delete the entry:
```
delete this.devicesLoggingData[deviceIdentifier];
```

The benefits of this solution are:
1. No need of multiple dictionaries - A single data object containing all the information.
2. No need to set true/false values, just check if there's an entry in the dictionary.
3. The code in start method will be easier for reading as it will use a single object.

What do you think?

Plamen5kov: Basically, I want to keep it simple, so I implemented the simplest solution, but I have no preference.

rosen-vladimirov: You can create the interface in the same file and use it only here. This way it will not be visible outside of the file.

PR #1012, opened 2017-09-07
Add lockfile service and lock operation for writing user settings file

rosen-vladimirov:
Add new lockfile service (use lockfile npm package) and lock all operations for reading/writing user-settings.json.
These operations may be executed in different processes, so they must be synced between them.

### DIFF
--- a/services/lockfile.ts
+++ b/services/lockfile.ts
@@ -1,0 +1,57 @@
+import * as lockfile from "lockfile";
+import * as path from "path";
+
+export class LockFile implements ILockFile {
+
+	private get defaultLockFilePath(): string {
+		return path.join(this.$options.profileDir, "lockfile.lock");
+	}
+
+	private get defaultLockParams(): lockfile.Options {
+		// We'll retry 100 times and time between retries is 100ms, i.e. full wait in case we are unable to get lock will be 10 seconds.
+		// In case lock is older than 3 minutes, consider it stale and try to get a new lock.
+		const lockParams: lockfile.Options = {
+			retryWait: 100,
+			retries: 100,
+			stale: 180 * 1000,
+		};
+
+		return lockParams;
+	}
+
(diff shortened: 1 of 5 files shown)


services/lockfile.ts
```diff
@@ -0,0 +1,57 @@
+import * as lockfile from "lockfile";
+import * as path from "path";
+
+export class LockFile implements ILockFile {
+
+	private get defaultLockFilePath(): string {
```
Mitko-Kerezov: We could `@cache()` this

services/lockfile.ts
```diff
@@ -0,0 +1,57 @@
...
+		return path.join(this.$options.profileDir, "lockfile.lock");
+	}
+
+	private get defaultLockParams(): lockfile.Options {
```
Mitko-Kerezov: We could also `@cache()` this but I don't believe it necessary

services/user-settings-service.ts
```diff
@@ -22,44 +30,67 @@ export class UserSettingsServiceBase implements IUserSettingsService {
...
+		return this.executeActionWithLock(action);
 
-		delete this.userSettingsData[key];
-		await this.saveSettings();
 	}
 
-	public async saveSettings(data?: any): Promise<void> {
-		await this.loadUserSettingsFile();
-		this.userSettingsData = this.userSettingsData || {};
+	private async executeActionWithLock(action: () => Promise<any>): Promise<any> {
```
Mitko-Kerezov: I believe this could be generic:
`private async executeActionWithLock<T>(action: () => Promise<T>): Promise<T> {`
that way we'd be able to get a little bit more out of the static type checking

services/lockfile.ts
rosen-vladimirov: I agree it is not necessary here ;)

PR #1017, opened 2017-10-16
fix java version detect

intellild:
related issue:
https://github.com/NativeScript/nativescript-cli/issues/3148

### DIFF
--- a/sys-info-base.ts
+++ b/sys-info-base.ts
@@ -18,7 +18,7 @@ export class SysInfoBase implements ISysInfo {
 		if (!this.javaVerCache) {
 			try {
 				// different java has different format for `java -version` command
-				const output = (await this.$childProcess.spawnFromEvent("java", ["-version"], "exit")).stderr;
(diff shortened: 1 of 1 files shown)


rosen-vladimirov:
Hey @intellild ,
First of all, thank you for your efforts and please excuse us for the late reply.
You are correct that with Java 9, the version is printed on stdout. However, in JAVA 8, the `javac -version` is printed on the stderr. That's why the current code is reading the version from stderr.
Your fix will work for Java 9, but it will fail when Java 8 is used. That's why the unit tests are failing on this PR.
I've investigated the code and in fact we are not using the Java version detection - we care about Java Compiler version. So I've created a PR to remove the Java version detection and also fix the detection of Java Compiler version. You can take a look at the PR [here](https://github.com/telerik/mobile-cli-lib/pull/1037).
So I'm closing this PR in favor of the other one.

Thanks again for your contribution - it allowed us to easily determine the issue with Java version detection and find the appropriate fix.

PR #1021, opened 2017-11-02
Do not track help command when other command fails 

rosen-vladimirov:
In case any command fails, we execute help command in order to show the help content. This leads to multiple trackings, i.e. - user executes only one command, but in Analytics we see two commands.
Instead of executing help command, introduce new method in htmlHelpService, that prints the help to the terminal and call it instead. Use the same method in the help command itself.
Rename htmlHelpService to helpService - it has been incorrectly named from the beginning.
Remove `helpTextPath` from staticConfig interface - this property is not used for more than 2 years.
Introduce tests for `helpService` - get the tests from AppBuilder CLI.

### DIFF
--- a/services/help-service.ts
+++ b/services/help-service.ts
@@ -74,7 +74,12 @@ export class HelpService implements IHelpService {
 		this.$logger.printMarkdown(help);
 	}
 
-	public async getCommandLineHelpForCommand(commandName: string): Promise<string> {
+	/**
+	 * Gets the help content for a specific command that should be shown on the terminal.
+	 * @param {string} commandName Name of the command for which to read the help.
+	 * @returns {Promise<string>} Help content of the command parsed with all terminal rules applied (stripped content that should be shown only for html help).
+	 */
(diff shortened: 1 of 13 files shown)


KristianDD:
Looks ok.

services/help-service.ts
```diff
@@ -38,23 +38,59 @@ export class HtmlHelpService implements IHtmlHelpService {
...
+	public async showCommandLineHelp(commandName: string): Promise<void> {
+		const help = await this.getCommandLineHelpForCommand(commandName);
+		if (this.$staticConfig.FULL_CLIENT_NAME) {
+			this.$logger.info(this.$staticConfig.FULL_CLIENT_NAME.green.bold + EOL);
+		}
+
+		this.$logger.printMarkdown(help);
+	}
+
+	public async getCommandLineHelpForCommand(commandName: string): Promise<string> {
```
Mitko-Kerezov: Do we need this to be public?

PR #1030, opened 2017-12-04
Fix deploy on device to work with string and object identifiers.

KristianDD:
Related to https://github.com/NativeScript/nativescript-cli/pull/3120

### DIFF
--- a/mobile/android/logcat-helper.ts
+++ b/mobile/android/logcat-helper.ts
@@ -72,6 +72,24 @@ export class LogcatHelper implements Mobile.ILogcatHelper {
 		return logcatStream;
 	}
 
+	public async dump(deviceIdentifier: string): Promise<void> {
+		const adb: Mobile.IDeviceAndroidDebugBridge = this.$injector.resolve(DeviceAndroidDebugBridge, { identifier: deviceIdentifier });
+		const logcatDumpStream = await adb.executeCommand(["logcat", "-d"], { returnChildProcess: true });
+
+		const lineStream = byline(logcatDumpStream.stdout);
+		lineStream.on('data', (line: NodeBuffer) => {
+			const lineText = line.toString();
+			this.$logger.trace(lineText);
+		});
+
+		logcatDumpStream.on("close", (code: number) => {
+			logcatDumpStream.removeAllListeners();
+			lineStream.removeAllListeners();
+		});
+
+		this.$processService.attachToProcessExitSignals(this, logcatDumpStream.kill);
+	}
+
 	/**
 	 * Stops the logcat process for the specified device if keepSingleProcess is not passed on start
 	 */
--- a/mobile/mobile-core/devices-service.ts
+++ b/mobile/mobile-core/devices-service.ts
@@ -413,9 +413,20 @@ export class DevicesService extends EventEmitter implements Mobile.IDevicesServi
 	}
 
 	@exported("devicesService")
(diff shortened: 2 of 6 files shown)


mobile/mobile-core/devices-service.ts
```diff
@@ -344,9 +344,20 @@ export class DevicesService extends EventEmitter implements Mobile.IDevicesServi
...
 
 	@exported("devicesService")
-	public deployOnDevices(deviceIdentifiers: string[], packageFile: string, packageName: string): Promise<void>[] {
+	public deployOnDevices(deviceIdentifiers: string[], packageFile: string, packageName: string | Mobile.IProjectIdentifier): Promise<void>[] {
 		this.$logger.trace(`Called deployOnDevices for identifiers ${deviceIdentifiers} for packageFile: ${packageFile}. packageName is ${packageName}.`);
-		return _.map(deviceIdentifiers, deviceIdentifier => this.deployOnDevice(deviceIdentifier, packageFile, packageName));
+		return _.map(deviceIdentifiers, async deviceIdentifier => {
+			const device = this.getDeviceByIdentifier(deviceIdentifier);
+
+			let identifier: string;
```
rosen-vladimirov: the next few lines should be outside of the map, as they are not specific for device and there's no need to execute them each time

mobile/mobile-core/devices-service.ts
```diff
@@ -344,9 +344,20 @@ export class DevicesService extends EventEmitter implements Mobile.IDevicesServi
 	}
 
 	@exported("devicesService")
-	public deployOnDevices(deviceIdentifiers: string[], packageFile: string, packageName: string): Promise<void>[] {
+	public deployOnDevices(deviceIdentifiers: string[], packageFile: string, packageName: string | Mobile.IProjectIdentifier): Promise<void>[] {
 		this.$logger.trace(`Called deployOnDevices for identifiers ${deviceIdentifiers} for packageFile: ${packageFile}. packageName is ${packageName}.`);
-		return _.map(deviceIdentifiers, deviceIdentifier => this.deployOnDevice(deviceIdentifier, packageFile, packageName));
+		return _.map(deviceIdentifiers, async deviceIdentifier => {
```
rosen-vladimirov: you do not need the `async` keyword here

KristianDD: I put it there because the method returns an array of promises and if getDeviceByIdentifier fails an exception was thrown instead of returning a rejected promise.

mobile/mobile-core/devices-service.ts
KristianDD: I can't get the package name if I don't have the platform of the device.

rosen-vladimirov: :man_facepalming: I'm stupid, ignore my comment

mobile/android/logcat-helper.ts
```diff
@@ -72,6 +72,24 @@ export class LogcatHelper implements Mobile.ILogcatHelper {
 		return logcatStream;
 	}
 
+	public async dump(deviceIdentifier: string): Promise<void> {
+		const adb: Mobile.IDeviceAndroidDebugBridge = this.$injector.resolve(DeviceAndroidDebugBridge, { identifier: deviceIdentifier });
+		const logcatDumpStream = await adb.executeCommand(["logcat", "-d"], { returnChildProcess: true });
```
rosen-vladimirov: does this command work on all API levels?

KristianDD: Tested the command manually on a device running Android 4.2.1(API Level 17) and it works as expected.

PR #1031, opened 2017-12-05
use process.env.JDK_HOME > process.env.JAVA_HOME

bluelovers:
some ppl use JRE and JDK at same time

and some JRE didn't have javac

this small change make it use JDK_HOME first if it exists


rosen-vladimirov:
Hey @bluelovers ,
Please excuse us for the delay and thanks for your efforts. I'm merging this PR and it will be included in the next major version of NativeScript CLI.

PR #1032, opened 2017-12-06
Make bundle option string

Mitko-Kerezov:
Ping @KristianDD @rosen-vladimirov

### DIFF
--- a/options.ts
+++ b/options.ts
@@ -184,8 +184,13 @@ export class OptionsBase {
 		// For backwards compatibility
 		// Previously profileDir had a default option and calling `this.$options.profileDir` always returned valid result.
 		// Now the profileDir should be used from $settingsService, but ensure the `this.$options.profileDir` returns the same value.
-		this.$settingsService.setSettings({ profileDir: this.argv.profileDir});
+		this.$settingsService.setSettings({ profileDir: this.argv.profileDir });
 		this.argv.profileDir = this.argv["profile-dir"] = this.$settingsService.getProfileDir();
+
+		// Default to "nativescript-dev-webpack" if only `--bundle` is passed
+		if (this.argv.bundle !== undefined) {
+			this.argv.bundle = this.argv.bundle || "nativescript-dev-webpack";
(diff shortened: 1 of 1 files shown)


options.ts
```diff
@@ -184,9 +184,14 @@ export class OptionsBase {
 		// For backwards compatibility
 		// Previously profileDir had a default option and calling `this.$options.profileDir` always returned valid result.
 		// Now the profileDir should be used from $settingsService, but ensure the `this.$options.profileDir` returns the same value.
-		this.$settingsService.setSettings({ profileDir: this.argv.profileDir});
+		this.$settingsService.setSettings({ profileDir: this.argv.profileDir });
 		this.argv.profileDir = this.argv["profile-dir"] = this.$settingsService.getProfileDir();
 
+		// Default to "nativescript-dev-webpack" if only `--bundle` is passed
+		if (this.argv.bundle !== undefined) {
+			this.argv.bundle = this.argv.bundle || "nativescript-dev-webpack";
```
KristianDD: It might be better to move as a constant and even better if the option is "webpack" and we have a map like:
```typescript
const bundlersMap = {
    webpack: "nativescript-dev-webpack"
}
```
This will make us less dependent on the actual name of the plugin and would be easier for users.

Mitko-Kerezov: I'd rather we didn't have a map, because introducing a new bundling plugin would require the release of a new CLI version in order for said plugin to pass this validation.

PanayotCankov: I think it is ok for now, and if we have to support two bundlers, we will need a map, and an extensibility model that will let each bundle register itself in that map by a alias (e.g. webpack: nativescript-dev-webpack) or by convention...

PanayotCankov: Yet after reading this: https://github.com/NativeScript/nativescript-cli/pull/3246/files#diff-a67359fb6feeda0833946840738b31a3R20, I would prefer `--bundle webpack`, `--bundle nativescript-dev-webpack` is way too verbose.

PR #1033, opened 2017-12-10
Proxy improvements

rosen-vladimirov:
### Use proxy-lib package when working behind proxy
The `proxy-lib` package provides required methods for setting and getting proxy. Its implementation is more secure when stroing credentials, so switch to it instead of current implementation.
Remove the credentials-manager from CLI as it is no longer needed.
Remove the CredentialsManager.exe as well and its source code.

### Fix the error `Cannot read property ExceptionMessage of null` when wokring behind proxy
Add correct null checks, so we'll no longer see this error.

### Add `proxyAuthenticationRequired` property to errors
In case http request fails because a proxy requires authentication, add new property to the error object - `proxyAuthenticationRequired`. This will allow the callers to easily detect such cases and handle them in the way they need.
Add the option to have such property to error object thrown by `errors.fail` method.


appbuilder/services/npm-service.ts
```diff
@@ -338,6 +338,18 @@ export class NpmService implements INpmService {
 		return this.$childProcess.spawnFromEvent(this.npmExecutableName, npmArguments, "close", { cwd: projectDir, stdio: "inherit" });
 	}
 
+	private getCredentialsFromAuth(auth: string): ICredentials {
```
rosen-vladimirov: remove this method - use it from proxy-lib package

declarations.d.ts
```diff
@@ -906,35 +903,22 @@ interface ICredentialsService {
...
+	 * @param cacheData {IProxyLibSettings} Data to be cached.
+	 * @returns {Promise<void>} The cache.
 	 */
-	setCache(cacheData: IProxyCache): IProxyCache;
+	setCache(settings: IProxyLibSettings): Promise<void>;
 
 	/**
 	 * Retrieves proxy cache data.
-	 * @returns {IProxyCache} The cache.
+	 * @returns {Promise<IGetProxySettings>} The cache.
```
Mitko-Kerezov: I believe the doc string is a bit off here

PR #1037, opened 2017-12-29
Fix detection of Javac version

rosen-vladimirov:
The command `javac -version` prints result to stderr when JAVA 8 is used and to stdout when JAVA 9 is used. Current check in CLI uses the stderr output, so when JAVA 9 is installed it fails to detect the correct version.
In order to support both JAVA 8 and JAVA 9, capture both stdout and stderr and get the version from there.
Also remove unneeded check for Java version - we care about JAVA Compiler, which is included in JDK.

### DIFF
--- a/test/unit-tests/sys-info-base.ts
+++ b/test/unit-tests/sys-info-base.ts
@@ -19,7 +19,6 @@ interface IChildProcessResultDescription {
 interface IChildProcessResults {
 	uname: IChildProcessResultDescription;
 	npmV: IChildProcessResultDescription;
-	javaVersion: IChildProcessResultDescription;
 	javacVersion: IChildProcessResultDescription;
 	nodeGypVersion: IChildProcessResultDescription;
 	xCodeVersion: IChildProcessResultDescription;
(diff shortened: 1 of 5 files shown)


test/unit-tests/sys-info-base.ts
```diff
@@ -214,12 +210,33 @@ describe("sysInfoBase", () => {
...
+				const actualJavaCompilerVersion = await sysInfoBase.getJavaCompilerVersion();
+				assert.deepEqual(actualJavaCompilerVersion, javaCompilerVersion);
+			};
+
+			it("returns correct javac version when it is printed on stderr (Java 8)", () => {
+				return verifyJavaCompilerVersion("1.8.0_152", "stderr");
+			});
+
+			it("returns correct javac version when it is printed on stdout (Java 9)", () => {
+				return verifyJavaCompilerVersion("9.0.1", "stdout");
```
Plamen5kov: @rosen-vladimirov some distributions of java return `9` when `java -version` is called, in which case I think the regex might break. I haven't tested it but I encountered such an issue in the past.

rosen-vladimirov: Thanks for pointing this. In case 9 is returned, the current code will work fine, but [this one](https://github.com/NativeScript/nativescript-cli/pull/3294/files#diff-8288e0e434a7344884ce096d3dfd239bR115) will fail. I'll handle the case there.

rosen-vladimirov: @Plamen5kov fixed in this PR https://github.com/NativeScript/nativescript-cli/pull/3294/

Plamen5kov: @rosen-vladimirov I tested it, and CLI provides a message to install java 1.8.0.

PR #1040, opened 2018-01-11
Add getSpinner method to $progressIndicator

rosen-vladimirov:
Add `getSpinner` method to `$progressIndicator` - it will return a new instance of clui.Spinner in case terminal is interactive. In case it is not - a mocked instance will be returned.
This way in CI builds the spinner will not print tons of repeated messages.

### DIFF
--- a/progress-indicator.ts
+++ b/progress-indicator.ts
@@ -1,23 +1,29 @@
+import { isInteractive } from './helpers';
+
+const clui = require("clui");
+
 export class ProgressIndicator implements IProgressIndicator {
 	constructor(private $logger: ILogger) { }
 
 	public async showProgressIndicator<T>(promise: Promise<T>, timeout: number, options?: { surpressTrailingNewLine?: boolean }): Promise<T> {
 		const surpressTrailingNewLine = options && options.surpressTrailingNewLine;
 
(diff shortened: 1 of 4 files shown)


progress-indicator.ts
```diff
@@ -26,5 +32,18 @@ export class ProgressIndicator implements IProgressIndicator {
...
 		return tempPromise;
 	}
+
+	public getSpinner(message: string): ISpinner {
+		if (isInteractive()) {
+			return new clui.Spinner(message);
+		} else {
+			let msg = message;
+			return {
+				start: () => this.$logger.info(msg),
```
TsvetanMilanov: For cloud builds this will print for example `Installing tns-android`  and we are going to loose the broken "progress indicator" which we currently have. Without it our users may wonder is the build slow or is there some other problem (network connectivity issue on their side, or some bug on our side).
Do you think it will be better to have some kind of progress indicator instead of just one message?

rosen-vladimirov: I think we shouldn't rely on the cloud build logs in order to show progress indication on the client side. IMO this should be handled on the client side in any other meaningful way.
@tsvetie what do you think?

PR #1043, opened 2018-01-29
feat(net): Add method to check if port is in LISTEN state

rosen-vladimirov:
Add method in net service to check if port is in LISTEN state. Add tests for it.
The method has different implementation for macOS, Linux and Windows - added tests for each of the implementations.

### DIFF
--- a/declarations.d.ts
+++ b/declarations.d.ts
@@ -1736,6 +1736,25 @@ interface IVersionData {
 	patch: string;
 }
 
+interface IWaitForPortListenData {
+	/**
+	 * Port to be checked.
+	 * @type {number}
+	 */
+	port: number;
+
+	/**
+	 * Max amount of time in milliseconds to wait.
+	 * @type {number}
+	 */
+	timeout: number;
+	/**
+	 * @optional The amount of time between each check.
+	 * @type {number}
+	 */
+	interval?: number;
+}
+
 /**
  * Wrapper for net module of Node.js.
  */
@@ -1761,6 +1780,13 @@ interface INet {
 	 * @return {Promise<boolean>} true if the port is available.
 	 */
 	isPortAvailable(port: number): Promise<boolean>;
+
+	/**
+	 * Waits for port to be in LISTEN state.
+	 * @param {IWaitForPortListenData} waitForPortListenData Data describing port, timeout and interval.
+	 * @returns {boolean} true in case port is in LISTEN state, false otherwise.
+	 */
+	waitForPortToListen(waitForPortListenData: IWaitForPortListenData): Promise<boolean>;
 }
 
 interface IProcessService {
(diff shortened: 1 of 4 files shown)


services/net-service.ts
```diff
@@ -71,6 +77,53 @@ export class Net implements INet {
...
+			}
+		};
+
+		const platform = this.$osInfo.platform();
+		const currentPlatformData = platformData[platform];
+		if (!currentPlatformData) {
+			this.$errors.failWithoutHelp(`Unable to check for free ports on ${platform}. Supported platforms are: ${_.keys(platformData).join(", ")}`);
+		}
+
+		while (true) {
```
Mitko-Kerezov: Instead of `while (true)` you can set the condition `while (new Date().getTime() < endTime)` and delete the `break;` code below.

IMO it'd be a bit clearer than looking at what seems to be an endless loop at a first glance

services/net-service.ts
```diff
@@ -71,6 +77,53 @@ export class Net implements INet {
 
 		return startPort;
 	}
+
+	public async waitForPortToListen(port: number, timeout: number, interval?: number): Promise<boolean> {
+		interval = interval || Net.DEFAULT_INTERVAL;
+		const endTime = new Date().getTime() + timeout;
+		const platformData: IDictionary<any> = {
```
Mitko-Kerezov: You can introduce an strongly typed interface here instead of `any`. Not a merge-stopper though

declarations.d.ts
```diff
@@ -1761,6 +1761,15 @@ interface INet {
...
 	isPortAvailable(port: number): Promise<boolean>;
+
+	/**
+	 * Waits for port to be in LISTEN state.
+	 * @param {number} port Port to be checked.
+	 * @param {number} timeout Max amount of time in milliseconds to wait.
+	 * @param {number} inteval @optional The amount of time between each check.
+	 * @returns {boolean} true in case port is in LISTEN state, false otherwise.
+	 */
+	waitForPortToListen(port: number, timeout: number, interval?: number): Promise<boolean>;
```
Mitko-Kerezov: You can optionally introduce an interface here instead of accepting three different arguments

services/net-service.ts
rosen-vladimirov: No need to introduce an interface for an object that's created and used inside a single method.

services/net-service.ts
rosen-vladimirov: The current flow executes at least one check, no matter of the passed timeout. Changing the `while` statement will change this behavior.
Also the current workflow checks if a timeout is passed and performs at least one check after that. Changing the `while` statement will change this behavior as well, as we have `sleep` at the end of the `while`. The `sleep` cannot be moved at the first line of the `while` as we may not have to wait for it.
So the desired workflow is:
1. Check if the port is in LISTEN state.
2. If yes - break. If not, check if timeout is passed and in case not - sleep for specified interval. 
3. Check if port is in LISTEN state.
4. If yes - break. If not, check if timeout is passed and in case not - sleep for specified interval.
... Repeat until timeout is reached.

rosen-vladimirov:
ping @Mitko-Kerezov , comments are fixed

PR #1047, opened 2018-02-12
feat(extensions): Allow generation of help content for extension commands

rosen-vladimirov:
Each CLI extension may add new commands to CLI. Allow showing command help for these commands in case:
1. Commmand execution fails.
2. Users need more information for command - `tns <command> --help`
3. Users need full html help for command - `tns help <command>`.

Each extension that wants to use this functionality will have to add `nativescript` key in its `package.json` and add `docs` key in it. The `docs` key must point to the directory where the help content (.md files) is located, relative to the root directory of the extension.
When CLI needs to show command line help for extension's command, it will search the docs directories of all extensions.
When CLI needs to show HTML help for the extension's commands, it will generate `html` directory right next to the docs dir in the extension.

Move extensibility.d.ts from {N} to mobile-cli-lib just to ensure correct transpilation.

### Add different basic page for extensions html pages
When generating html pages for extensions commands, use a different html
page that has the extension name as header at the top.
Add a placeholder and replace it when generating the html page.

### DIFF
--- a/services/help-service.ts
+++ b/services/help-service.ts
@@ -10,6 +10,7 @@ export class HelpService implements IHelpService {
 	private static RELATIVE_PATH_TO_STYLES_CSS_REGEX = /@RELATIVE_PATH_TO_STYLES_CSS@/g;
 	private static RELATIVE_PATH_TO_IMAGES_REGEX = /@RELATIVE_PATH_TO_IMAGES@/g;
 	private static RELATIVE_PATH_TO_INDEX_REGEX = /@RELATIVE_PATH_TO_INDEX@/g;
+	private static EXTENSION_NAME_REGEX = /@EXTENSION_NAME@/g;
 	private static MARKDOWN_LINK_REGEX = /\[([\w \-\`\<\>\*\:\\]+?)\]\([\s\S]+?\)/g;
 	private static SPAN_REGEX = /([\s\S]*?)(?:\r?\n)?<span.*?>([\s\S]*?)<\/span>(?:\r?\n)*/g;
 	private static NEW_LINE_REGEX = /<\/?\s*?br\s*?\/?>/g; // <br>, <br > <br/> <br />
@@ -25,6 +26,10 @@ export class HelpService implements IHelpService {
 
 	private get pathToBasicPage(): string {
 		return path.join(this.$staticConfig.HTML_COMMON_HELPERS_DIR, "basic-page.html");
+	}
+
+	private get pathToBasicPageForExtensions(): string {
+		return path.join(this.$staticConfig.HTML_COMMON_HELPERS_DIR, "basic-extensions-page.html");
 	}
 
 	private pathToImages = this.$staticConfig.HTML_CLI_HELPERS_DIR;
@@ -66,6 +71,7 @@ export class HelpService implements IHelpService {
 
 		const installedExtensionsData = this.$extensibilityService.getInstalledExtensionsData();
 
+		const basicHtmlPageForExtensions = this.$fs.readText(this.pathToBasicPageForExtensions);
 		for (const extensionData of installedExtensionsData) {
 			const docsDir = extensionData.docs;
 
@@ -80,7 +86,7 @@ export class HelpService implements IHelpService {
 				this.$fs.ensureDirectoryExists(htmlDirFullPath);
 				const extensionMdFiles = this.$fs.enumerateFilesInDirectorySync(docsDir);
 				try {
-					await Promise.all(_.map(extensionMdFiles, markdownFile => this.createHtmlPage(basicHtmlPage, markdownFile, docsDir, htmlDirFullPath)));
+					await Promise.all(_.map(extensionMdFiles, markdownFile => this.createHtmlPage(basicHtmlPageForExtensions, markdownFile, docsDir, htmlDirFullPath, extensionData.extensionName)));
 				} catch (err) {
 					this.$logger.warn(`Unable to generate html help for extension ${extensionData.extensionName}. Error is: ${err.message}`);
 				}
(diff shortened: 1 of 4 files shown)


services/help-service.ts
```diff
@@ -61,7 +67,34 @@ export class HelpService implements IHelpService {
...
+			const docsDir = extensionData.docs;
+
+			if (docsDir) {
+				this.$logger.trace(`Start generation of html help content for extension ${extensionData.extensionName}`);
+				if (!this.$fs.exists(docsDir)) {
+					this.$logger.warn(`Unable to generate html help pages for extension ${extensionData.extensionName} as the docs directory ${docsDir} does not exist.`);
+					continue;
+				}
+
+				const htmlDirFullPath = path.join(path.dirname(docsDir), "html");
```
Mitko-Kerezov: You can extract `"html"` as a constant either in `constants.ts` or in this file and reuse it here and in `tryOpeningSelectedPage`

services/help-service.ts
```diff
@@ -153,7 +202,20 @@ export class HelpService implements IHelpService {
...
 		this.$logger.trace("Reading help for command '%s'. FileName is '%s'.", commandName, mdFileName);
 
-		const markdownFile = _.find(this.$fs.enumerateFilesInDirectorySync(this.pathToManPages), file => path.basename(file) === mdFileName);
+		let markdownFile = _.find(this.$fs.enumerateFilesInDirectorySync(this.pathToManPages), file => path.basename(file) === mdFileName);
+		if (!markdownFile) {
+			// Try getting help from extensions
+			const installedExtensionsData = this.$extensibilityService.getInstalledExtensionsData();
+			for (const extensionData of installedExtensionsData) {
+				if (extensionData.docs) {
+					markdownFile = _.find(this.$fs.enumerateFilesInDirectorySync(extensionData.docs), file => path.basename(file) === mdFileName);
```
Mitko-Kerezov: You can extract `file => path.basename(file) === mdFileName` lambda to a separate `const` variable and reuse it here and up above on line 205

services/help-service.ts
```diff
@@ -137,7 +171,22 @@ export class HelpService implements IHelpService {
 	private tryOpeningSelectedPage(htmlPage: string): boolean {
 		const fileList = this.$fs.enumerateFilesInDirectorySync(this.pathToHtmlPages);
 		this.$logger.trace("File list: " + fileList);
-		const pageToOpen = _.find(fileList, file => path.basename(file) === htmlPage);
+		let pageToOpen = _.find(fileList, file => path.basename(file) === htmlPage);
+		if (!pageToOpen) {
+			const installedExtensionsData = this.$extensibilityService.getInstalledExtensionsData();
+
+			for (const extensionData of installedExtensionsData) {
```
Mitko-Kerezov: You can extract [this](https://github.com/telerik/mobile-cli-lib/blob/9412c49b11f6a69b58b6970a9428e916c6bda736/services/help-service.ts#L178-L188) as a separate function and reuse it in `tryOpeningSelectedPage` and in `readMdFileForCommand`

test/unit-tests/services/help-service.ts
```diff
@@ -434,5 +439,70 @@ and another one`
...
+					version: "1.2.3",
+					pathToExtension: "extension3path"
+				});
+
+				$extensibilityService.getInstalledExtensionsData = (): IExtensionData[] => extensionsData;
+
+				const helpService = injector.resolve<IHelpService>("helpService");
+				await helpService.showCommandLineHelp("foo");
+				const output = injector.resolve("logger").output;
+				assert.isTrue(output.indexOf("bla end") >= 0);
```
Mitko-Kerezov: You can extract `"bla end"` to a `const` variable

test/unit-tests/services/help-service.ts
```diff
@@ -434,5 +439,70 @@ and another one`
...
+
+				assert.equal(enumerateFilesInDirectorySyncCalledCounter, expectedEnumerateFilesInDirectorySyncCalledCounter, "The enumerateFilesInDirectorySync method must be called exactly three times - one for CLI help, one for extension2 and one for extension3");
```
Mitko-Kerezov: This error message is incorrect because the number of called times differs and the extensions differ with each test

rosen-vladimirov:
Ping @Mitko-Kerezov - comments are fixed.

test/unit-tests/services/help-service.ts
```diff
@@ -434,5 +445,72 @@ and another one`
...
+
+				assert.equal(enumerateFilesInDirectorySyncCalledCounter, expectedEnumerateFilesInDirectorySyncCalledCounter,
+					`The enumerateFilesInDirectorySync method must be called exactly ${enumerateFilesInDirectorySyncCalledCounter} times.`);
+			};
+
+			it("help from a single installed extension is successfully loaded", async () => {
+				await assertData(2);
+			});
+
+			it("when multiple extensions are installed, the full ", async () => {
```
Mitko-Kerezov: this test's title has a rather abrupt ending

services/help-service.ts
```diff
@@ -93,7 +126,7 @@ export class HelpService implements IHelpService {
 	}
 
 	// This method should return Promise in order to generate all html pages simultaneously.
-	private async createHtmlPage(basicHtmlPage: string, pathToMdFile: string): Promise<void> {
+	private async createHtmlPage(basicHtmlPage: string, pathToMdFile: string, pathToMdPages: string, pathToHtmlPages: string, extensionName?: string): Promise<void> {
```
Mitko-Kerezov: Not a merge-stopper as this is a `private` method, but I'd rather we didn't have as many arguments when we could do with a single object

test/unit-tests/services/help-service.ts
rosen-vladimirov: 😸

PR #1049, opened 2018-02-19
fix: allow force dispose

Mitko-Kerezov:
Allow passing `force` to `dispose` operation. Some services which implement `IDisposable` need to have a way to forcefully shut down.

Ping @rosen-vladimirov

### DIFF
--- a/declarations.d.ts
+++ b/declarations.d.ts
@@ -169,7 +169,7 @@ declare module Server {
 }
 
 interface IDisposable {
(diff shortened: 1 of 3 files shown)


Mitko-Kerezov:
Ping @rosen-vladimirov with fixed comments

Mitko-Kerezov:
Closing this in regard to https://github.com/NativeScript/nativescript-cli/pull/3376

PR #1050, opened 2018-02-19
feat(iOSSimulator): Support for multiple simulators

Fatme:


test/unit-tests/mobile/ios-simulator-discovery.ts
```diff
@@ -38,17 +34,30 @@ function createTestInjector(): IInjector {
...
+		vendor: 'Apple',
+		platform: 'iOS',
+		status: 'Connected',
+		errorHelp: null,
+		isTablet: false,
+		type: 'Emulator'
+	};
+}
+
+describe.only("ios-simulator-discovery", () => {
```
Fatme: `.only` should  be removed

PR #1054, opened 2018-02-27
feat(Analytics): Respect playground key from package.json file

Fatme:
We need to track users that export their projects from playground and opens them in CLI or Sidekick. To support that we introduce playground key in nativescript key in package.json file. For example:
```
{
	"nativescript": {
		"playground": {
			"id": "some user quid",
			"usedTutorial": false // is not obligatory. In case it is present, can be true or false
		}
	}
}
```

In  case when package.json file contains playground key, {N} CLI reads playground data and saves it in userSettings file. If usedTutorial=true is already saved in userSettings file, {N} CLI does not overwrite it. After that {N} CLI deletes playground key from package.json file.

In case when package.json file does not contain playground key, {N} CLI checks if playground key is already saved in userSettings file. If this is the case, {N} CLI reads playground data from userSettings file.

### DIFF
--- a/services/commands-service.ts
+++ b/services/commands-service.ts
@@ -16,7 +16,8 @@ export class CommandsService implements ICommandsService {
 	private static HOOKS_COMMANDS_DELIMITER = "-";
 	private areDynamicSubcommandsRegistered = false;
 
-	constructor(private $commandsServiceProvider: ICommandsServiceProvider,
+	constructor(private $analyticsSettingsService: IAnalyticsSettingsService,
+		private $commandsServiceProvider: ICommandsServiceProvider,
 		private $errors: IErrors,
 		private $fs: IFileSystem,
 		private $hooksService: IHooksService,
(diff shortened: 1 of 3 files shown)


services/commands-service.ts
```diff
@@ -23,6 +23,7 @@ export class CommandsService implements ICommandsService {
 		private $injector: IInjector,
 		private $logger: ILogger,
 		private $options: ICommonOptions,
+		private $playgroundService: IPlaygroundService,
```
rosen-vladimirov: Can we use analyticsSettingsService here instead of playgroundService. It sounds more appropriate to use it in commandsService, but this is not a merge stopper.

Fatme: Yes, this is also an option. But in this case I need to pass `null` as parameter because `getPlaygroundInfo()` from `analyticsSettingsService` is publicly exported and has mandatory `projectDir` param.  The call should be:
```
const playgroundInfo = await this.$analyticsSettingsService.getPlaygroundInfo(null);
```

rosen-vladimirov: Most probably `analyticsSettingsService` also does not require projectDir as mandatory arg

PR #1055, opened 2018-02-27
refactor: update helpers

Plamen5kov:
based on branch: `kddimitrov/nsconfig-app-folder`
update for helper methods necessary in the CLI
related PR: https://github.com/NativeScript/nativescript-cli/pull/3398

### DIFF
--- a/helpers.ts
+++ b/helpers.ts
@@ -77,6 +77,17 @@ export function trackDownloadProgress(destinationStream: NodeJS.WritableStream, 
 	return progressStream;
 }
 
+export function isBuildFromCLI(item: string) {
+	return item.match(/.*\.aar/) ||
+			item.match(".*include.gradle");
+}
+
+export function isRecommendedAarFile(foundAarFile: string, packageJsonPluginName: string) {
+	const filename = foundAarFile.replace(/^.*[\\\/]/, '');
+	packageJsonPluginName = packageJsonPluginName.replace(/[\-]/g, "_");
+	return `${packageJsonPluginName}.aar` === filename;
+}
+
 export async function executeActionByChunks<T>(initialData: T[] | IDictionary<T>, chunkSize: number, elementAction: (element: T, key?: string | number) => Promise<any>): Promise<void> {
 	let arrayToChunk: (T | string)[];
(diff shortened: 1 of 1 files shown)


helpers.ts
```diff
@@ -52,6 +52,17 @@ export function trackDownloadProgress(destinationStream: NodeJS.WritableStream,
 	return progressStream;
 }
 
+export function isBuildFromCLI(item: string) {
```
rosen-vladimirov: this function needs a return type.
Also its name is misleading. The implementation checks if the passed string contains the string `.aar` or `include.gradle`.

helpers.ts
```diff
@@ -52,6 +52,17 @@ export function trackDownloadProgress(destinationStream: NodeJS.WritableStream,
...
+	return item.match(/.*\.aar/) ||
+			item.match(".*include.gradle");
+}
+
+export function isRecommendedAarFile(foundAarFile: string, packageJsonPluginName: string) {
```
rosen-vladimirov: this function needs a return type.

helpers.ts
```diff
@@ -52,6 +52,17 @@ export function trackDownloadProgress(destinationStream: NodeJS.WritableStream,
 	return progressStream;
 }
 
+export function isBuildFromCLI(item: string) {
+	return item.match(/.*\.aar/) ||
```
rosen-vladimirov: this regex will match strings like:
* filename.aar
* .aar
* test.aar.backup

Is this the desired behavior?

helpers.ts
```diff
@@ -52,6 +52,17 @@ export function trackDownloadProgress(destinationStream: NodeJS.WritableStream,
...
+			item.match(".*include.gradle");
```
rosen-vladimirov: this regex will match strings like:
* filename.include.gradle
* include.gradle
* include.gradle.backup

Is this the desired behavior?

helpers.ts
Plamen5kov: yes

helpers.ts
Plamen5kov: yes

PR #1058, opened 2018-03-07
Getting started improvements

Fatme:
When build or run command is executed, checks if the environment is properly configured for executing local builds. If it is not properly configured, {N} CLI shows a prompt and user is able to choice if to run setup script or use cloud builds.

If cloud builds option is selected, {N} CLI installs the nativescript-cloud extension and prints that user can use the $ tns login command to log in with his/her account and then $ tns cloud run/build android/ios command to build his/her app in the cloud.

If setup script option is selected, {N} CLI runs the setup script to try and automatically configure the environment. If the environment is not properly configured after executing the setup script, {N} CLI shows a prompt and user is able to choice if to use cloud builds or manually setup the environment. If the manually setup option is selected, {N} CLI prints: "To be able to build for android/ios, verify that your environment is configured according to the system requirements described at https://docs.nativescript.org/start/ns-setup-os-x.}"

### DIFF
--- a/declarations.d.ts
+++ b/declarations.d.ts
@@ -1354,7 +1354,7 @@ interface IDoctorService {
 	runSetupScript(): Promise<ISpawnResult>;
 	/**
 	 * Checks if the envrironment is properly configured and it is possible to execute local builds
-	 * @param platform The current platform
+	 * @param platform @optional The current platform
 	 * @returns {Promise<boolean>} true if the environment is properly configured for local builds
 	 */
 	canExecuteLocalBuild(platform?: string): Promise<boolean>;
--- a/declarations.d.ts
+++ b/declarations.d.ts
@@ -550,7 +550,7 @@ interface IOpener {
 
 interface IErrors {
 	fail(formatStr: string, ...args: any[]): never;
-	fail(opts: { formatStr?: string; errorCode?: number; suppressCommandHelp?: boolean, proxyAuthenticationRequired?: boolean }, ...args: any[]): never;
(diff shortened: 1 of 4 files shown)


services/commands-service.ts
```diff
@@ -33,6 +38,7 @@ export class CommandsService implements ICommandsService {
 	}
 
 	public async executeCommandUnchecked(commandName: string, commandArguments: string[]): Promise<boolean> {
+		this.commands.push({ commandName: commandName, commandArguments: commandArguments });
```
Mitko-Kerezov: You can use the short syntax here `this.commands.push({ commandName, commandArguments });`

declarations.d.ts
```diff
@@ -1301,6 +1301,17 @@ interface IDoctorService {
...
 	 */
 	printWarnings(configOptions?: { trackResult: boolean }): Promise<boolean>;
+	/** 
+	 * Runs the setup script on host machine
+	 * @returns {Promise<ISpawnResult>}
+	*/
+	runSetupScript(): Promise<ISpawnResult>;
+	/**
+	 * Checks if the envrironment is properly configured and it is possible to execute local builds
+	 * @param platform The current platform
```
Mitko-Kerezov: You can mark this as `@optional`

services/commands-service.ts
```diff
@@ -72,8 +78,11 @@ export class CommandsService implements ICommandsService {
 				this.$logger.printMarkdown(~suggestionText.indexOf('%s') ? require('util').format(suggestionText, commandArguments) : suggestionText);
 			}
 
+			_.remove(this.commands, commandData => commandData.commandName === commandName);
```
Mitko-Kerezov: I think you can safely use `pop` here and below instead of `_.remove` because commands would be pushed one over the other just like a stack

PR #1059, opened 2018-03-09
fix(userSettings): Don't fail when user settings file is not a valid json

Fatme:
In case when user executes some command and by some reasons user settings file is not a valid json, {N} CLI throws "Unexpected token in JSON at position 0" error. After that user is not able to execute any cli command until the file exists.

https://github.com/NativeScript/nativescript-cli/issues/3151

### DIFF
--- a/services/user-settings-service.ts
+++ b/services/user-settings-service.ts
@@ -1,4 +1,5 @@
 import * as path from "path";
+import { parseJson } from "../helpers";
 
 export class UserSettingsServiceBase implements IUserSettingsService {
 	private userSettingsFilePath: string = null;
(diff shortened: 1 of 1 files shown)


Fatme:
@rosen-vladimirov @Mitko-Kerezov

services/user-settings-service.ts
```diff
@@ -90,7 +91,14 @@ export class UserSettingsServiceBase implements IUserSettingsService {
 			}
 		}
 
-		this.userSettingsData = this.$fs.readJson(this.userSettingsFilePath);
+		const data = this.$fs.readText(this.userSettingsFilePath);
+
+		try {
+			this.userSettingsData = parseJson(data);
+		} catch (err) {
+			this.$logger.trace(`Error while trying to parseJson ${this.userSettingsFilePath} file.`);
```
rosen-vladimirov: can you add the error in the message and the `data` as well.

PR #1082, opened 2018-04-05
fix (ios-simulator): Show logs from all running iOS simulators

Fatme:
In case when more than one simulator is running, logs are shown only from one of them.
Also fixes the behaviour when simulator's log process is interrupted in case when cli is used as library.

### DIFF
--- a/mobile/ios/simulator/ios-simulator-log-provider.ts
+++ b/mobile/ios/simulator/ios-simulator-log-provider.ts
@@ -1,20 +1,32 @@
 import { ChildProcess } from "child_process";
 
 export class IOSSimulatorLogProvider implements Mobile.IiOSSimulatorLogProvider {
-	private isStarted: boolean;
+	private simulatorsLoggingEnabled: IDictionary<boolean> = {};
 
 	constructor(private $iOSSimResolver: Mobile.IiOSSimResolver,
 		private $deviceLogProvider: Mobile.IDeviceLogProvider,
 		private $devicePlatformsConstants: Mobile.IDevicePlatformsConstants,
+		private $logger: ILogger,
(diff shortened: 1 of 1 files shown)


mobile/ios/simulator/ios-simulator-log-provider.ts
```diff
@@ -1,21 +1,27 @@
 import { ChildProcess } from "child_process";
 
 export class IOSSimulatorLogProvider implements Mobile.IiOSSimulatorLogProvider {
-	private isStarted: boolean;
+	private map: IDictionary<boolean> = {};
```
rosen-vladimirov: this could be named better, for example simulatorsLoggingEnabled
The usage will be:
```TypeScript
this.simulatorsLoggingEnabled[deviceIdentifier] = true;
```

mobile/ios/simulator/ios-simulator-log-provider.ts
```diff
@@ -1,21 +1,27 @@
...
-		if (!this.isStarted) {
+		if (!this.map[deviceIdentifier]) {
 			const deviceLogChildProcess: ChildProcess = this.$iOSSimResolver.iOSSim.getDeviceLogProcess(deviceIdentifier, 'senderImagePath contains "NativeScript"');
 
 			const action = (data: NodeBuffer | string) => {
 				this.$deviceLogProvider.logData(data.toString(), this.$devicePlatformsConstants.iOS, deviceIdentifier);
 			};
 
+			if (deviceLogChildProcess) {
+				deviceLogChildProcess.once("close", () => {
```
rosen-vladimirov: maybe we should attach to `.on("error"` of the deviceLogChildProcess

PR #1091, opened 2018-05-15
feat: Add method for getting macOS version

rosen-vladimirov:
Add new method to get the macOS version. Also change the verifyNodeVersion implementation, so the messages that are created there, can be used by Sidekick.

### DIFF
--- a/host-info.ts
+++ b/host-info.ts
@@ -1,3 +1,5 @@
+import { cache } from "./decorators";
+
 export class HostInfo implements IHostInfo {
 	private static WIN32_NAME = "win32";
 	private static PROCESSOR_ARCHITEW6432 = "PROCESSOR_ARCHITEW6432";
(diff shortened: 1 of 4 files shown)


host-info.ts
```diff
@@ -31,6 +46,50 @@ export class HostInfo implements IHostInfo {
...
+			this.$logger.trace(`Unable to get macOS version from ${systemProfileCommand}. Error is: ${err}`);
+		}
+
+		// https://en.wikipedia.org/wiki/Darwin_(operating_system)#Release_history
+		// Each macOS version is labeled 10.<version>, where it looks like <versions> is taken from the major version returned by os.release() (16.x.x for example) and subtracting 4 from it.
+		// So the version becomes "10.12" in this case.
+		const osRelease = this.$osInfo.release();
+		const majorVersion = osRelease && _.first(osRelease.split("."));
+		const macOSVersion = majorVersion && `10.${+majorVersion - 4}`;
+		this.$logger.trace(`macOS version based on os.release() is ${macOSVersion}.`);
```
Fatme: Maybe we need to log here `osRelease` value

test/unit-tests/host-info.ts
```diff
@@ -0,0 +1,96 @@
+import { Yok } from "../../yok";
+import { HostInfo } from "../../host-info";
+import { ErrorsStub, CommonLoggerStub } from "./stubs";
+import { assert } from "chai";
+
+// Use custom class as isDarwin has only getter in HostInfo, while for current tests we need to set it to true or false.
+class DarwinHostInfo extends HostInfo {
```
Fatme: Maybe this should be `MockHostInfo` or something similar.
Currently it's a little bit strange. If the name is `DarwinHostInfo` I'm expecting
```
public get isDarwin(): boolean {
     return true;
}
```

PR #1099, opened 2018-07-02
feat: Speed up device detection

rosen-vladimirov:
Speed up device detection by:
* removing duplicate calls to all device detection services
* cache the Promise of initialize method - this way even when multiple calls are executed simultaneously, the result of the first one will be used. This allows Sidekick to call the method asynchronously.
* do not execute iOS device detection in case `--emulator` is passed

Remove several public methods from the service - make them protected as we already have tests for them.
Remove all calls to detectCurrentlyAttachedDevices method outside of the service - the initialize method will do it anyway, so no need to call it externally.
Remove duplicate calls in the startLookingForDevices method - previously it had to filter which device detection services to call. As now each device detection service checks its platform, we can safely call all of them and rely on their filters to skip the execution.

### DIFF
--- a/declarations.d.ts
+++ b/declarations.d.ts
@@ -1246,7 +1246,11 @@ interface IProfileDir {
 	profileDir: string;
 }
 
-interface ICommonOptions extends IRelease, IDeviceIdentifier, IJustLaunch, IAvd, IAvailableDevices, IProfileDir {
+interface IHasEmulatorOption {
+	emulator: boolean;
+}
+
+interface ICommonOptions extends IRelease, IDeviceIdentifier, IJustLaunch, IAvd, IAvailableDevices, IProfileDir, IHasEmulatorOption {
 	argv: IYargArgv;
 	validateOptions(commandSpecificDashedOptions?: IDictionary<IDashedOption>): void;
 	options: IDictionary<any>;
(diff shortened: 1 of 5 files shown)


declarations.d.ts
```diff
@@ -1246,7 +1246,11 @@ interface IProfileDir {
 	profileDir: string;
 }
 
-interface ICommonOptions extends IRelease, IDeviceIdentifier, IJustLaunch, IAvd, IAvailableDevices, IProfileDir {
+interface IEmulator {
```
DimitarTachev: IEmulator -> IHasEmulator?

rosen-vladimirov: I've moved the interface from NativeScript CLI. The idea of such interfaces is to wrap the properties used in `$options` (the ones that you can pass with `--` on the terminal).In this case, the option the user can pass is `--emulator`, so the interface became `IEmulator`.

Ping @KristianDD , @Fatme  - what do you think - should we change this to IHasEmulator (without changing the `--` option.

KristianDD: IEmulatorTarget also sounds good to me

DimitarTachev: What I mean is that IEmulator should be an interface for something that wanna call itself an Emulator. Here we need an interface for something that just has an emulator property => we need something like IHasEmulator or IHasEmulatorOption/IOptionsWithEmulator if we wanna specify that these interfaces are wrapping some options values.

PR #1106, opened 2018-07-16
Kddimitrov/android livesync sockets

KristianDD:


mobile/android/android-device-hash-service.ts
```diff
@@ -52,22 +52,32 @@ export class AndroidDeviceHashService implements Mobile.IAndroidDeviceHashServic
 		return false;
 	}
 
-	public async generateHashesFromLocalToDevicePaths(localToDevicePaths: Mobile.ILocalToDevicePathData[], shasums: IStringDictionary): Promise<string[]> {
+	public async generateHashesFromLocalToDevicePaths(localToDevicePaths: Mobile.ILocalToDevicePathData[], initialShasums: IStringDictionary = {}): Promise<IStringDictionary> {
```
DimitarTachev: Why do we need to both update initialShasums and return the same object as a result?

mobile/android/android-device-hash-service.ts
```diff
@@ -52,22 +52,32 @@ export class AndroidDeviceHashService implements Mobile.IAndroidDeviceHashServic
...
 		const devicePaths: string[] = [];
```
DimitarTachev: Do we still need this array?

mobile/android/android-device-hash-service.ts
```diff
@@ -52,22 +52,32 @@ export class AndroidDeviceHashService implements Mobile.IAndroidDeviceHashServic
...
 			devicePaths.push(`"${localToDevicePathData.getDevicePath()}"`);
 		};
 
 		await executeActionByChunks<Mobile.ILocalToDevicePathData>(localToDevicePaths, DEFAULT_CHUNK_SIZE, action);
 
-		return devicePaths;
+		return initialShasums;
+	}
+
+	public getDevicePaths(localToDevicePaths: Mobile.ILocalToDevicePathData[]): string[] {
```
DimitarTachev: We can merge `getDevicePaths` and `generateHashesFromLocalToDevicePaths` (by returning another interface) in order to avoid iterating `localToDevicePaths` twice.

mobile/android/android-device-hash-service.ts
KristianDD: I prefer to implicitly return a result(more readable). However there is some code that depends on the shsums to be updated. I guess it can be accomplished only with return value, but I am not sure its worth the extra complication of the code.

mobile/android/android-device-hash-service.ts
KristianDD: This will make the name of the method generate HashesAndDevicePaths.... which raises a red flag for me. It can be implemented, but my personal preference is to remain as separate methods.

Also getDevicePaths is used only in one place, and generateHashesFromLocalToDevicePaths in 3, which means more often we don't really need them.

PR #1122, opened 2018-08-03
fix: fix lodash vulnerability and unit tests with Node.js 10

rosen-vladimirov:
## fix: update lodash
Update lodash to its latest version in order to fix a vulnerability issue.

## fix: unit tests fail with Node.js 10.x.x
Unit tests fail with Node.js 10.x.x as the behavior of `util.format` has been changed. When an object is passed to it and some of the properties of the object are long strings, the util.format adds new lines in some unexpected places.
In order to ensure tests work correctly in Node.js 10 and 8, do not pass object to util.format in the ErrorsStub


TsvetanMilanov:
:+1:

PR #1124, opened 2018-08-07
fix: `tns run android` command fails when no emulators are available

rosen-vladimirov:
`tns run android` command fails when there are no Android emulators available. The problem is that the output of `avdmanager list avds` is incorrectly parsed by CLI and it fails (output is `Available Android Virtual Devices:\n`).
Fix this by improving the parsing logic.
Also fix incorrect checks for executables on Windows - the files have `.bat` extension, which was not added, so the check if the file exist was failing.
Unify the calls to `adb devices` - curently there are two implementations on different places in the code - unify them and add a new method that does not fail in case `adb devices` command fails. The new method is usable when we want to list currently available simulators/emulators.

### DIFF
--- a/mobile/mobile-core/android-device-discovery.ts
+++ b/mobile/mobile-core/android-device-discovery.ts
@@ -1,7 +1,5 @@
 import { DeviceDiscovery } from "./device-discovery";
-import * as helpers from "../../helpers";
 import { AndroidDevice } from "../android/android-device";
-import { EOL } from "os";
 
 interface IAdbAndroidDeviceInfo {
 	identifier: string;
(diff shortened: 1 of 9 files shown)


mobile/mobile-core/android-device-discovery.ts
```diff
@@ -39,61 +37,22 @@ export class AndroidDeviceDiscovery extends DeviceDiscovery implements Mobile.IA
...
-				return {
-					identifier: identifier,
-					status: status
-				};
-			});
+	private async checkCurrentData(result: string[]): Promise<void> {
+		const currentDevices: IAdbAndroidDeviceInfo[] = result.map((element: string) => {
+			// http://developer.android.com/tools/help/adb.html#devicestatus
+			const data = element.split('\t'),
+				identifier = data[0],
```
DimitarTachev: its more readable with three consts

test/unit-tests/mobile/android-debug-bridge.ts
```diff
@@ -166,16 +180,25 @@ describe('androidDebugBridge', () => {
...
-			assert.deepEqual(allSpawnedEventsArgs[0], [ 'shell', 'mkdir', '-p', sampleDeviceFolder ]);
-			assert.deepEqual(allSpawnedEventsArgs[1], [ 'push', sampleLocalFilePath, deviceFilePath ]);
-			assert.deepEqual(allSpawnedEventsArgs[2], [ 'shell', 'chmod', '0777', sampleDeviceFolder ]);
+			assert.deepEqual(allSpawnedEventsArgs[0], ['shell', 'mkdir', '-p', sampleDeviceFolder]);
+			assert.deepEqual(allSpawnedEventsArgs[1], ['push', sampleLocalFilePath, deviceFilePath]);
+			assert.deepEqual(allSpawnedEventsArgs[2], ['shell', 'chmod', '0777', sampleDeviceFolder]);
+		});
+	});
+
+	describe("tryGetDevices", () => {
```
DimitarTachev: don't forget to rename the describe if you rename the method

